PR #8, opened 2020-05-07
propose buildpack.yml schema

dwillist:
propose buildpack.yml schema and outline what it should look like across language families
Signed-off-by: Daniel Thornton <dthornton@pivotal.io>


ForestEckhardt:
Along with whether or not the name should change, should we continue to use yaml or should we switch to another format such as toml or json?

accepted/0000-buildpack-yml.md
```diff
@@ -0,0 +1,130 @@
...
+    - <string>
+    - <string>
+```
+
+#### PHP
+```yaml
+php:
+  dist:
+    version: <semver>
+  web: # currently unsupported, all keys underneat this are also unsupported.
```
kvedurmu, line 75: typo

accepted/0000-buildpack-yml.md
```diff
@@ -0,0 +1,130 @@
...
+    import-path: <string>
+    ldflags:
+      <string>: <string>
+      <string>: <string>
+    targets:
+    - <string>
+    - <string>
+```
+
+#### Ruby
```
kvedurmu, line 103: I find it odd that we're adding this, even though Ruby isn't an official Paketo implementation yet.

accepted/0000-buildpack-yml.md
```diff
@@ -0,0 +1,130 @@
...
+
+The rationale for choosing this format is that language-family and implementation namespaces are currently the only mechanism for grouping configuration settings that can be consumed by buildpacks.
+
+An alternative to this type of grouping would be to allow buildpack authors to choose any mechanism that fits their own needs. Those systems, being unstandardized, would be difficult to integrate and likely lead to incompatibilities of buildpacks that may rely upon the same keys for different configuration settings.
+
+### Decision Points
+
+#### Should all keys be listed under a language family?
+It logically groups configuration for a buildpack family and reduces top level key pollution.
+
```
arjun024, line 34: Since language-family/meta-cnbs are abstract ordered lists, can't anybody write a meta cnb with their own name? In that case, it seems like an implementation buildpack would have no awareness of which "language-family" section its data sits under.
If the cnb searches the entire tree to find the relevant key, there's a chance of duplication and the burden of conflict resolution. How about a shallower tree with buildpack-ids (or whatever is unique) being the first level keys?

accepted/0000-buildpack-yml.md
```diff
@@ -0,0 +1,130 @@
...
+### Steps
+
+1. Language families should be reviewed to identify all properties that will need to be realigned.
+2. Issues or PRs should be opened to address moving each property, taking into account that the property might need to be "deprecated" so that users will not have a terrible experience once the property moves to its new location.
+3. Once all properties have been realigned for a language family, properties should be documented and outlined for users and buildpack authors.
+
+## Unresolved Questions and Bikeshedding
+
+1. Is `buildpack.yml` the right name?
+
```
arjun024, line 130: Would love to see this named something else. config yaml?

dmikusa-pivotal: I don't mind the current name. It specific to what is contained in the file & it's unlikely to collide with anything that might already exist in a user's application.

dmikusa-pivotal:
I'm not sure I understand the problem this is meant to address or the value proposition of it.

My perception is that what authors are storing in buildpack.yml has naturally organized itself into namespaced groups. Once an author has picked the properties and documented them, it also seems very unlikely the author would change them, so I don't personally anticipate many if any, interop challenges.

If the concern is interop though, maybe we just provide some guidelines that a well behave CNB should follow? Like...

1.) Pick a top-level namespace under which config for your CNB should reside. This needs to be unique, so please avoid selecting a namespace used by an existing CNB (maybe we could have a list of these so it's easy for authors to find ones that are in use already?).
2.) Consider the structure of settings in buildpack.yml a public interface or contract with your users. If it needs to change, announce the change in advance, go through a deprecation period, and attempt to do this in a way that is considerate to users given your version schema (ie. don't change in a maintenance release).

I'm sure there's other best-practices we could offer up to authors, but that is the gist of it.

sclevine:
Would like to propose that this be an optional extension that recommends how to provide UX for buildpack authors who have this use case.

sclevine:
Also: consider CF-compatibility buildpack use cases (do we need to translate old file formats into new file formats?)

nebhale:
Closing in lieu of #9

PR #9, opened 2020-05-13
RFC: Replace buildpack.yml by using the build-plan buildpack

ForestEckhardt:
[Readable](https://github.com/paketo-buildpacks/rfcs/blob/replace-buildpack-yml/accepted/0000-replace-buildpack-yml.md)

### DIFF
--- a/accepted/0000-replace-buildpack-yml.md
+++ b/accepted/0000-replace-buildpack-yml.md
@@ -1,0 +1,21 @@
+# Replace buildpack.yml with Build Plan (TOML)
+
+## Summary
+
+`buildpack.yml` is a Paketo specific construct that requires increased overhead for buildpack authors. Everything that is done in `buildpack.yml` can be done with the [`build-plan` buildpack](https://github.com/ForestEckhardt/build-plan), which would offload some logic onto the lifecycle.
+
+## Motivation
+
+With this change we can completely eliminate `buildpack.yml` which is a file that adds undue complexity to the buildpacks which could be offloaded entirely to the lifecycle.
+
+## Implementation
+
+The `build-plan` buildpack would need to be added to the end of every build order. If you would like a specific dependency then you can specify that is a `plan.toml`, which is the `requires` portion of a Build Plan (TOML). This format is more powerful than `buildpack.yml` because it allows for fine grain controls as to when that dependency is available by setting `build` and `launch` flags. Any other data that is not the dependency name or version can be put in the metadata section of requires and handled during the `build` phase. By using the `build-plan` buildpack we can completely replace the `buildpack.yml` while losing no functionality, increase user configurability with very low overhead cost for buildpack authors, and vastly increase interoperability.
+
+## Unresolved Questions and Bikeshedding
+
+- Can we fold `plan.toml` into another configuration file (like `project.toml`)?
+- Are there things we can do/change to make the interface easier for users to understand?
+
+{{THIS SECTION SHOULD BE REMOVED BEFORE RATIFICATION}}
+
--- a/accepted/0000-replace-buildpack-yml.md
+++ b/accepted/0000-replace-buildpack-yml.md
@@ -14,6 +14,8 @@ The format used in `build-plan` is also more powerful than `buildpack.yml` becau
 
 By using the `build-plan` buildpack we can completely replace the `buildpack.yml` while losing no functionality, increase user configurability with very low overhead cost for buildpack authors, and vastly increased interoperability.
 
+The ideal end state for this track or work would be to bake this functionality directly into the lifecycle because of how closely the `build-plan` buildpack already interacts with the lifecycle. The `build-plan` buildpack is a middle ground to test out whether or not this model is a good thing for users and buildpack authors. Once this model has proven (or disproven) itself, learnings from this stop gap should be used to inform further actions (whether that be pursuing integration in the lifecycle or abandoning this model for something else).
+
 ## Implementation
 
 The `build-plan` buildpack would need to be added to the end of every build order. If you would like a specific dependency then you can specify that is a `plan.toml`, which is the `requires` portion of a Build Plan (TOML).
--- a/accepted/0000-replace-buildpack-yml.md
+++ b/accepted/0000-replace-buildpack-yml.md
@@ -6,11 +6,37 @@
 
 ## Motivation
 
-With this change we can completely eliminate `buildpack.yml` which is a file that adds undue complexity to the buildpacks which could be offloaded entirely to the lifecycle.
+With this change we can completely eliminate `buildpack.yml` which is a file that adds undue complexity to the buildpacks because in order for `buildpack.yml` influence what appears in the application image it must be parsed during detection in every buildpack. `buildpack.yml` also has a non-standard format meaning that every single buildpack needs to have slightly unique parsing logic in order to grab a buildpack specific key name and then any other additional metadata.
+
+The big advantage of moving to using `build-plan` is that every buildpack that currently has a `buildpack.yml` parser could remove that code entirely and switch to using Buildpack Plan metadata during the `build` phase and there would be no loss of functionality. The job of parsing build configuration will be condensed into the `build-plan` and then all subsequent parsing will be handled by the buildpack library or the lifecyle.
+
+The format used in `build-plan` is also more powerful than `buildpack.yml` because it allows for fine grain controls as to when that dependency is available by setting `build` and `launch` flags. Any other data that is not the dependency name or version can be put in the metadata section of requires and handled during the `build` phase.
+
+By using the `build-plan` buildpack we can completely replace the `buildpack.yml` while losing no functionality, increase user configurability with very low overhead cost for buildpack authors, and vastly increased interoperability.
 
 ## Implementation
 
-The `build-plan` buildpack would need to be added to the end of every build order. If you would like a specific dependency then you can specify that is a `plan.toml`, which is the `requires` portion of a Build Plan (TOML). This format is more powerful than `buildpack.yml` because it allows for fine grain controls as to when that dependency is available by setting `build` and `launch` flags. Any other data that is not the dependency name or version can be put in the metadata section of requires and handled during the `build` phase. By using the `build-plan` buildpack we can completely replace the `buildpack.yml` while losing no functionality, increase user configurability with very low overhead cost for buildpack authors, and vastly increase interoperability.
+The `build-plan` buildpack would need to be added to the end of every build order. If you would like a specific dependency then you can specify that is a `plan.toml`, which is the `requires` portion of a Build Plan (TOML).
+
+An example of what the conversion from `buildpack.yml` to `plan.toml` would look like the following:
+
+`buildpack.yml`
+```yaml
+nodejs:
+  engine:
+    version: ~10
+    optimize-memory: true
+```
+
+`plan.toml`
+```toml
+[[requires]]
+name = "node-engine"
+version = "~10"
+
+[requires.metadata]
+    optimize-memory = true
+```
 
 ## Unresolved Questions and Bikeshedding
 
@@ -18,4 +44,3 @@ The `build-plan` buildpack would need to be added to the end of every build orde
 - Are there things we can do/change to make the interface easier for users to understand?
 
 {{THIS SECTION SHOULD BE REMOVED BEFORE RATIFICATION}}
-
--- a/accepted/0000-replace-buildpack-yml.md
+++ b/accepted/0000-replace-buildpack-yml.md
@@ -14,15 +14,11 @@ The format used in `build-plan` is also more powerful than `buildpack.yml` becau
 
 By using the `build-plan` buildpack we can completely replace the `buildpack.yml` while losing no functionality, increase user configurability with very low overhead cost for buildpack authors, and vastly increased interoperability.
 
-The ideal end state for this track or work would be to bake this functionality directly into the lifecycle because of how closely the `build-plan` buildpack already interacts with the lifecycle. The `build-plan` buildpack is a middle ground to test out whether or not this model is a good thing for users and buildpack authors. Once this model has proven (or disproven) itself, learnings from this stop gap should be used to inform further actions (whether that be pursuing integration in the lifecycle or abandoning this model for something else).
-
-It is worth mentioning that the `build-plan` buildpack will be able to `require` any other buildpack, but if that buildpack is not in the group that the `build-plan` buildpack is running in then detection will fail. The `build-plan` buildpack needs to be running in a group with all of its required buildpacks which is something that can be configured by an operator or someone with operator level permissions.
-
 ## Implementation
 
 The `build-plan` buildpack would need to be added to the end of every build order. If you would like a specific dependency then you can specify that is a `plan.toml`, which is the `requires` portion of a Build Plan (TOML).
 
-In order to facilitate the use of the `build-plan` buildpack it should be migrate from its current location into the `paketo-community` organization. This will allow it to be more easily integrated into the Paketo ecosystem as a whole.
+In order to facilitate the use of the `build-plan` buildpack it should be migrated from its current location into the `paketo-community` organization. This will allow it to be more easily integrated into the Paketo ecosystem as a whole.
 
 An example of what the conversion from `buildpack.yml` to `plan.toml` would look like the following:
 
@@ -48,5 +44,15 @@ version = "~10"
 
 - Can we fold `plan.toml` into another configuration file (like `project.toml`)?
 - Are there things we can do/change to make the interface easier for users to understand?
+
+## Additional Information and Consideration
+
+The `build-plan` buildpack will be able to `require` any other buildpack, but if that buildpack is not in the group that the `build-plan` buildpack is running in then detection will fail. The `build-plan` buildpack needs to be running in a group with all of its required buildpacks which is something that can be configured by an operator or someone with operator level permissions.
+
+In order for configuration information to be passed to a buildpack by the `build-plan` buildpack, that buildpack must be making some provision. There are two approaches to this problem, either we classify any buildpack that has configuration as making a provision or we find a way to make configuration happen in buildpacks where a provision doesn't make sense (allowing configuration to be set by environment variables or some other alternative).
+
+## Potential Long-Term Goals
+
+The ideal end state for this track or work would be to bake this functionality directly into the lifecycle because of how closely the `build-plan` buildpack already interacts with the lifecycle. The `build-plan` buildpack is a middle ground to test out whether or not this model is a good thing for users and buildpack authors. Once this model has proven (or disproven) itself, learnings from this stop gap should be used to inform further actions (whether that be pursuing integration in the lifecycle or abandoning this model for something else).
 
 {{THIS SECTION SHOULD BE REMOVED BEFORE RATIFICATION}}
--- a/accepted/0000-replace-buildpack-yml.md
+++ b/accepted/0000-replace-buildpack-yml.md
@@ -22,6 +22,8 @@ It is worth mentioning that the `build-plan` buildpack will be able to `require`
 
 The `build-plan` buildpack would need to be added to the end of every build order. If you would like a specific dependency then you can specify that is a `plan.toml`, which is the `requires` portion of a Build Plan (TOML).
 
+In order to facilitate the use of the `build-plan` buildpack it should be migrate from its current location into the `paketo-community` organization. This will allow it to be more easily integrated into the Paketo ecosystem as a whole.
+
 An example of what the conversion from `buildpack.yml` to `plan.toml` would look like the following:
 
 `buildpack.yml`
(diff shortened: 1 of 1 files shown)


dmikusa-pivotal:
1.) Can you expand on the motivation? You say there's undue complexity. Do you have some examples of how buildpack.yml adds complexity, and what could be removed without it?

2.) My initial thought is that this will make things easier for buildpack authors but harder for end-users. The requires/provides concept isn't the easiest thing to wrap one's brain around. I'd rather deal with a little extra complication as an author to make life easier for users. 

Could you perhaps expand on the benefits of this approach? Like what can you do using this approach that you couldn't do with just buildpack.yml? Perhaps a side-by-side comparison? Like I do this now with buildpack.yml and after this proposal, it'd look like this?

ForestEckhardt:
@dmikusa-pivotal I just pushed up a set of changes that hopefully drive out the motivation and give you a better sense of the advantages. I also added an example of what it would look like to convert from `buildpack.yml` to `plan.toml`. Please let me know if I could add anything (this is a draft after all and I welcome any and all suggestion).

I agree with you point that the UX (which is something I have put in the unanswered questions so that we can discuss it further) may be a little rough and I am willing to discuss how to make that UX better. I think that some of the cognitive burden that you are worried about is not necessarily present because we do not support a full `provides` `requires` workflow we only support the `requires` half with `plan.toml`. I have also talked to @ekcasey very briefly and there maybe some room to put something like this in `project.toml` but I don't have a bead on what that would look like either.

dmikusa-pivotal:
Thanks, that helps. Thinking through the PHP workflows, I can see that might simplify things in the CNBs for the selection of the webserver. We don't have to translate between the buildpack.yml option and add requires entries to the build plan. The support person in me is thinking that this opens opportunities for some additional failure scenarios though, like putting multiple webservers or fat fingering the name. The current method allows us to inject some friendly errors for situations like these.

One other thing that I thought about after my previous post is that we have a third option, the model used by the Java-related buildpacks. They do not have buildpack.yml, but instead, take user-provided configuration as env variables. This option would seem to address some of the pain points you've highlighted also. It removes the need to parse a file, so we could remove the custom parsing code & config objects. It also allows setting separate config for build or run, like you called out. Plus it would remove buildpack specific configuration files from user projects, which is kind of nice. As an app dev, there's already a lot of little config files littering my projects getting rid of one would be nice. I'd say the main drawback that comes to mind is having to set bunches of env variables can sometimes be cumbersome, like with `docker run`, you end up with a really long command.

At any rate, it might be worth considering env variables as an option. If we're talking about recommending a standard way for buildpack authors to take user configuration.

ForestEckhardt:
@dmikusa-pivotal I think that environment variables could work for supplying configuration during the `build` phase but I think that it falls short if the app dev wants to specify during which phase a dependency is present. I think that you end up in the same trap of having lots of buildpack specific detection logic for something that one buildpack and the lifecycle will handle for you automatically.

Perhaps an example might help. Say that I am using a static website generating tool written in Ruby and then runs using NGINX. Currently I can, using `buildpack.yml`, install a version of Ruby but I have no control as to when it is installed, same with NGINX. So there is no way for me to indicate that I just want Ruby during the `build` phase and just want NGINX during the `launch` phase.

To do that with `buildpack.yml` I would have to essentially turn it into a worse form of `plan.toml` and I cannot really think of a super sane way to do that with straight environment variables. If you can make requirements from a configuration then we as buildpack authors can be very reserved about what phases dependency appear in and let other buildpacks only pull in exactly what they require or enable the user to pull in what they require with out having to write a buildpack.

I also think your point about how hard it would be to inject a large number of environment variable during your `pack build` command is worth mentioning and maybe we could do something to easy that (I am think another buildpack :)) but having to inject your environment configuration like that feels as accident prone as writing a config file to me.

dmikusa-pivotal:
>Perhaps an example might help. Say that I am using a static website generating tool written in Ruby and then runs using NGINX. Currently I can, using buildpack.yml, install a version of Ruby but I have no control as to when it is installed, same with NGINX. So there is no way for me to indicate that I just want Ruby during the build phase and just want NGINX during the launch phase.

I agree I don't think this kind of thing should be done through buildpack.yml or env variables. It seems like something you'd want a CNB to do, and I think that's where I struggle with this. Buildpack.yml and env variables seem like a good way to provide settings and config to a buildpack (override default values mainly), but not to alter the composition of buildpacks. If you want to make some new combinations like what you're describing here, it seems like some other mechanism should be used. Similarly, I agree that in cases like for the PHP CNBs where we use buildpack.yml to pick a webserver that we've crossed over into something that should be done through the build plans.

If this was just a new way to provide configuration, like just the metadata portion of your examples, I think I would have fewer reservations. It's primarily the ability for app devs/consumers of a set of CNBs to require stuff seems to cross boundaries I'm not sure should be crossed. I worry what someone could do with that to break a carefully crafted set of CNBs, or perhaps even pull in some code that should not be allowed. My $0.02 though.

ForestEckhardt:
> If this was just a new way to provide configuration, like just the metadata portion of your examples, I think I would have fewer reservations. It's primarily the ability for app devs/consumers of a set of CNBs to require stuff seems to cross boundaries I'm not sure should be crossed. I worry what someone could do with that to break a carefully crafted set of CNBs, or perhaps even pull in some code that should not be allowed. My $0.02 though.

I think that the way `buildpack.yml` is wired up this can already happen. If I have something declared in my `buildpack.yml` with a version it will make require that dependency. The problem with the `buildpack.yml` requirement is that it is more a shotgun than a scalpel, you don't know what other buildpacks you might hit along the way. Using `plan.toml` you can be much more precise and granular. So I think in the end we have a tool that is more predictable and still accomplishes the same task.

It is also worth noting that I believe there is a lot of control as to what buildpacks you have access to at the platform layer. So being able to pull something you shouldn't is less of concern for me.

You do raise a solid point about potentially separating the requirement section for the build configuration section. But I think that keeping `buildpack.yml` does not solve that problem and really only creates more issue in terms of extensibility. Maybe we could move toward that model in the future if it becomes a serious problem.

ForestEckhardt:
A thought that should be captured is that this leads to a world where a buildpack never requires itself which I think makes more sense than the current state of the world where a buildpack can and often does both provide and require itself which is a concept that is kind of hard to wrap your head around when you are new to the buildpacks ecosystem.

dmikusa-pivotal:
>It is also worth noting that I believe there is a lot of control as to what buildpacks you have access to at the platform layer. So being able to pull something you shouldn't is less of concern for me.

With `pack`, sure, but as you start talking about other systems that will integrate and use CNBs, I don't think that's an assumption we should make. 

For example, with Cloud Foundry now it's quite possible for an operator to lock down the list of usable buildpacks. I suspect that going forward with CNBs operators will want similar functionality. At any rate, this is speculative, but I think it's legitimate that this proposal should consider how/if someone could use the functionality exposed to bypass restrictions or worse, inject unexpected code or operations.

dmikusa-pivotal:
>The build-plan buildpack would need to be added to the end of every build order.

Why wouldn't this be added to the lifecycle? If it should happen every time, it seems like it could be enforced more easily and consistently by having the functionality in the lifecycle. Is that not an option or are there reasons to do it this way, like cases where you might want to bypass this extra CNB?

ForestEckhardt:
@dmikusa-pivotal I think that it would be great to eventually get this functionality baked into the lifecycle but I think that some implementation first would be hugely beneficial to proving out whether or not this is a good idea.

dmikusa-pivotal:
>I think that the way buildpack.yml is wired up this can already happen. If I have something declared in my buildpack.yml with a version it will make require that dependency.

Right, but what you can do in buildpack.yml is extremely limited. It's intentionally constrained by the code in the buildpack so it's predictable and auditable. An operator can look at the code and know the limits of what can be changed by an app dev. Unrestricted ability to require additional CNBs is a *lot* of power.

I think if you want to keep operators happy with a change like this, you're going to need to give them some power to whitelist or constrain in some way what app devs can do. Maybe there's a way to do that already, idk, but it seems like a use case to cover in the rfc.

The other similar use case would be as a buildpack developer, I would like something similar. If I knowingly test my CNBs in certain ways and combinations I would want to be able to enforce that or at least provide proper validation. 

Right now when there is a detection problem, I don't think it's clear enough what is happening. If I write a CNB that works with HTTPD & Nginx, but not Lighttpd and the user requires Lighttpd you're just going to get something like this:

```
===> DETECTING
[detector] ERROR: No buildpack groups passed detection.
[detector] ERROR: Please check that you are running against the correct path.
[detector] ERROR: failed to detect: no buildpacks participating
ERROR: failed with status code: 6
```

The default output really tells you nothing except that a problem occurred. While the verbose output is definitely more verbose, it's can be, depending on the size of your builder, super easy to miss something in what can be a ton of output.

```
===> DETECTING
[detector] ======== Results ========
[detector] pass: paketo-buildpacks/bellsoft-liberica@2.5.3
[detector] skip: paketo-buildpacks/gradle@1.2.0
[detector] skip: paketo-buildpacks/maven@1.4.0
[detector] skip: paketo-buildpacks/sbt@1.2.0
[detector] pass: paketo-buildpacks/executable-jar@1.2.3
[detector] pass: paketo-buildpacks/apache-tomcat@1.1.3
[detector] pass: paketo-buildpacks/dist-zip@1.3.0
[detector] pass: paketo-buildpacks/spring-boot@1.5.3
[detector] skip: paketo-buildpacks/procfile@1.3.3
[detector] skip: paketo-buildpacks/azure-application-insights@1.1.3
[detector] skip: paketo-buildpacks/debug@1.2.3
[detector] skip: paketo-buildpacks/google-stackdriver@1.1.3
[detector] skip: paketo-buildpacks/jmx@1.1.3
[detector] skip: paketo-buildpacks/encrypt-at-rest@1.2.3
[detector] skip: paketo-buildpacks/image-labels@1.0.3
[detector] Resolving plan... (try #1)
[detector] skip: paketo-buildpacks/executable-jar@1.2.3 requires jvm-application
[detector] skip: paketo-buildpacks/apache-tomcat@1.1.3 requires jvm-application
[detector] skip: paketo-buildpacks/dist-zip@1.3.0 requires jvm-application
[detector] skip: paketo-buildpacks/spring-boot@1.5.3 requires jvm-application
[detector] fail: paketo-buildpacks/bellsoft-liberica@2.5.3 provides unused jdk
[detector] Resolving plan... (try #2)
[detector] skip: paketo-buildpacks/executable-jar@1.2.3 requires jre
[detector] skip: paketo-buildpacks/apache-tomcat@1.1.3 requires jre
[detector] skip: paketo-buildpacks/dist-zip@1.3.0 requires jre
[detector] skip: paketo-buildpacks/executable-jar@1.2.3 requires jvm-application
[detector] skip: paketo-buildpacks/apache-tomcat@1.1.3 requires jvm-application
[detector] skip: paketo-buildpacks/dist-zip@1.3.0 requires jvm-application
[detector] skip: paketo-buildpacks/spring-boot@1.5.3 requires jvm-application
[detector] fail: paketo-buildpacks/bellsoft-liberica@2.5.3 provides unused jdk
[detector] Resolving plan... (try #3)
[detector] skip: paketo-buildpacks/executable-jar@1.2.3 requires jvm-application
[detector] skip: paketo-buildpacks/apache-tomcat@1.1.3 requires jvm-application
[detector] skip: paketo-buildpacks/dist-zip@1.3.0 requires jvm-application
[detector] skip: paketo-buildpacks/spring-boot@1.5.3 requires jvm-application
[detector] fail: paketo-buildpacks/bellsoft-liberica@2.5.3 provides unused jre
[detector] ======== Output: paketo-buildpacks/yarn-install@0.1.19 ========
[detector]
[detector] failed
[detector] ======== Results ========
[detector] pass: paketo-buildpacks/node-engine@0.0.178
[detector] fail: paketo-buildpacks/yarn-install@0.1.19
[detector] ======== Output: paketo-buildpacks/npm@0.1.11 ========
[detector]
[detector] failed
[detector] ======== Results ========
[detector] pass: paketo-buildpacks/node-engine@0.0.178
[detector] fail: paketo-buildpacks/npm@0.1.11
[detector] ======== Output: paketo-buildpacks/go-mod@0.0.96 ========
[detector] no "go.mod" found at: /workspace/go.mod
[detector] ======== Results ========
[detector] pass: paketo-buildpacks/go-compiler@0.0.112
[detector] fail: paketo-buildpacks/go-mod@0.0.96
[detector] ======== Output: paketo-buildpacks/dep@0.0.109 ========
[detector] failed detection: no Gopkg.toml found at root level
[detector] ======== Results ========
[detector] pass: paketo-buildpacks/go-compiler@0.0.112
[detector] fail: paketo-buildpacks/dep@0.0.109
[detector] ======== Output: paketo-buildpacks/dotnet-core-build@0.0.70 ========
[detector] no proj file found
[detector] ======== Output: paketo-buildpacks/dotnet-core-conf@0.0.122 ========
[detector] *.runtimeconfig.json file not found and expecting only a single *.csproj file in the app directory
[detector] ======== Results ========
[detector] pass: paketo-buildpacks/node-engine@0.0.178
[detector] pass: paketo-buildpacks/icu@0.0.52
[detector] pass: paketo-buildpacks/dotnet-core-runtime@0.0.135
[detector] pass: paketo-buildpacks/dotnet-core-aspnet@0.0.128
[detector] pass: paketo-buildpacks/dotnet-core-sdk@0.0.133
[detector] skip: paketo-buildpacks/dotnet-core-build@0.0.70
[detector] fail: paketo-buildpacks/dotnet-core-conf@0.0.122
[detector] ======== Output: paketo-buildpacks/php-composer@0.0.65 ========
[detector] no "composer.json" found in the following locations: [/workspace/composer.json /workspace/htdocs/composer.json]
[detector] ======== Results ========
[detector] pass: paketo-buildpacks/httpd@0.0.107
[detector] pass: paketo-buildpacks/nginx@0.0.110
[detector] pass: paketo-buildpacks/php-dist@0.0.135
[detector] skip: paketo-buildpacks/php-composer@0.0.65
[detector] fail: paketo-buildpacks/php-web@0.0.93
[detector] ======== Results ========
[detector] pass: paketo-buildpacks/httpd@0.0.107
[detector] Resolving plan... (try #1)
[detector] fail: paketo-buildpacks/httpd@0.0.107 provides unused httpd
[detector] ======== Results ========
[detector] pass: paketo-buildpacks/nginx@0.0.110
[detector] Resolving plan... (try #1)
[detector] fail: paketo-buildpacks/nginx@0.0.110 provides unused nginx
[detector] ======== Results ========
[detector] fail: paketo-buildpacks/procfile@1.3.3
[detector] ERROR: No buildpack groups passed detection.
[detector] ERROR: Please check that you are running against the correct path.
[detector] ERROR: failed to detect: no buildpacks participating
ERROR: failed with status code: 6
```

What failed here? It's a lot to sift through.

This isn't really a problem with this proposal, it's just the current UI/output from detection, but this proposal puts more reliance on the detection output because it offloads more work onto the detection process so it's going to be more critical to users that there is good clear output indicating what happened and why. 

Contrast the output above to what you get now. If there's a problem with buildpack.yml input buildpack authors can validate and provide custom, clear messages to indicate what happened.

Again, I think this proposal should address the validation of input & error handling as well.

dmikusa-pivotal:
>@dmikusa-pivotal I think that it would be great to eventually get this functionality baked into the lifecycle but I think that some implementation first would be hugely beneficial to proving out whether or not this is a good idea.

OK, I would suggest clarifying that in the RFC then. What you see as the end state & how you can use the cnb you have to test things now. I think normally an RFC would describe the final/ideal state trying to be achieved.

accepted/0000-replace-buildpack-yml.md
```diff
@@ -0,0 +1,52 @@
...
+
+The format used in `build-plan` is also more powerful than `buildpack.yml` because it allows for fine grain controls as to when that dependency is available by setting `build` and `launch` flags. Any other data that is not the dependency name or version can be put in the metadata section of requires and handled during the `build` phase.
+
+By using the `build-plan` buildpack we can completely replace the `buildpack.yml` while losing no functionality, increase user configurability with very low overhead cost for buildpack authors, and vastly increased interoperability.
+
+The ideal end state for this track or work would be to bake this functionality directly into the lifecycle because of how closely the `build-plan` buildpack already interacts with the lifecycle. The `build-plan` buildpack is a middle ground to test out whether or not this model is a good thing for users and buildpack authors. Once this model has proven (or disproven) itself, learnings from this stop gap should be used to inform further actions (whether that be pursuing integration in the lifecycle or abandoning this model for something else).
```
nebhale, line 17: I'm not totally sold on this being a goal for the CNB spec.  At this time, we don't have a single situation where the `lifecycle` knows about a file in the source that it is transforming, and this would cause that to happen.  I think this may end up being part of platform implementations, but this feels like we're getting ahead of ourselves.

/cc @ekcasey

accepted/0000-replace-buildpack-yml.md
```diff
@@ -0,0 +1,52 @@
...
+
+It is worth mentioning that the `build-plan` buildpack will be able to `require` any other buildpack, but if that buildpack is not in the group that the `build-plan` buildpack is running in then detection will fail. The `build-plan` buildpack needs to be running in a group with all of its required buildpacks which is something that can be configured by an operator or someone with operator level permissions.
```
nebhale, line 19: Detection will only fail if the `build-plan` buildpack is non-`optional`.  I think each `order` writer will have the opportunity to make that choice (correctness vs. resilience) and we should reason about the behaviors of this proposal with an eye towards buildpack failure *not* being terminal.

accepted/0000-replace-buildpack-yml.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Implementation
+
+The `build-plan` buildpack would need to be added to the end of every build order. If you would like a specific dependency then you can specify that is a `plan.toml`, which is the `requires` portion of a Build Plan (TOML).
```
nebhale, line 23: A better syntax be to be the _full_ `buildplan.toml` hierarchy, including `[or]`, in which all keys other than `[requires]` and it's children are ignored.  Best to properly mimic the thing you are duck-typing to ensure you get the full functionality and ease of future upgrades.

ForestEckhardt: I am fine with implementing the `[or]`structure into `plan.toml` but I feel that it would be inappropriate to include the `[provides]` section of `buildplan.toml` as the provisions should be given by the functional implementation buildpacks.

accepted/0000-replace-buildpack-yml.md
ForestEckhardt: I am not strictly opposed to this idea but I am also skeptical of it. It feels that you could end up in a situation that is confusing to a user if you make the buildpack `optional`. A user could specify a buildpack that is not in the `order` and then receive no feedback that the buildpack they specified is not being used in the build process because it was not present in the `order` grouping. By making the buildpack required they the user immediately get feedback that the buildpack that they required was not detect as present in the `order` grouping.

accepted/0000-replace-buildpack-yml.md
ForestEckhardt: I agree that this is out of the purview of the this particular RFC. I think I should move it to a different section that is `Potential Long-Term Goals`. This was added because of some prior feedback about how this RFC could be extended but I think that it is in an inappropriate section to indicate as such.

accepted/0000-replace-buildpack-yml.md
nebhale: I totally agree, but you, the buildpack author have *zero* control over whether an order makes a buildpack non-optional.  So you have to assume that it's a realistic scenario in the wild.  I think you're falling into the trap of someone who authors both a buildpack _and_ the order that will use it.

ForestEckhardt: So would you remove any reference as to how it is running in the groups at all and then let the group constructor decide whether or not they want to make it `optional`? If that is the case I think that is a fair point and I would be more than happy to remove this section.

accepted/0000-replace-buildpack-yml.md
ForestEckhardt: I have moved this point into a section at the bottom titled `Potential Long-Term Goals`.

accepted/0000-replace-buildpack-yml.md
nebhale: You can achieve this by following [Postel's Law](https://en.wikipedia.org/wiki/Robustness_principle).  If you have a strict subset of the build plan syntax, ignoring what you don't care about `[provides]` then you get the most compatible option without expanding the types of documents we already have.

ForestEckhardt: I'm not sure how to interpret that comment. 
1. If you a purposing also adding `[provides]` I would argue that having a field in configuration that is ignored during execution feels like bad user experience.
1. If you are purposing I make the `plan.toml` look like `buildplan.toml` except for `[provides]` then I would be all for that and I would happily make the necessary changes to the `build-plan` buildpack

nebhale: The latter precisely.

accepted/0000-replace-buildpack-yml.md
nebhale: Yep, removal sounds right to me.  We should also, during discussion, keep an eye on our assumptions as to usage.

accepted/0000-replace-buildpack-yml.md
ForestEckhardt: Changes are being made to the `build-plan` buildpack as we speak, I will update the RFC when those are landed.

accepted/0000-replace-buildpack-yml.md
ForestEckhardt: This section has been removed in the latest commit.

accepted/0000-replace-buildpack-yml.md
ForestEckhardt: The changes to the `build-plan` buildpack have been added and a generic `plan.toml` structure has been added to the RFC in the latest commit.

Issue #13, opened 2020-08-09
All Paketo buildpacks should follow naming convention "Paketo <runtime> Buildpack"


kvedurmu:
As discussed in the last working group meeting, all Paketo Buildpacks should follow naming convention "Paketo <runtime> Buildpack". We should update this in all buildpacks and write up an RFC for this as well so it's documented.

kvedurmu:
Closing this. This change has been made to all Paketo Buildpacks

PR #16, opened 2020-08-13
Buildpack Naming RFC

ryanmoran:
[Readable](https://github.com/ryanmoran/rfcs/blob/naming/accepted/0000-buildpack-naming.md)


ryanmoran:
Resolves #13

PR #18, opened 2020-08-20
Dependency server RFC

mdelillo:
[Readable](https://github.com/mdelillo/rfcs/blob/dep-server/accepted/0000-dep-server.md)


mdelillo:
@nebhale good catch, I updated the commit

nebhale (approved):

PR #21, opened 2020-09-18
Proposal to create a Paketo Project Programming Style Guide

fg-j:
This RFC will ultimately include a complete proposed Style Guide. This differs from the contribution guidelines in that it speaks directly to **how Paketo buildpacks code should be written** and focuses less on high-level project process.

[Readable](https://github.com/floragj/rfcs/blob/programming-style-guide/accepted/0000-programming-style-guide.md)


sophiewigmore:
I agree with the motivation and rationale that you have pointed out so far.

I am wondering how specific will the guide be able to get in how we actually style the buildpacks? For example, if we pointed people towards a sample buildpack as a "good example", would we want it to not use packit so as not to impose that library on anyone? Also is providing concrete examples at all too rigid for the guide? 

Overall, I guess it boils down to if our goal here to have people write code that abides by some standard, or are we also concerned with making it specific to writing good buildpacks, and the practices that come along with that?

fg-j:
**Some open questions from 11/3 WG conversation:**

* Do we want the style guide to allow for buildpacks that aren't written in Go and/or don't use the Paketo libraries?
* To what extent does the current version of the Style Guide meet the following criteria:
    * aspects of the guide are automate-able (PRs can be automatically checked)
    * the guide is a *useful tool for code review*
    * the guide reflects the project's "need to haves" - it's neither too prescriptive nor too vague
    * the guide is applicable to contributions to Java and non-Java buildpacks?

fg-j:
@arjun024 @ForestEckhardt @joshzarrabi @ryanmoran @sophiewigmore @thitch97 can't request you as reviewers, but would love to hear from y'all on this :)

sophiewigmore:
@fg-j For your first question "Do we want the style guide to allow for buildpacks that aren't written in Go and/or don't use the Paketo libraries?" I think yes - didn't we recently talk to a user who found it easier to write a buildpack in bash? I don't think we should add a barrier to entry of mandating someone use Paketo libraries; however, we should clearly define expectations if someone chooses not to.

accepted/0000-programming-style-guide.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Summary
+
+This RFC proposes a Programming Style Guide for the Paketo project. It will be
+used to evaluate and polish pull requests, educate contributors, and overall
+maintain consistent code quality in the project
+
+## Motivation
+
+The Programming Style Guide will addrress the challenge of maintaining
```
ryanmoran, line 11:
```suggestion
The Programming Style Guide will address the challenge of maintaining
```

accepted/0000-programming-style-guide.md
```diff
@@ -0,0 +1,103 @@
...
+Maintaining a Style Guide also improves the fairness and inclusivity of the
+project. Reviewers and contributors can ground their efforts in a set of
+explicit and agreed upon standards, which reduces the barrier to entry for new
+contributors to the project. Furthermore, standards-based feedback reduces the
+amount that implicit biases about whose work is "high quality" come into play.
+
+
+{{Why are we doing this? What pain points does this resolve? What use cases
+does it support? What is the expected outcome? Use real, concrete examples to
+make your case!}}
```
ryanmoran, line 35:
```suggestion
```

accepted/0000-programming-style-guide.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Detailed Explanation
+
+Explain the logistics of where the guide will be placed and referred to. Also
+talk about guiding principles for what's included in the guide.
+{{Describe the expected changes in detail.}}
```
ryanmoran, line 41:
```suggestion
```

accepted/0000-programming-style-guide.md
```diff
@@ -0,0 +1,103 @@
...
+
+
+{{Why are we doing this? What pain points does this resolve? What use cases
+does it support? What is the expected outcome? Use real, concrete examples to
+make your case!}}
+
+## Detailed Explanation
+
+Explain the logistics of where the guide will be placed and referred to. Also
+talk about guiding principles for what's included in the guide.
```
ryanmoran, line 40: Is this a note to your future self that this section needs to be expanded?

fg-j: Yes - I'm in general trying to go about this RFC a bit differently than others. It seems like the topics in this note should be addressed after the content of a style guide has been mostly established.

Though I'm beginning to wonder if this RFC should only be about approving the *existence* of a programming guide, and the actual implementation of the guide can happen... via other RFCs? Or by some other mechanism? What do you think about that @ryanmoran?

ryanmoran: This sounds reasonable. I don't think it makes sense for us to try to develop the entirety of the styleguide in this RFC. It makes more sense to decide where it will reside and what its initial contents might be. I'd be happy to ✅ an RFC that answered those questions.

ryanmoran (approved):

ForestEckhardt (approved):

ekcasey (approved):

PR #29, opened 2020-11-24
Propose go-http-fn buildpack

mattmoor:
* A short explanation of the proposed change:

Propose a `go-http-fn` buildpack for up-leveling the Go buildpack programming model from "app" to "function".

* An explanation of the use cases your change enables:

Please confirm the following:


mattmoor:
We've knocked out many of the outstanding AI's we'd set for ourselves with: https://github.com/mattmoor/http-go-fn, so if any of the Go buildpack folks have a chance to TAL and let us know if/what else might need work, we'd appreciate it.

mattmoor:
@sclevine @nebhale Just want to make sure I understand the next steps on this, so I know if/when I should bug folks for attention.

I think the two things I wanted to follow up on from the WG call were:
1. Find out how to do the order composition from @ryanmoran (any good examples / pointers?)
1. Talk to @sclevine about how to split out the generated stuff into a "build layer" (any good examples / pointers?)

thanks again for the discussion this morning!

mattmoor:
Also, on `1.` I presume we'd want a new repository `go-fn` that composes assorted go function buildpacks with the standard `go` order as optional contributors.  Once I understand the proposal a bit better, I can either open a new RFC or augment this one with those details.  Please LMK which makes the most sense, and I'd be happy to add those details.

mattmoor:
I chatted with @ryanmoran a bit on slack, and I've put together https://github.com/mattmoor/go-fn which does what he suggested with the new 0.2.7 release of the go buildpack collection.

accepted/0000-go-http-fn.md
```diff
@@ -0,0 +1,65 @@
...
+
+
+It is also notable that this "trivial" function signature example is a sort of canary, which will help us assess the readiness of downstream buildpacks for "function" style development, and act as a blueprint for future function buildpacks.
+
+
+## Detailed Explanation
+
+The intention here is to build a wholly separate buildpack based on https://github.com/mattmoor/http-go-fn which will eventually compose with the existing `go` Order like this: https://github.com/mattmoor/cloudevents-go-fn/blob/2e23ad960875b9b1c3633c041ed3011cefd9f6e2/buildpacks/order/buildpack.toml#L24-L27
+
+The PoC leverages golang utilities to analyze the signature structure of target methods, so the idea is that `bin/detect` would keep this function from kicking in unless it's what the user desired.  This also ultimately will allow us to compose with other types of function signatures (e.g. in the above example a cloudevents signature).
```
ekcasey, line 34: I worry that the proposed function signature may be common in non-serverless (serverful?) applications which may lead to a lot of false positives during detection. For example, `ServeHTTP` on [`http.Handler`](https://godoc.org/net/http#Handler) has the same signature. Although I assume an implementation of `http.Handler` would not detect, because of the receiver, this might be a clue indicating how common this signature is likely to be in the wild.

Many false positives during detection would make it risky/unattractive to add the function buildpack to `paketo-buildpacks/go`. It may be safer to require apps to opt-in, similar to how the Google buildpacks with `GOOGLE_FUNCTION_TARGET`.

mattmoor: Ack the signature is common, it is by definition `http.HandlerFunc`, so `ServeHTTP` naturally implements it.

I expect that folks will want the ability to customize the package and function name similar to `BP_GO_TARGETS` for the former in the Go buildpacks today, so I think configuration is necessary in general, but keeping it out of the most basic examples should still be the goal where we can avoid it.  I can see "Handler" being a bad default for this 😁 

I'd rather see us disambiguate with a better choice of unambiguous default function name than mandating configuration for the simplest case.  Something like `func PaketoFunction(r *http.Request, rw http.ResponseWriter)` seems like it'd be fairly unambigous. 🤷

ekcasey: `PaketoFunction` sounds good to me.

ryanmoran:
I don't believe there are any remaining objections to the general idea outlined in this RFC. There are certainly a number of smaller details to be worked out that would make sense in an incubation period in the `paketo-community` org. At this point, I think the only things that remain are to:

1. Establish a maintainer team to help manage the `http-go-fn` buildpack in its current state and shepherd the refinement of that buildpack into something we can GA.
2. Transfer the existing `http-go-fn` repo into the `paketo-community` org under this new team.
3. Create a top-level `functions` language-family buildpack that can compose the existing Go buildpack with the new `http-go-fn` buildpack and serve as a test bed for other additions.

From there, we can start to refine these buildpacks and build out our functions support. @ekcasey @mattmoor does this sound reasonable?

mattmoor:
That SGTM.  I've been bad about attending the WG calls since the holiday (and I'm on PTO next week).  Would it make sense to set up time to sync about how we coordinate next steps for this stuff, and what the punchlist items are?

ryanmoran:
After discussion with @ekcasey, we have decided to do a bit less than I outlined here.

We think it will be easier to just have the `http-go-fn` buildpack adopted into `paketo-community` under the existing Go subteam. This will ensure we have a set of maintainers to help drive that buildpack forward.

We've also decided to dispense with the top-level `functions` buildpack in favor of just documenting how you can combine the `http-go-fn` buildpack with the Go language-family buildpack at runtime. This should allow enough functionality for folks to start using the buildpack without us having to establish too much infrastructure to support it.

Assuming that this is the plan going forward, and being an existing maintainer on the Go language-family team, I'm happy to approve this RFC.

ryanmoran (approved):

ekcasey (approved):

PR #30, opened 2020-11-30
Dependency library RFC

martyspiewak:
[Readable](https://github.com/martyspiewak/rfcs-1/blob/666711d1f38873f806fb8ecf65cf0eaf2b3fb630/accepted/0000-dependency-library.md)

### DIFF
--- a/accepted/0000-dependency-library.md
+++ b/accepted/0000-dependency-library.md
@@ -1,0 +1,38 @@
+# Dependency Library to Provide Dependency Info
+
+## Summary
+
+The dependency metadata that is provided by the [Dep Server](https://github.com/paketo-buildpacks/dep-server) is backed partially by a Go library which provides information about upstream dependencies (available versions, release dates, source code URLs and SHAs, etc.). This RFC proposes creating a new repo that will expose this underlying resource.
+
+## Motivation
+
+By exposing this library, buildpack authors who use the Dep Server will have access to and insight into more of the underlying resources backing the server.
+
+Exposing this library will also allow authors to make contributions to add additional languages or modify the logic surrounding existing languages.
+
+## Detailed Explanation
+
+The library will contain two public methods that can be called for all upstream dependencies:
+  * `GetAllVersionRefs`:
+    * Lists all available versions of the dependency
+  * `GetDependencyVersion`:
+    * Returns information about a specific version (release date, source code URL and SHA, deprecation date (if available))
+
+## Rationale and Alternatives
+
+### Not expose this library
+We could choose not to add this library to paketo-buildpacks. Ideally all of the code backing the Dep Server would live in paketo-buildpacks and so choosing not to add this library would make it difficult to reach that goal.
+
+## Implementation
+
+The dependency library will live in a new repo (perhaps `github.com/paketo-buildpacks/dependency`).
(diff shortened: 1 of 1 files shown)


accepted/0000-dependency-library.md
```diff
@@ -0,0 +1,38 @@
...
+    * Returns information about a specific version (release date, source code URL and SHA, deprecation date (if available))
+
+## Rationale and Alternatives
+
+### Not expose this library
+We could choose not to add this library to paketo-buildpacks. Ideally all of the code backing the Dep Server would live in paketo-buildpacks and so choosing not to add this library would make it difficult to reach that goal.
+
+## Implementation
+
+The dependency library will live in a new repo (perhaps `github.com/paketo-buildpacks/dependency`).
```
ekcasey, line 28: What do you see as the main advantage of a new repo vs. adding an importable client package to the existing `dep-server` repo?

accepted/0000-dependency-library.md
```diff
@@ -0,0 +1,38 @@
+# Dependency Library to Provide Dependency Info
+
+## Summary
+
+The dependency metadata that is provided by the [Dep Server](https://github.com/paketo-buildpacks/dep-server) is backed partially by a Go library which provides information about upstream dependencies (available versions, release dates, source code URLs and SHAs, etc.). This RFC proposes creating a new repo that will expose this underlying resource.
+
+## Motivation
+
+By exposing this library, buildpack authors who use the Dep Server will have access to and insight into more of the underlying resources backing the server.
```
ekcasey, line 9: Can you elaborate a bit more about how/when paketo buildpack authors are currently using the dep server? Selfishly, given the Java buildpacks don't use the dep server, I am not as familiar with it. Capturing some of the background/context in the RFC might be good for posterity as well.

accepted/0000-dependency-library.md
martyspiewak: The library isn't called directly by the dep-server- the dep-server just returns JSON files from S3. The library is used to populate those S3 files.

accepted/0000-dependency-library.md
martyspiewak: I can't speak to who is using the server yet, but they would use it to find dependencies that have compiled and uploaded to a public S3 bucket so that they can include them in their buildpack.toml. I think [this RFC](https://github.com/paketo-buildpacks/rfcs/blob/master/accepted/0011-dep-server.md#detailed-explanation) might help explain how they might use it.

accepted/0000-dependency-library.md
ryanmoran: I still think it is reasonable to have the client library live side-by-side with the server.

accepted/0000-dependency-library.md
ryanmoran: We are not yet using the Dependency Server for the other language family buildpacks yet. This is [definitely in](https://github.com/paketo-buildpacks/feedback/discussions/2#discussioncomment-263526) the Roadmap for 2021 though. Having a client library for this would help a bit to enable developing this type of automation.

martyspiewak:
We are going to pull the library into the dep-server repo, as suggested.

PR #33, opened 2021-01-22
Radiate Buildpack and Community Metadata via a Dashboard

ryanmoran:
[Readable](https://github.com/paketo-buildpacks/rfcs/blob/dashboard-rfc/accepted/0000-dashboard.md)


sophiewigmore:
I like the idea of having a centralized place for users/developers to get an overview of the state of buildpacks at a glance. I am on board with having this be a part of the paketo-buildpacks Github org, as well as on the website for visibility.

One feature I feel strongly about including is the ability to filter the dashboard by language family. As a maintainer/contributor on Paketo Buildpacks it's easy for me to find what I'm looking for, but it might not be so clear to an outside user which buildpacks they should be looking at. I feel that being able to filter which buildpacks show up could make the user experience more straightforward.

ryanmoran:
> One feature I feel strongly about including is the ability to filter the dashboard by language family

This certainly sounds reasonable. I've created an issue on my current repo that will be transferred if/when this RFC is accepted and implemented: https://github.com/ryanmoran/paketo-dashboard/issues/2.

kvedurmu:
Do you think it makes sense to include this on the top level website nav or would it be enough to link to this from the eventual ["Contribute" page](https://github.com/paketo-buildpacks/paketo-website/issues/21)?

accepted/0000-dashboard.md
```diff
@@ -0,0 +1,132 @@
...
+Why are we doing this? What pain points does this resolve? What use cases does
+it support? What is the expected outcome? Use real, concrete examples to make
+your case!
+-->
+
+While GitHub notifications and search can provide some amount of insight into
+the overall health and state of the Paketo buildpack orgs, finding answers to
+many questions is not a simple task.  I personally have found myself with
+multiple tabs pinned in my browser showing a number of different details about
+the state of buildpacks across the orgs. I'd like to know:
```
kvedurmu, line 23: There was some discussion in the past of having some sort of tool to track the time it takes from when a high/medium/low CVE is resolved upstream by Canonical to releasing a patched buildpack/stack. Seems like this dashboard would be a good place to surface this information (once it's built). cc @paketo-buildpacks/core-dependencies

ryanmoran:
> Do you think it makes sense to include this on the top level website nav or would it be enough to link to this from the eventual ["Contribute" page](https://github.com/paketo-buildpacks/paketo-website/issues/21)?

It makes sense to me to link from the website to the dashboard, but where that link should live is less important to me. I'd be fine with either of these suggestions.

ekcasey (approved):

PR #36, opened 2021-02-02
Adds self-hosted blog RFC

ryanmoran:
[Readable](https://github.com/paketo-buildpacks/rfcs/blob/blog/accepted/0000-blog.md)


ekcasey (approved):

ForestEckhardt (approved):

kvedurmu (approved):
This looks great. Once the actual blog.paketo.io is implemented I think we should discontinue use of the Medium blog. Perhaps we can keep the existing posts up (so links from tweets and stuff still work), and figure out some way to link from medium to the Paketo blog.

PR #49, opened 2021-04-08
Add Semantic Versioning RFC

arjun024:
## Summary
[Rendered](https://github.com/arjun024/rfcs/blob/semantic-versioning/text/0029-semantic-versioning.md)

## Checklist


text/0026-semantic-versioning.md
```diff
@@ -0,0 +1,80 @@
...
+#### Buildpack
+| `Major` | `Minor` | `Patch` |
+|-|-|-|
+| \<Any incompatible change\> | \<Feature additions but compatible with previous release\> | \<Fixes but no major feature changes\> |
+| Change in Buildpack ID | Addition/removal of dependency<br>(but default-version not changed to a version with breaking change) | Bug fixes |
+| Change in supported Stacks | Buildpack runs a new command the causes change in behavior<br>(e.g. x-install buildpack adds running "x clean") | Performance improvement, internal tweaks |
+| Change in API (Provides/Requires) | Change of required build tool version to package buildpack<br>(e.g. go version) | CVE fixes |
+| Default version of dependency bumped to incompatible version<br>(e.g. php7 to php8) | New configuration option (e.g. new environment variable support) |  |
+| Removal of configuration options<br>(e.g. buildpack.yml) | Change in Buildpack API version |  |
+| Removal of publishing to a certain image registry<br>(e.g. gcr.io, dockerhub) | Publishing to a new image registry |  |
```
ryanmoran, line 64: I'm not sure I follow why we'd need to change the version of the buildpack given where it is published. There shouldn't be anything in the image that indicates that it can from any registry in particular. To me, no longer updating the images at a certain registry location doesn't strike me as a breaking change. It certainly isn't breaking to those using the image at the registries that still host the latest images, and it isn't breaking for those using the older images on the registry that will no longer be updated.

text/0026-semantic-versioning.md
```diff
@@ -0,0 +1,80 @@
...
+* Bumping PATCH is analogous to a "`fixes issue`" relation between releases.
+
+The following table lists the usual circumstances/events where bumping a particular component is expected:
+
+#### Buildpack
+| `Major` | `Minor` | `Patch` |
+|-|-|-|
+| \<Any incompatible change\> | \<Feature additions but compatible with previous release\> | \<Fixes but no major feature changes\> |
+| Change in Buildpack ID | Addition/removal of dependency<br>(but default-version not changed to a version with breaking change) | Bug fixes |
+| Change in supported Stacks | Buildpack runs a new command the causes change in behavior<br>(e.g. x-install buildpack adds running "x clean") | Performance improvement, internal tweaks |
```
ryanmoran, line 60: > Buildpack runs a new command the causes change in behavior
If this change in behavior was backwards incompatible, maybe requiring a change by the app developer, then it would require a major version bump, is that right?

text/0026-semantic-versioning.md
```diff
@@ -0,0 +1,80 @@
...
+| Change in API (Provides/Requires) | Change of required build tool version to package buildpack<br>(e.g. go version) | CVE fixes |
```
ryanmoran, line 61: > Change of required build tool version to package buildpack<br>(e.g. go version)
Do we really want to signal to end-users that the version of go we use to package the buildpack changed? Is that useful to them? Would they make different decisions knowing that?

I think any changes to packaging tools, as long as they do not affect the functioning of the buildpack probably just warrant a patch.

text/0026-semantic-versioning.md
```diff
@@ -0,0 +1,80 @@
...
+
+In terms of Paketo Buildpacks/Builders, this translates to:
+
+* Bumping MAJOR is analogous to an "`app or related buildpacks may not work without change`" relation between releases.
+* Bumping MINOR is analogous to an "`updates or adds features`" relation between releases.
+* Bumping PATCH is analogous to a "`fixes issue`" relation between releases.
+
+The following table lists the usual circumstances/events where bumping a particular component is expected:
+
+#### Buildpack
```
ryanmoran, line 55: Can we break out implementation buildpacks and composite buildpacks separately?

text/0026-semantic-versioning.md
```diff
@@ -0,0 +1,80 @@
...
+| Removal of configuration options<br>(e.g. buildpack.yml) | Change in Buildpack API version |  |
+| Removal of publishing to a certain image registry<br>(e.g. gcr.io, dockerhub) | Publishing to a new image registry |  |
+| (Language-family buildpacks)<br>Implementation buildpack additions/removals/major bumps<br>that cause incompatible change | Change in <br>[buildpack-specific metadata](https://github.com/buildpacks/spec/blob/buildpack/v0.6/buildpack.md#build-plan-toml) |  |
+
+#### Builder
+| `Major` | `Minor` | `Patch` |
+|-|-|-|
+| Lifecycle major bumps | Buildpack major bumps | Minor/patch bumps of components |
+| Stack Image major bumps |  |  |
+|  |  |  |
```
ryanmoran, line 72: Given the "User Policy" below, why wouldn't we also include buildpack major bumps in the "Major" column?

ryanmoran (commented):
This is a really great start. I have some comments about specific items outlined. To me, SemVer is about versioning the artifacts based on changes in their external behaviors. In some of the cases outlined, I don't understand how the change effects the external behavior for a buildpack user. Maybe outlining those effects would help to clarify the discussion.

text/0026-semantic-versioning.md
arjun024: I made it into a generic buildpack section and a separate section for things specific to language-family buildpacks

text/0026-semantic-versioning.md
arjun024: I moved that to the patch section, and added change of executable type to minor bump. The latter I feel is a significant yet compatible change from the perspective of a platform author.

text/0026-semantic-versioning.md
arjun024: Explicitly called out "backward compatible"

text/0026-semantic-versioning.md
arjun024: I added this on purpose as an exception to the breaking-change rule that could warrant a major bump. It could be better changed to `Change in official image URI`, the proposal here being - since consumption of buildpack will largely be driven through official image registry uri and not via .cnb downloads, consumers will have to make changes to their build systems to keep consuming the same buildpack.

text/0026-semantic-versioning.md
```diff
@@ -0,0 +1,109 @@
...
+### Intial Development
+
+A version of `0.0.x` is considered to be in initial development and may not be suitable for
+user consumption.
+
+## User Policy
+
+* **Patch** bump: User should update their buildpack/builder without hesitation.
+
+* **Minor** bump: User can update their buildpack/builder to avail new features. Nothing will break.
```
genevieve, line 89: For the "User Policy", I wonder if this statement is true "User can update their buildpack/builder to avail new features. Nothing will break."

Since one of the acceptable buildpack changes for a minor bump include:
> Addition/removal of dependency
(but default-version not changed to a version with breaking change)

I think this would easily break applications with specified dependency versions. For instance, Gemfiles specify the version of bundler to use (sometimes down to the patch) and if that dependency version is removed, the application will fail to build with that buildpacks latest minor update.

fg-j:
Currently, "Change in Buildpack API version" falls in the minor column. However, changing the buildpack API version could force users to upgrade the platform and/or lifecycle in order to have the buildpack continue to work properly. (The lifecycle may be backward-compatible, but they'd still need to upgrade the lifecycle.) Should this go in the major bump column for that reason?

text/0026-semantic-versioning.md
arjun024: Good point. The policy of "**Nothing** will break" on minor bumps might be a little too much of a claim. Also, major bump on every dependency update sounds like a lot of noise to the user. Maybe something along the lines of `User can update their buildpack/builder to avail new features and dependencies. Major lines of dependency versions won't be removed`? Removal of major line of versions can be added as an example to the major column in examples.

@kvedurmu Curious to hear if you have thoughts?

text/0026-semantic-versioning.md
ryanmoran: But that change isn't to the buildpack. The buildpack doesn't contain breaking changes. If you are already using the new registry, seeing a major version bump, for apparently no reason, would be really surprising. I agree with the intent to ensure folks get ample notification that the buildpack will be moving registry locations, but we shouldn't try to convey that through a version number.

arjun024:
@fg-j My thinking was that this change is abstracted away by the lifecycle that comes with the builder. But I think it's a fair point that given the same environment and buildpack bumps API version, there's no guarantee that their builder's lifecycle will support this API. I will move it to the major column.

text/0026-semantic-versioning.md
```diff
@@ -0,0 +1,109 @@
...
+| Implementation buildpack additions/removals/major bumps<br>that cause incompatible change |  |  |
+
+
+#### Builder
+| `Major`                 | `Minor`                   | `Patch`                   |
+|-------------------------|---------------------------|---------------------------|
+| Lifecycle major bumps   | Minor bumps of components | Patch bumps of components |
+| Stack Image major bumps |                           |                           |
+| Buildpack major bumps   |                           |                           |
+
```
sophiewigmore, line 79:
```suggestion
#### Stack
| `Major`                 | `Minor`                   | `Patch`                   |
|-------------------------|---------------------------|---------------------------|
| Removal of  a package   | Addition of a new package | Update of a package       |
```

@martyspiewak does this look accurate for what we would expect for stacks?

text/0026-semantic-versioning.md
arjun024: Sounds good. Removed the entry.

text/0026-semantic-versioning.md
```diff
@@ -0,0 +1,109 @@
...
+#### Buildpack
+| `Major` | `Minor` | `Patch` |
+|-|-|-|
+| \<Any incompatible change\> | \<Feature additions but compatible with previous release\> | \<Fixes but no major feature changes\> |
+| Change in Buildpack ID | Addition/removal of dependency<br>(when not superseded by conditions in the Major column) | Bug fixes |
+| Change in supported Stacks | Buildpack runs a new command the causes backward compatible<br>change in behavior (e.g. x-install buildpack adds running "x clean") | Performance improvement, internal tweaks |
+| Change in Buildpack API version | Change of file type of build/detect executable<br>(e.g. text to binary) | CVE fixes |
+| Change in API (Provides/Requires) | New configuration option (e.g. new environment variable support) | Change of required build tool version to package buildpack<br>(e.g. go version) |
+| Removal of configuration options<br>(e.g. buildpack.yml) | Change in <br>[buildpack-specific metadata](https://github.com/buildpacks/spec/blob/buildpack/v0.6/buildpack.md#build-plan-toml) |  |
+| Default version of dependency bumped to incompatible version<br>(e.g. php7 to php8) | Publishing to a new image registry |  |
```
divoxx, line 64: I think it would be useful to expand more on how dependency versions will be managed, specifically in the context of languages/compilers and language tooling, **where the dependency is the main thing being provided to the end-user and multiple versions are supported in parallel**.

For example, let's say the Ruby's MRI buildpack currently supports (hypothetically):

* `2.5.9`, `2.5.8`
* `2.6.7`, `2.6.6`, `2.6.5`
* `2.7.3`, `2.7.2`, `2.7.1`

How would the buildpack versioning change in the following cases?

* A new series of patch versions are introduced, adding support for `2.6.8` and `2.7.4`.
* A new minor version is introduced, adding support for `2.8.0`
* A new major version is introduced, adding support for `3.0.1`
* An older patch version is removed, removing support for `2.6.5`
* An older minor version is removed, removing support for all `2.5.X` series.

Here are the same cases with my own thoughts on them, but I'm sure there are things I'm overlooking, so take it with a grain of salt.

* A new series of patch versions are introduced, adding support for `2.6.8` and `2.7.4`.
  * This would probably be safe to release as a patch release, since it should be backwards compatible to the user even if they don't have a specific version set for the project, or if they are set as `2.7.*`.
* A new minor version is introduced, adding support for `2.8.0`
  *  This would probably be safe to release as a patch release IF the new version is not set as default version, since it should be backwards compatible to the user even if they don't have a specific version set for the project, or if they are set as `2.*`.
  * If the new version is the new default, then it should probably be a new minor release.
* A new major version is introduced, adding support for `3.0.1`
  *  This would probably be safe to release as a patch release IF the new version is not set as default version, since it should be backwards compatible to the user even if they don't have a specific version set for the project, or if they are set as `2.*`.
  * If the new version is the new default, then it should probably be a new major release.
* An older patch version is removed, removing support for `2.6.5`
  * This is probably more tricky since it can break an application which is depending on the specific version. In order to be fully compatible with semantic versioning, this might require a major release. That said, it might be OK to remove as a minor release if the expected best practice is to pin projects to `M.N.*` format instead of specific patch level.
* An older minor version is removed, removing support for all `2.5.X` series.
  * Requires major version release.

Similarly, buildpacks that provides toolings, such as Ruby`s bundler, could follow the same approach.

divoxx (commented):
Hi :wave: 

Sorry to jump into this conversation but I work with @genevieve and followed the link in the original [discussion she started](https://github.com/paketo-buildpacks/feedback/discussions/10) to this pull request.

We are currently in the process of defining our own versioning process for our custom builder and custom buildpacks, and being a user of paketo I wanted to contribute to the conversation as well, if possible.

text/0026-semantic-versioning.md
genevieve: > That said, it might be OK to remove as a minor release if the expected best practice is to pin projects to M.N.* format instead of specific patch level.

Is it recommended formally in any of the paketo docs that projects pin to minor instead of patch versions of dependencies? If not, this might be great to document.

text/0026-semantic-versioning.md
martyspiewak: Yup, looks right to me.

fg-j:
One thing I'd like to see in the RFC is a proposal for process that will allow maintainers to easily determine whether a release is a major, minor, or patch release when it's time to ship. One simple approach would be a requirement that each PR be labeled as a major, minor, or patch change AND some easy way to roll up this information for all PRs since the last release.

I think this is important both for implementation buildpacks and composite buildpacks. As discussed in the working group meeting on 4/13, I think the Java buildpacks currently have a system for this. cc @ekcasey and @dmikusa-pivotal

text/0026-semantic-versioning.md
```diff
@@ -0,0 +1,109 @@
...
+## Motivation
+
+As a Buildpack/Builder maintainer,<br/>
+I need a release versioning guideline,
+so I can make my release versions consistent.
+
+As a Buildpack/Builder user,<br/>
+I want to know how versioning is done,
+so I can use the appropriate version and
+decide when to upgrade.
```
dmikusa-pivotal, line 31: Not 100% sure it matters, cause I see plenty of reason to do this, but I would say another motivation is that "As a Buildpack/Builder user, I want versioning to be consistent across all Paketo buildpacks so I don't have to understand multiple policies."

text/0026-semantic-versioning.md
```diff
@@ -0,0 +1,109 @@
...
+>   - MINOR version when you add functionality in a backwards compatible manner, and
+>   - PATCH version when you make backwards compatible bug fixes.
+
+The proposal is to conform to the above semantics as much as possible.
+
+### Changes
+
+In terms of Paketo Buildpacks/Builders, this translates to:
+
+* Bumping MAJOR is analogous to an "`app or related buildpacks may not work without change`" relation between releases.
```
dmikusa-pivotal, line 49: >app or related buildpacks may not work without change

IMHO, for a major bump I think the mantra should be more like `app or related buildpacks is likely to not work without change`. 

In other words, if we change the major release odds are it won't just work, i.e. it's incompatible.

I think that's compatible with the examples you've listed, like removal of major dep versions, removal of config options or changing the stack id.

text/0026-semantic-versioning.md
```diff
@@ -0,0 +1,109 @@
...
+* Bumping PATCH is analogous to a "`fixes issue`" relation between releases.
+
+The following table lists the usual circumstances/events where bumping a particular component is expected:
+
+#### Buildpack
+| `Major` | `Minor` | `Patch` |
+|-|-|-|
+| \<Any incompatible change\> | \<Feature additions but compatible with previous release\> | \<Fixes but no major feature changes\> |
+| Change in Buildpack ID | Addition/removal of dependency<br>(when not superseded by conditions in the Major column) | Bug fixes |
+| Change in supported Stacks | Buildpack runs a new command the causes backward compatible<br>change in behavior (e.g. x-install buildpack adds running "x clean") | Performance improvement, internal tweaks |
```
dmikusa-pivotal, line 60: >Change in supported Stacks

I understand why the removal of a stack would trigger a major bump, but why would the addition of a stack? That seems like it could be under a minor version since it's just adding functionality and not likely to break something.

My thought would be to change this to "Removal of a stack or renaming of an existing stack". Then specifically add "Adding a stack to the "Minor" column, to indicate additions can be done under a minor bump.

text/0026-semantic-versioning.md
dmikusa-pivotal: > A new series of patch versions are introduced, adding support for 2.6.8 and 2.7.4.
> This would probably be safe to release as a patch release, since it should be backwards compatible to the user even if they don't have a specific version set for the project, or if they are set as 2.7.*.

+1

> A new minor version is introduced, adding support for 2.8.0
> This would probably be safe to release as a patch release IF the new version is not set as default version, since it should be backwards compatible to the user even if they don't have a specific version set for the project, or if they are set as 2.*.
> If the new version is the new default, then it should probably be a new minor release.

+1 

> A new major version is introduced, adding support for 3.0.1, 3.0.0
> This would probably be safe to release as a patch release IF the new version is not set as default version, since it should be backwards compatible to the user even if they don't have a specific version set for the project, or if they are set as 2.*.
> If the new version is the new default, then it should probably be a new major release.

+1

> An older patch version is removed, removing support for 2.6.5
> This is probably more tricky since it can break an application which is depending on the specific version. In order to be fully compatible with semantic versioning, this might require a major release. That said, it might be OK to remove as a minor release if the expected best practice is to pin projects to M.N.* format instead of specific patch level.

I think this should be able to be done in a patch release. No one should be pinning to a specific version, that's an anti-pattern we don't want to facilitate or encourage. In short, as an app dev I should care at all when a a patch release is dropped. 

From the author's standpoint, I also worry about this rule bumping minor versions unnecessarily. Because don't keep many patch releases, most new patch release are going to result in an old patch falling off and being removed. If that triggers a minor bump, the minor version is going to go up quick.

> An older minor version is removed, removing support for all 2.5.X series.
> Requires major version release.

My $0.02 is that this would only trigger a minor bump. IMHO, the goal of the build pack is that you're always on the latest release of a major branch. I should be able to say give me Ruby 2 or 3 or Java 8 or 11, with the rest being details. Thus if a new minor comes out, you should get that right away. In addition, the oldest minor should drop off, so we're keeping at most two. That gives a dev the ability to roll back one version if there's a problem, time to fix the problem, then get back up to the latest version. I think if you're following that path, the removal of the N-2 release, should be a non-event. I could almost see it being done in a patch release, but that is probably going too far.

dmikusa-pivotal: >Default version of dependency bumped to incompatible version

I don't really like the word "incompatible" here. Why not just say major? so when the default version of a dependency changes to a new major version, that triggers a major buildpack version bump.

text/0026-semantic-versioning.md
```diff
@@ -0,0 +1,109 @@
...
+| Change in Buildpack API version | Change of file type of build/detect executable<br>(e.g. text to binary) | CVE fixes |
```
dmikusa-pivotal, line 61: >Change of file type of build/detect executable

I'm curious why would this matter? If you change the build/detect file type, that's internal to the buildpack. What addition of functionality is this to merit a minor bump? It seems like an implementation detail only.

text/0026-semantic-versioning.md
dmikusa-pivotal: +1 I think this seems a little strong. You should be able to upgrade without problems if you're using default versions or following good hygiene (i.e. using a wildcard like `2.*`).

Personally, I don't think we should encourage version pinning to minor and definitely not patch releases, so to adjust this policy to accommodate someone pinning bundler to version 1.2.3 seems like a bad idea to me. Buildpacks should promote good version hygiene, like using the latest versions and if you're following that process you should be rewarded with frustration-free upgrades. If you decided to pin your minor/patch versions, then you opt-in to more work to do additional testing and maintenance on buildpack updates. My $0.02.

text/0026-semantic-versioning.md
divoxx: > My $0.02 is that this would only trigger a minor bump. IMHO, the goal of the build pack is that you're always on the latest release of a major branch. I should be able to say give me Ruby 2 or 3 or Java 8 or 11, with the rest being details.

Unfortunately, at least in my experience, this is almost never true. Most companies that I'm aware tend to pin to a specific minor version, because there are many minor changes that can lead to a unexpected change in behavior. Specifically, while minor versions don't usually break APIs, it introduces new features, refactoring and implementation changes, as well as changes to the runtime such as GC, or upgrade to linked dependencies, etc. Those can lead to unintended production consequences.

I definitely see this as a pain for the buildpack authors, though. Having to bump major in order to drop a minor release would/is likely going to be painful, but not doing so would lead to a breaking change if the downstream user has pinned a minor version, or if the language's minor release has added a unintentional change in behavior.

arjun024: I added a separate table for dependency updates - so please take a look.

> I don't really like the word "incompatible" here. Why not just say major?

That language was to not assume that upstream dependencies always follow the same versioning semantics. An example would be `pipenv` who uses ` YYYY.MM.DD` based on release date. ([Pipenv version history](https://pipenv.pypa.io/en/latest/changelog/)) The onus is on maintainers to determine if the new version is incompatible.

> An older minor version is removed, removing support for all 2.5.X series.
Requires major version release.

I think this strategy will probably work for something like `mri` that makes a minor release [about once a year](https://www.ruby-lang.org/en/downloads/releases/), but not for something like `node` that makes [multiple minor bumps a month](https://nodejs.org/en/download/releases/). IMO releasing a large number of major version releases for node could cause confusion and resistance to update among users. But due to this difference, I'd expect node's default version to be constrained to `major` and that of mri to be constrained to `major.minor`. So probably the semantics should be tied to if that removal also causes a change of buildpack's default version.

text/0026-semantic-versioning.md
arjun024: It was meant as a compatible change in the perspective of a platform author. I guess if this is abstracted away by lifecycle binaries, I can take it out

text/0026-semantic-versioning.md
arjun024: Added. Thanks

text/0026-semantic-versioning.md
dmikusa-pivotal: @arjun024 - Table looks good! Thanks

text/0026-semantic-versioning.md
ryanmoran: Although the `Gemfile.lock` does show a version number that is specific down to the patch, Bundler itself is loose with the specifics of what versions can be used. The `bundler` buildpack will install the latest version of Bundler matching the same major version as what is specified in the `Gemfile.lock`: https://github.com/paketo-buildpacks/bundler/blob/498cc180ad03991d79aa9fbf0ec2ec3391d144c6/gemfile_lock_parser.go#L18-L43. So, unless you are explicitly pinning a specific version number using either `buildpack.yml` or `BP_BUNDLER_VERSION`, you should not see a breaking change in this case.

text/0026-semantic-versioning.md
divoxx: @arjun024 that's a really good point. I was mainly thinking of Ruby and Go buildpacks in my use cases. That table looks good to me. Thanks!

arjun024:
> One thing I'd like to see in the RFC is a proposal for process that will allow maintainers to easily determine whether a release is a major, minor, or patch release when it's time to ship. One simple approach would be a requirement that each PR be labeled as a major, minor, or patch change AND some easy way to roll up this information for all PRs since the last release.

I added a Process section. Please take a look.

dmikusa-pivotal:
> > One thing I'd like to see in the RFC is a proposal for process that will allow maintainers to easily determine whether a release is a major, minor, or patch release when it's time to ship. One simple approach would be a requirement that each PR be labeled as a major, minor, or patch change AND some easy way to roll up this information for all PRs since the last release.
> 
> I added a Process section. Please take a look.

In terms of:

>semver/major - denoting accepting the change will require a major version bump in the next release.
>semver/minor - denoting accepting the change will require a minor version bump in the next release.
>semver/patch - denoting accepting the change will require a patch version bump in the next release.

The JBP team is using `semver:major`, `semver:minor` and `semver:patch`. Basically just a `:` instead of `/`. Can we use that?

Or would it be easier to not specify the exact label in the RFC and leave that as an implementation detail? The RFC could instead say something like you need to have a label for each type major, minor and patch but not what the label should be.

text/0026-semantic-versioning.md
```diff
@@ -0,0 +1,189 @@
...
+  [paketo-bot](https://github.com/paketo-bot)) must be tagged with either of
+  the 3 labels:
+	* `semver/major` - denoting accepting the change will require a major
+	  version bump in the next release.
+	* `semver/minor` - denoting accepting the change will require a minor
+	  version bump in the next release.
+	* `semver/patch` - denoting accepting the change will require a patch
+	  version bump in the next release.
+
+* When releasing the next version, the maintainer must pick the most
```
fg-j, line 174: The way this is currently worded, it sounds like the onus is on maintainers to manually figure out which PRs will go into the release and manually set the correct version. It'd be great if the draft release automation did this for us. Not sure if it's overly prescriptive to lay that out here, but perhaps the project would benefit from having some codified process around this.

Seems like the [Java buildpacks](https://github.com/paketo-buildpacks/libjvm/blob/main/.github/workflows/update-draft-release.yml) are currently using [this release-drafter Github Action](https://github.com/release-drafter/release-drafter) which is highly configurable. Features include: automatic version resolution based on PR labels, automatic changelog generation.

text/0026-semantic-versioning.md
```diff
@@ -0,0 +1,189 @@
...
+## Process
+
+This section proposes a process that will help maintainers easily determine
+whether a given change or a set of changes require a bump in the major, minor,
+or patch component.
+
+* Changes should be accepted to the main line of code only via Pull Requests.
+
+* All Pull Requests (including automated PRs by the
+  [paketo-bot](https://github.com/paketo-bot)) must be tagged with either of
```
fg-j, line 165: Also not sure if this gets too in the weeds, but I want to call out that implementing this RFC will require new automation that figures out how to correctly label bot PRs.

Some different cases that will need to be implemented:
- a dependency update in an implementation buildpack's `buildpack.toml` (e.g. `go-dist` loses go 1.15, gains go 1.17)

- an implementation buildpack update in a composite `buildpack.toml` (e.g. paketo-buildpacks/go-build v3.1 replaces v3.0 in the paketo-buildpacks/go buildpack)

- lifecycle version is updated in a builder
- build image version is updated in a builder
- buildpack version is updated in a builder

text/0026-semantic-versioning.md
```diff
@@ -0,0 +1,189 @@
...
+
+| Dependency change | Buildpack Version expectation |
+|-|-|
+| ADD Patch | Bump Patch |
+| ADD Minor | - Bump Minor if set as default<br>  version<br>- Bump Patch otherwise |
+| ADD Major | - Bump Major if set as default<br>  version<br>- Bump Minor otherwise |
+| REMOVE Patch | Bump Patch |
+| REMOVE Minor | - Bump Major if unset as default<br>  version<br>- Bump Minor otherwise |
+| REMOVE Major | Bump Major |
+<br/>
```
ekcasey, line 102: I am finding the add/remove+default semantics a bit hard to follow. Maybe we should separating the defaulting behavior from the add/remove. i.e.
- Add
  - patch -> patch
  - minor -> minor
  - major -> minor
Remove
  - patch-> patch
  - minor -> minor
  - major -> major
- change default:
   - patch (e.g. 1.0.0 -> 1.0.1) -> patch
   - minor (e.g. 1.0.1 -> 1.1.0) -> minor
   - major (e.g. 1.1.0 -> 2.0.0) -> major

Sometimes I do two things at the same time (e.g. replace a default dependency with the next patch version of the same dependency). But I could also theoretically change the default without adding or removing anything.

text/0026-semantic-versioning.md
```diff
@@ -0,0 +1,189 @@
...
+
+
+#### Composite Buildpack
```
ekcasey, line 105: I would propose that we add:
Major - Removal of a component buildpack
Minor - Addition of a component buildpack

text/0026-semantic-versioning.md
```diff
@@ -0,0 +1,189 @@
...
+#### Stack
+| `Major`                 | `Minor`                   | `Patch`                   |
+|-------------------------|---------------------------|---------------------------|
+| Removal of  a package   | Addition of a new package | Update of a package       |
+
+### Intial Development
+
+A version of `0.0.x` is considered to be in initial development and may not be
+suitable for user consumption.
+
```
ekcasey, line 136: Might be worth calling out that for alpha versions (0.0.x) all changes are patchs and for beta versions (0.x.y) breaking changes (listed as major versions above) will be minor bumps.

text/0026-semantic-versioning.md
```diff
@@ -0,0 +1,189 @@
...
+| REMOVE Patch | Bump Patch |
+| REMOVE Minor | - Bump Major if unset as default<br>  version<br>- Bump Minor otherwise |
+| REMOVE Major | Bump Major |
+<br/>
+
+
+#### Composite Buildpack
+| `Major` | `Minor` | `Patch` |
+|-|-|-|
+| Implementation buildpack major bumps | Implementation buildpack minor bumps | Implementation buildpack patch bumps |
```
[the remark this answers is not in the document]
arjun024: How about `Implementation buildpack major bumps causing change in external behavior`?

text/0026-semantic-versioning.md
arjun024: Done

text/0026-semantic-versioning.md
arjun024: Done

text/0026-semantic-versioning.md
arjun024: Fixed

text/0026-semantic-versioning.md
arjun024: I'd say detailing the implementation of this process is out of scope for this RFC.

text/0026-semantic-versioning.md
```diff
@@ -0,0 +1,189 @@
...
+
+The following table lists the usual circumstances/events where bumping a particular component is expected:
+
+#### Buildpack (General)
+| `Major` | `Minor` | `Patch` |
+|-|-|-|
+| \<Any incompatible change\> | \<Feature additions but compatible with previous release\> | \<Fixes but no major feature changes\> |
+| Change in Buildpack ID | Support for a new stack | Bug fixes |
+| Change in Buildpack API version | Buildpack runs a new command the causes backward compatible<br>change in behavior (e.g. x-install buildpack adds running "x clean") | Performance improvement, internal tweaks |
+| Change in API (Provides/Requires) | New configuration option (e.g. new environment variable support) | CVE fixes |
```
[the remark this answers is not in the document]
arjun024: Done

text/0026-semantic-versioning.md
```diff
@@ -0,0 +1,189 @@
...
+| Removal of a stack or renaming of an existing stack | Change in <br>[buildpack-specific metadata](https://github.com/buildpacks/spec/blob/buildpack/v0.6/buildpack.md#build-plan-toml) | Change of required build tool version to package buildpack<br>(e.g. go version) |
```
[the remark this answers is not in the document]
arjun024: Done

text/0026-semantic-versioning.md
```diff
@@ -0,0 +1,189 @@
...
+| Removal of configuration options<br>(e.g. buildpack.yml) |  |  |
```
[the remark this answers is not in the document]
arjun024: Done

text/0026-semantic-versioning.md
```diff
@@ -0,0 +1,189 @@
...
+
+For buildpacks whose primary function is to provide dependencies, the following versioning is expectated
+when versions of dependencies are added or removed. The table below assumes that upstream dependencies
+follow [semantic versioning](https://semver.org/). For those that do not, maintainers may determine
+their versioning independently.
+
+| Dependency change | Buildpack Version expectation |
+|-|-|
+| ADD Patch | Bump Patch |
+| ADD Minor | - Bump Minor if set as default<br>  version<br>- Bump Patch otherwise |
```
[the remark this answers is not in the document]
arjun024: Done

text/0026-semantic-versioning.md
arjun024: Done

text/0026-semantic-versioning.md
arjun024: My idea was that - 0.0.x do not follow any of these guidelines as they are not yet advertised as suitable for general consumption, and when you release 0.1.0 or higher you will abide by these semver guidelines, i.e. an immediate breaking change after 0.1.0 should be 1.0.0. Happy to change this per your comment if you feel strongly

text/0026-semantic-versioning.md
arjun024: I see a similar case addressed at https://github.com/paketo-buildpacks/rfcs/pull/49#discussion_r615011293. I'll separate them out.

text/0026-semantic-versioning.md
```diff
@@ -0,0 +1,201 @@
...
+	  version bump in the next release.
+	* `semver:patch` - denoting accepting the change will require a patch
+	  version bump in the next release.
+
+* When releasing the next version, the maintainer must pick the most
+  significant required version bump from among the accepted Pull Requests and
+  bump that component of the release. The order of significance is `major` >
+  `minor` > `patch`.
+
+* Pull requests with critical security updates must be additionally tagged as
```
fg-j, line 191: Under our current branch management and PR process, tagging something `security-fix` and immediately releasing as a patch could roll in changes on `main` that warrant a minor or major bump.

Is this RFC advocating for a new branch management process where PRs are merged into major/minor/patch branches?

arjun024: No, this part is simply standardizing a label for critical security fixes. See User Policy section - If a user is on the latest Major and Minor versions, critical security updates when they occur should be available for users on the latest Patch version.
Assume there is a PR that removes PHP 7.5 and adds PHP 7.6. This would be naturally labeled as `semver:minor`. But then if it is known that PHP 7.5 has a security vulnerability that is fixed by 7.6, then the maintainer must release this as a patch due to the user policy commitment stated above. In such a case, Pull Request creator or whoever responsible can tag it as `security-fix` so that a patch can be released.

text/0026-semantic-versioning.md
arjun024: @ekcasey (friendly ping to see if this has to be changed)

ekcasey: `1.0.0` indicates a level of maturity and readiness for production use cases. I wouldn't want us to move to `1.0.0` just because we wanted to make a breaking change. Yes we could use 0.0.x for this but 0.0.x basically communicates "experimental/unstable do not use". Whereas an "0.x.y" version is more stable and robust ready for feedback from early adopters.

Semver docs seem to align with making braking changes in an `0.y.0` version:

>Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

ekcasey (approved):

ryanmoran:
@arjun024 this looks good to me. Can you rebase this off of `main` and include a modification to the `README.md` in the section for [Accepted RFCs](https://github.com/paketo-buildpacks/rfcs#accepted-rfcs)?

ryanmoran (approved):

PR #50, opened 2021-04-15
RFC: Standardize log level configuration

ryanmoran:
## Summary

[Readable](https://github.com/paketo-buildpacks/rfcs/blob/log-level-rfc/text/0026-log-levels.md)

## Checklist


text/0026-log-levels.md
```diff
@@ -0,0 +1,57 @@
...
+needs is not possible. Instead, we should allow buildpack users to tune the log
+output to suite their needs.
+
+## Detailed Explanation
+
+All buildpacks will provide a `BP_LOG_LEVEL` environment variable to allow for
+the configuration of log output. This variable will support 3 acceptable
+options: `WARN`, `INFO`, and `DEBUG`. The default option, should none be
+provided by the user, will be `INFO`. While this RFC will not outline detailed
+specifics for what logs should appear at each level, a basic description of
```
sophiewigmore, line 28: I agree that the specifics of what should be logged shouldn't be covered here, but maybe there should be some standardization across what the formats should look like at some point. It's a bit early for this, but it would be nice to have this laid out somehow so that if it is implemented in both `packit` and `libpak` for example, we have a paper trail of what that agreed format should be. 

The logging across some of the Java buildpacks is already different from the logging in other buildpacks in terms of color output (for example, see the CA Certificate Buildpack logs when it runs in the Node language family) it would be nice if we could minimize any other differences ahead of time.

sophiewigmore:
I love this idea, I think it would be really helpful for users. We've seen some user interest in [this issue](https://github.com/paketo-buildpacks/bundle-install/issues/104), and I think it would play nicely with the [pack interaction mode RFC](https://github.com/buildpacks/rfcs/pull/152)

text/0026-log-levels.md
ryanmoran: This RFC is meant to be a start for negotiating what a common log output might look like for all of the buildpacks. I want this RFC and others that follow on this path to be small so that we aren't debating some massive omnibus RFC that overhauls the entirety of log output. I think we all agree that we would like to get to a common log output, but it will be quite a bit of work. I'd like to see us break the problem down into reasonable pieces so that we can iterate towards a solution.

kvedurmu (commented):
This looks awesome. One question: in the current proposal could you define what level of log output this applies to? As in, if I set `BP_LOG_LEVEL=DEBUG` will this provide debug output into the buildpack process (build commands, detection, etc.) as well as debug output for the underlying language command being run (e.g `NPM_LOG_LEVEL=DEBUG npm install`)?

ryanmoran:
> This looks awesome. One question: in the current proposal could you define what level of log output this applies to? As in, if I set `BP_LOG_LEVEL=DEBUG` will this provide debug output into the buildpack process (build commands, detection, etc.) as well as debug output for the underlying language command being run (e.g `NPM_LOG_LEVEL=DEBUG npm install`)?

This only impacts the log level of the buildpack itself. The tools that are invoked by some of the buildpacks should be invoked in such a way that their configuration settings can be passed through by the buildpack user. So, for your example, the user would provide both `BP_LOG_LEVEL=DEBUG` and `NPM_LOG_LEVEL=DEBUG` to see the debug output of both the buildpack itself and the `npm` tool.

kvedurmu:
> So, for your example, the user would provide both `BP_LOG_LEVEL=DEBUG` and `NPM_LOG_LEVEL=DEBUG` to see the debug output of both the buildpack itself and the `npm` tool.

Got it, this makes sense. I like that this is defined in scope. I can push my ideas in a separate RFC/discussion thread 😀 .

ryanmoran:
After the discussion in WG this week, I've decided that it makes more sense for this RFC to omit the definition of a `WARN` log level as I felt most of the RFC was becoming a definition of what does or does not appear at that level. Instead, we will limit the current set of levels to `INFO` and `DEBUG` which covers what is already supported in the existing buildpacks while leaving the definition of `WARN` for an RFC to follow.

ekcasey (approved):

PR #53, opened 2021-04-21
RFC: Paketo Buildpackless Builders

fg-j:
[Rendered](https://github.com/fg-j/rfcs/blob/2ee644443bb45969f83af24032b87058f2657bc9/text/0030-buildpackless-builders.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@
 * [0026: Environment Variable Configuration Of Buildpack](./text/0026-environment-variable-configuration-of-buildpacks.md)
 * [0027: Common Logging Levels for Buildpacks](./text/0027-log-levels.md)
 * [0029: Semantic Versioning of Buildpacks and Builders](./text/0029-semantic-versioning.md)
+* [0030: Buildpackless Builders](./text/0030-buildpackless-builders.md)
 
 ## Implemented RFCs
 
--- a/text/0030-buildpackless-builders.md
+++ b/text/0030-buildpackless-builders.md
@@ -1,0 +1,313 @@
+# Buildpackless Builders
+
+## Summary
+
+Paketo builders are a way of distributing Paketo buildpacks and stacks along
+with the CNB lifecycle. Currently, Paketo distributes the Full, Base and Tiny
+builders, which include different sets of buildpacks and whose stacks each
+have different mixins. While they are a simple way of packaging all the
+necessary components for a Paketo build, they often include more than a user
+actually needs. This RFC proposes that Paketo publish buildpackless versions of its
+three builders. These will include the same build- and run-images as their
+standard counterparts, but will include **neither buildpacks nor orders**. This will
+speed up builds by reducing the number of unnecessarily-pulled layers, without
+forcing Paketo users to maintain their own builders if they want to curate the
+buildpacks that are included in a build.
+
+## Motivation
+
+Paketo currently distributes the
+[full](github.com/paketo-buildpacks/full-builder),
+[base](github.com/paketo-buildpacks/base-builder), and
+[tiny](github.com/paketo-buildpacks/tiny-builder) builders. These each include
+a set of buildpacks, a build- and run-image, and a version of the CNB
+lifecycle. The most recently-released version of the Full Builder (v0.1.130)
+contains 106 layers. These include layers for the build- and run-images and
+layers for the included buildpacks. If a user builds their app with `pack`
+using the Full Builder, all 106 layers of the builder must be pulled onto the
+local Docker daemon before the build can begin. As many Paketo users and
+maintainers can likely attest, pulling such a large builder can be
+time-consuming.
+
+What's more, users rarely need all of the buildpacks published in the builder
+for their build. An app dev user likely knows which language family their app
+belongs to and can select individual buildpacks accordingly. An operator user
+likely knows which subset of buildpacks are needed for the developers they work
+with. Thus, for most users, the Paketo builders offer more language family
+buildpacks than will be used. Users experience a time cost with no appreciable
+benefit. To improve their build times, users might maintain their own builders
+that only include the buildpacks they require. Paketo's jam CLI has a `jam
+update-builder` command that facilitates keeping a `builder.toml` up to date
+with the latest versions, but users must still set up automation to
+periodically run the builder update. This is overhead that many users may not
+want to take on.
+
+Within Paketo's testing, buildpack test suites frequently use the base builder,
+but overwrite the buildpacks included in the builder using the `--buildpack`
+option for pack (more specifically, using the corresponding `WithBuildpacks()`
+configuration in occam). In this scenario, the entire builder image is pulled
+onto the test runner's daemon – and then all of the pulled buildpacks are
+promptly ignored. Pulling unnecessary layers slows down Paketo's tests.
+
+This RFC proposes that Paketo publish builders that package up build- and
+run-images with the lifecycle but exclude buildpacks and orders. These builders
+would be significantly smaller than the existing builders, leading to shorter
+builder image pull times.
+
+Since the CNB spec currently [supports including `[[ buildpacks ]]` in a
+project file
+(`project.toml`)](https://buildpacks.io/docs/reference/config/project-descriptor/),
+users can specify the buildpack(s) for their build in an easily stored and
+reproducible way. Users can, of course, also specify which buildpacks to
+include in the build with the `--buildpack` flag for pack.
+
+The proposed new builders will thus speed up users' builds and Paketo's
+buildpack tests while relying only upon the existing functionality that the
+pack CLI provides.(No upstream contributions required.)
+
+
+## Detailed Explanation
+
+### Definitions
+- **buildpackless builder** : A
+  [builder](https://buildpacks.io/docs/concepts/components/builder/) that
+  includes a build-image, a reference to a run-image, and a lifecycle, _but no
+  buildpacks nor order groups_.
+- **standard builder**: A builder that includes buildpacks and order groups in
+  addition to the build-image, lifecycle, and reference to a run-image.
+
+### Calculating Builder Size
+
+To determine the number of layers in a given builder image, one can use the
+`docker history` command. Its output shows one line per layer. For example, to
+calculate the number of layers in the latest Paketo Full builder, simply run:
+```
+docker pull paketobuildpacks/builder:full
+docker history -q paketobuildpacks/builder:full | wc -l
+```
+
+### What will the Paketo Buildpackless Builders look like?
+The `buildpackless-builder.toml` for a given Paketo Buildpackless <Name> Builder will be identical
+to the corresponding standard Paketo <Name> Builder, but will have the `[[
+buildpacks ]]` and `[[ order ]]` sections removed. For instance, the Buildpackless
+Full Builder will have a `buildpackless-builder.toml` that looks like:
+
+```
+description = "<some description that clearly states buildpacks must be specified with this builder>"
+
+[lifecycle]
+  version = "X.Y.Z"
+
+[stack]
+  build-image = "docker.io/paketobuildpacks/build:X.Y.Z-full-cnb"
+  id = "io.buildpacks.stacks.bionic"
+  run-image = "index.docker.io/paketobuildpacks/run:full-cnb"
+  run-image-mirrors = ["gcr.io/paketo-buildpacks/run:full-cnb"]
+```
+
+The above stated change results in a ~88% decrease in the number of layers
+as compared to Paketo Full Builder v0.1.130.
+
+### How will the Paketo Buildpackless Builders be distributed?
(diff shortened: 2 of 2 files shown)


text/0000-minimal-builders.md
```diff
@@ -0,0 +1,246 @@
...
+  run-image-mirrors = ["gcr.io/paketo-buildpacks/run:full-cnb"]
+```
+
+The above stated change results in a ~88% decrease in the number of layers
+as compared to Paketo Full Builder v0.1.130.
+
+### How will the Paketo Minimal Builders be distributed?
+Paketo will maintain and publish minimal builders that correspond to each of
+its three existing standard builders. Their config files (`builder.toml`) will
+be checked in to their own repos (paketo-buildpacks/minimal-full-builder,
```
arjun024, line 113: Any need to maintain them separately if one is a subset of the other?
How about for e.g. the full-builder also publishes the full-builder-minimal with the same version every time it is published, but minus [order]?

fg-j: I'm not opposed to this. Definitely reduces repo overhead. I wonder if it'd be confusing for the version to be bumped with no change to the actual builder (in the event of a buildpack update).

text/0000-minimal-builders.md
```diff
@@ -0,0 +1,246 @@
...
+## Summary
+
+Paketo builders are a way of distributing Paketo buildpacks and stacks along
+with the CNB lifecycle. Currently, Paketo distributes the Full, Base and Tiny
+builders, which include different sets of buildpacks and whose stacks each
+have different mixins. While they are a simple way of packaging all the
+necessary components for a Paketo build, they often include more than a user
+actually needs. This RFC proposes that Paketo publish "minimal" versions of its
+three builders. These will include the same build- and run-images as their
+standard counterparts, but will include **no buildpacks nor orders**. This will
```
ekcasey, line 12:
```suggestion
standard counterparts, but will include **neither buildpacks nor orders**. This will
```
apologies for the pedantry

text/0000-minimal-builders.md
```diff
@@ -0,0 +1,246 @@
...
+1. Make upstream contributions to pack that speed up builder pull times
+   - Benefits:
+     - Addresses the platform's problem of slow build times more directly. Will
+      benefit performance for users with _all_ builders.
+   - Drawbacks:
+     - Requires upstream contribution that may require an RFC. Optimizing image
+    pulls from a registry does not necessarily fall in scope of the Paketo
+    project currently. May require Paketo contributors to build up expertise
+    that is tangential to Paketo's main aims.
+
```
ekcasey, line 193: Another alternative would be to publish a builder for each language family. 

Benefits:
- Faster build times for end users because pack doesn't need to spend time constructing a builder image on the fly.

Drawbacks:
- Slower for Paketo contributors during development and test. When using a development version of a buildpack, paketo developers must still wait for layers containing the published version to be pulled.

text/0000-minimal-builders.md
```diff
@@ -0,0 +1,246 @@
...
+- instead of minimal builders, publish builders with curated sets of commonly-combined buildpacks (e.g. builders for each language family buildpack in isolation)
+--->
+
+## Implementation
+
+### Repositories
+Three new repositories should be created, for each of the new minimal builders:
+- paketo-buildpacks/minimal-full-builder
+- paketo-buildpacks/minimal-base-builder
+- paketo-buildpacks/minimal-tiny-builder
```
ekcasey, line 208: I wonder if we should call these empty builders or something else instead of minimal (empty still isn't great... naming is hard!). To me, the word minimal in this context evokes minimal images like tiny and might not convey that these builders are devoid of buildpacks.

text/0000-minimal-builders.md
```diff
@@ -0,0 +1,246 @@
...
+buildpacks are compatible with which builders (given that the
+`builder.toml`will not otherwise indicate this).
+
+<!---
+## Prior Art
+
+{{This section is optional if there are no actual prior examples in other tools.}}
+{{Discuss existing examples of this change in other tools, and how they've addressed various concerns discussed above, and what the effect of those decisions has been.}}
+--->
+
```
ekcasey, line 237: Would we want `pack builder suggest` to include these builders. Maybe not because they would provide a surprising/broken first build experience, but also it could help more users become aware of them.

text/0000-minimal-builders.md
ryanmoran: I had the same thought. Maybe something like `buildpackless` or just `packless`.

dmikusa-pivotal: +1 empty was the first word that came to mind but probably isn't specific enough. If we can spare the characters, I'd prefer something more specific like `no-buildpacks` or `buildpackless`.

text/0000-minimal-builders.md
dmikusa-pivotal: My vote would be to not include them, or break them into an "Advanced" grouping or something so they are set apart in the output. If it's not clearly marked, I agree it could be confusing for new users.

text/0000-minimal-builders.md
fg-j: I'm also interested in this angle for Paketo. I don't think the buildpackless and language-family-specific builders are mutually exclusive. Before we invest in builders for language families, I'd like to learn more about how Paketo users are interacting with builders. 

Are they using our builders for initial proof of concept? Are they creating and maintaining their own once they begin to incorporate Paketo into their workflows? Are they moving away from builders entirely as they move toward `kpack` and ClusterBuilders?

Perhaps folks like @genevieve or @Adam262 have thoughts about this from their own experiences.

For now, I think I'll keep the scope of this RFC focused on whether or not buildpackless builders are a good idea.

text/0000-minimal-builders.md
fg-j: If we wanted these builders surfaced in `pack builder suggest`, what would be the process for achieving that? An RFC elsewhere?

text/0000-minimal-builders.md
```diff
@@ -0,0 +1,246 @@
...
+builders, which include different sets of buildpacks and whose stacks each
+have different mixins. While they are a simple way of packaging all the
+necessary components for a Paketo build, they often include more than a user
+actually needs. This RFC proposes that Paketo publish "minimal" versions of its
+three builders. These will include the same build- and run-images as their
+standard counterparts, but will include **no buildpacks nor orders**. This will
+speed up builds by reducing the number of unnecessarily-pulled layers, without
+forcing Paketo users to maintain their own builders if they want to curate the
+buildpacks that are included in a build.
+
```
dmikusa-pivotal, line 16: Are there any numbers on how much of a tangible impact this change could make? 

For example, if it reduces the image size by X, and a pipeline is pulling Y times per day on avg, we can estimate a savings of Z by switching them to use a minimal builder like this. Similarly, do we have a comparison of the execution time for a typical pipeline job using the current builders versus a demo of the proposed builders?

I think it would be helpful to gauge the impact in terms of saved bandwidth, time, cost, carbon footprint, or something like that, given that there will be some additional overhead for maintaining and running pipelines to distribute the additional builders.

text/0000-minimal-builders.md
Adam262: Hey! My experience was I used the Paketo builder for a POC, but then found it was pretty trivial to make our existing custom builder work with the upstream pack by tweaking its stack to reference `"io.buildpacks.stacks.bionic"`.

text/0000-minimal-builders.md
fg-j: I switched the RFC to this approach.

text/0000-buildpackless-builders.md
```diff
@@ -0,0 +1,317 @@
...
+the builder.
+
+The buildpackless builders will be versioned the same as their standard
+counterparts. For each standard builder released and published, there will be a
+corresponding buildpackless builder with the same version. Some buildpackless
+builders with different versions may be identical. For instance, if a standard
+builder release upgrades only the version of a buildpack, the corresponding
+buildpackless builder release will have no changes.
+
+### How will users interact with Paketo Minimal Builders?
```
ForestEckhardt, line 127:
```suggestion
### How will users interact with Paketo Buildpackless Builders?
```

text/0000-minimal-builders.md
dmikusa-pivotal: The information you added is excellent. Thanks

dmikusa-pivotal (approved):

text/0000-minimal-builders.md
ekcasey: If we wanted it, I think we could make a PR to to pack directly.

text/0000-buildpackless-builders.md
```diff
@@ -0,0 +1,317 @@
...
+Paketo will maintain and publish buildpackless builders that correspond to each
+of its three existing standard builders. Their config files (`buildpackless-builder.toml`)
+will be checked in to the same repos as the standard builders
+(paketo-buildpacks/full-builder, etc.).  They'll be pushed to Dockerhub and GCR
+using tags `paketobuildpacks/builder:buildpackless-full` for the latest builder
+and `paketobuildpacks/builder:buildpackless-full-1.2.3` for version 1.2.3 of
+the builder.
+
+The buildpackless builders will be versioned the same as their standard
+counterparts. For each standard builder released and published, there will be a
```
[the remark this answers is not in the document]
fg-j: @ekcasey Check out the updated Implementation section. We can get crafty with automation and maintain 2 version lines on one repo without too much trouble.

text/0000-buildpackless-builders.md
```diff
@@ -0,0 +1,330 @@
...
+### How will the Paketo Buildpackless Builders be distributed?
+Paketo will maintain and publish buildpackless builders that correspond to each
+of its three existing standard builders. Their config files
+(`buildpackless-builder.toml`) will be checked in to the same repos as the
+standard builders (paketo-buildpacks/full-builder, etc.).  They'll be pushed to
+Dockerhub and GCR using tags `paketobuildpacks/builder:buildpackless-full` for
+the latest builder and `paketobuildpacks/builder:1.2.3-buildpackless-full` for
+version 1.2.3 of the builder. This parallels the [current tagging
+syntax](https://hub.docker.com/r/paketobuildpacks/builder/tags?page=1&ordering=last_updated)
+of `paketobuildpacks/builder:full` and `paketobuildpacks/builder:1.2.3-full`.
```
ekcasey, line 120: On first read I thought this conflicted with the plan to version them separately (the paragraph below) but I realize `paketobuildpacks/builder:1.2.3-buildpackless-full` is referrring the buildpackless builder version not the full builder version. Maybe we can update this to make it less ambigous? I misinterpreted which builder 
> `paketobuildpacks/builder:1.2.3-buildpackless-full` for
version 1.2.3 of the builder.

 was referring to.

ekcasey (approved):

text/0000-buildpackless-builders.md
```diff
@@ -0,0 +1,330 @@
...
+
+
+## Unresolved Questions and Bikeshedding
+
+- Does publishing buildpackless builders encourage users to interact with Paketo
+components in non-recomended ways?
+- Are people more likely to run into issues with build reproducibility or up-to-date-ness?
+- Will this make Paketo builders/buildpacks harder to support and troubleshoot?
+
+{{REMOVE THIS SECTION BEFORE RATIFICATION!}}
```
ryanmoran, line 330:
```suggestion
```

text/0000-buildpackless-builders.md
```diff
@@ -0,0 +1,330 @@
...
+co-located standard `builder.toml`).
+
+<!---
+## Prior Art
+
+{{This section is optional if there are no actual prior examples in other tools.}}
+{{Discuss existing examples of this change in other tools, and how they've addressed various concerns discussed above, and what the effect of those decisions has been.}}
+--->
+
+
```
ryanmoran, line 322:
```suggestion
```

ryanmoran (changes requested):
This looks great. I've asked for a couple of cleanup changes. I'm concerned that the release process seems pretty complicated, but I don't think that should block merging this RFC. I'll approve as soon as we remove these straggling comment sections from the RFC text.

ryanmoran (approved):

ryanmoran:
@fg-j before I merge this, can you add a link to the RFC text in the README under the ["Accepted RFCs" section](https://github.com/paketo-buildpacks/rfcs#accepted-rfcs)?

ryanmoran (approved):

ryanmoran:
@fg-j Sorry to be such a pain, but would you mind rebasing and squashing all of these commits so that I can make a clean, linear merge?

fg-j:
All tidied up now. @ryanmoran It would probably save future back and forth if there's a checklist somewhere for RFC authors to follow for "grooming" RFC PRs before merge.

ryanmoran (approved):

PR #54, opened 2021-04-25
RFC: Triage Party

andymoe:
## Summary

An RFC advocating for the deployment of and instance of the [Triage Party](https://github.com/google/triage-party) tool to `triage.paketo.io`

## Use Cases

See RFC

### DIFF
--- a/text/0026-triage-party.md
+++ b/text/0026-triage-party.md
@@ -1,0 +1,61 @@
+# Deploy and customize triage party for Paketo community use
+
+## Summary
+
+Host an instance of [Triage Party](https://github.com/google/triage-party) at `triage.paketo.io` to asist with issue triage, priortization and to keep an eye on community health using basic metrics like average issue response time and number of open PRs across all of Paketos repositories.
+
+## Motivation
+
+Keeping on top of all open issues, PRs and comments from the community and deciding which potential work to priortize is challenging. There is not an easy way to get visibility into the health of the community in terms of the age of open issues, issues requiring a response, and PRs awaiting review across all repositories.
+
+## Detailed Explanation
+
+The [Paketo project](https://paketo.io) is made up of many git repositories across the [paketo-buildpacks](https://github.com/paketo-buildpacks) and [paketo-community](https://github.com/paketo-community) Github orgs. These repositories contain implementation buildpacks, language family buildpacks, and auxiliary repositories related to managing the Paketo project and its community. **As of late April, 2021, there are 113 repositories across Paketo's two Github orgs.** At any given time each repository can have open issues in various stages of triage, and open pull requests requing the attention of language family maintainers and others in the community.
+
+## Rationale and Alternatives
+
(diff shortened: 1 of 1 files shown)


text/0026-triage-party.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Rationale and Alternatives
+
+Triage party provides unified views, filters, and metrics across repositories that can be customized to the needs of the Paketo project and community. Triage party is open source and already used by several large projects including [Minicube](http://tinyurl.com/mk-tparty) and [skaffold](http://tinyurl.com/skaffold-tparty)
+
+While there are a number of comercial services that could allow us to more easily triage issue across repositories their primary purpose is more around providing unified Kanban boards (Something Triage Party could also do) and licening costs may not allow these comercial services to be shared widly with the community. Because Triage Party provides read only views across n repositories there are also no user accounts or permissions to administer.
```
sophiewigmore, line 19: Should it be read-only access? Would the use case in the read-only scenario to allow users to see all open issues? Is there a way to get edit access if you have an account? I might be missing some context on how Triage Party actually works.

text/0026-triage-party.md
```diff
@@ -0,0 +1,61 @@
...
+
+Paketo has an an existing custom built dashboard at https://dashboard.paketo.io which gives you an overview of issue counts and a unified list of open issues. While this dashboard is great as a quick way to glance at the state of repos it suffers from a few shrotcommings:
+
+ 1. It requires each user to provide their personal github token which is a barier to use.
+
+ 1. There is no included rules engine that can be easily customized via a configuration file. Changes to filtering and and other enhancments require additional engineering time.
```
sophiewigmore, line 25: There is an issue open for this, but I'm not sure how easy it would be to make it configurable via a file.

text/0026-triage-party.md
```diff
@@ -0,0 +1,61 @@
...
+
+ 1. Given the breadth of the languages and buildpacks the Paketo project supports and the number of full time engineers known to be working on the project and current level of overal contributors, it's unlikley that the dashboard software will get much sustained development going forward.
+
+Finaly, Github search is somewhat useful and can be used to search for labels across repositories with some accuracy in a given Org. This approach, however, does not provide the at a gance overview of all repositories and metrics available by Triage Party.
```
sophiewigmore, line 29: I would say one drawback is that for me Triage Party is a bit confusing to look at. Maybe I need to get used to it/use it more to get acquainted with it. If we're directly comparing it to the Paketo Dashboard, it's less user friendly than the dashboard IMO.

dmikusa-pivotal:
I like the idea of having something to pull issues/prs/comments together across all the repos. I'm presently using github.com/issues with a filter string a mile long and I would definitely like something better.

Right now, what I'd like is:

a. A view across my language family (primarily Java, but also PHP & Rust)
b. The ability to sort/prioritize items, so anyone in the community can get a feel for what's in the queue for the language family.

There are probably other nice to have features, but it's kind of hard having not used the tool before to know what those might be. For now, my vote would be to have a demo that's been customized to at least show things by language family. Then we can play around with that and get some feedback.

text/0026-triage-party.md
andymoe: Everything is read only and open to the the public via triage party just like our existing Paketo git repos. Triage party provides different views of existing data across github repos.

andymoe:
Closing this for now until I have time to rework with a more useful demo.

Issue #56, opened 2021-05-10
Create RFC for Bill-of-Materials inclusion across Paketo projects


ForestEckhardt:
As part of the Paketo 2021 Roadmap Bill-of-Material(BOM) was called out as an important track of work to the Paketo project. This issue is to write up an RFC that will outline what this work will look like at a high level.

PR #57, opened 2021-05-10
RFC for Bill of Materials work

sophiewigmore:
Signed-off-by: Sophie Wigmore <swigmore@vmware.com>

## Summary

[Readable](https://github.com/sophiewigmore/rfcs/blob/sophie/bom/text/0028-bill-of-materials.md)

Resolves https://github.com/paketo-buildpacks/rfcs/issues/56

## Checklist


text/0028-bill-of-materials.md
```diff
@@ -0,0 +1,49 @@
...
+already make the BOM available. Another motivation for doing this work is to
+achieve a consistent user experience across the Paketo project.
+
+## Detailed Explanation
+
+The following Paketo components should have BOM metadata attached to them:
+
+### Stacks
+Stacks (such as those found in the [Paketo Stacks repository](https://github.com/paketo-buildpacks/stacks))
+should have BOM metadata that includes in-depth information on all of the OS
```
ekcasey, line 33: Can we document the label key and value format we would like to use here?

text/0028-bill-of-materials.md
```diff
@@ -0,0 +1,49 @@
...
+level packages installed as part of the stack.
+
+### Runtime and Compilation Dependencies
+Dependencies that provide runtimes and/or are used for compilation should have
+BOM metadata surfaced about them. This includes both the dependencies in the
```
ekcasey, line 38: Currently, build-only dependencies should be added to the build BOM in build.toml. It is probably worth calling out that distinction explicitly.


https://github.com/buildpacks/spec/blob/main/buildpack.md#bills-of-materials

text/0028-bill-of-materials.md
```diff
@@ -0,0 +1,49 @@
...
+a key value proposition of Cloud Native Buildpacks.
+
+As an implementation of Cloud Native Buildpacks, the Paketo project should
+support adding bill of materials metadata across the different components that
+go into an app image.
+
+## Motivation
+
+A BOM can include an array of different metadata about application dependencies
+and buildpack dependencies such as SHAs, URIs, versions, license information,
```
ekcasey, line 17: If the dependency is added to a buildpack-provided layer (rather than the app dir) java buildpacks currently include the name of the associated layer in the BOM entry. Maybe we should make this a standard practice across paketo buildpacks? Although the lifecycle might [help us out with this](https://github.com/buildpacks/rfcs/pull/151/files#diff-d31996aeb587aab806ae658bfb0199878b47d25545090827fa6b81723ca2d033R143) in the future.

ekcasey, line 17: I think we should try to standardize on the information we would ideally like to include in every entry (if possible) and the key names we will use. For example, It would be annoying for users if we used `url` in one buildpack and `uri` in another. I don't think this is likely to happen in practice, but using this RFC as a source of truth for the intended data format would make it more useful.

The list will not be comprehensive as there may be additional fields that are meaningful only in some ecosystems.

ForestEckhardt: What benefit does this provide to an end user inspecting the BOM? Is there a consumer of this field that I am unaware of?

text/0028-bill-of-materials.md
sophiewigmore: @ekcasey @matthewmcnew check out the section we added below and let us know your thoughts

text/0028-bill-of-materials.md
```diff
@@ -0,0 +1,95 @@
...
+## Detailed Explanation
+
+The following Paketo components should have BOM metadata attached to them:
+
+### Stacks
+Stacks (such as those found in the [Paketo Stacks repository](https://github.com/paketo-buildpacks/stacks))
+should have BOM metadata that includes in-depth information on all of the OS
+level packages installed as part of the stack.
+
+The existing metadata for stacks has the following structure:
```
matthewmcnew, line 36: Which label is this metadata on?

sophiewigmore: They can be found for each package on the config label named `io.paketo.stack.packages`.

text/0028-bill-of-materials.md
dmikusa-pivotal: I'm not 100% certain, but if you know the layer in with the item exists, you can tell what installed it. If you were trying to track down an issue with something that was installed, such as a dependency that wasn't the expected version that information would be helpful to know.

dmikusa-pivotal: +1 to what @ekcasey said

>The list will not be comprehensive as there may be additional fields that are meaningful only in some ecosystems.

I think the wording in the RFC should be such that it defines a *minimal* set of key/values that need to be included. Thus enabling individual buildpacks to include additional metadata that might be relevant to only its usecase.

text/0028-bill-of-materials.md
```diff
@@ -0,0 +1,95 @@
...
+
+  # Optional parameters
+  stacks = [<list of compatible stacks>]
+  cpe = "<version-specific common platfrom enumeration>"
+  licenses = [<licenses that the dependency has>]
+  source-uri = "<uri to the dependency source>"
+  source-sha256 = "<hash of the dependency source artifact from source-uri>"
+```
+
+### Language Specific Modules
```
dmikusa-pivotal, line 76: Calling this "Language Specific Modules" might be too restrictive. You could have non-language specific tools that install additional dependencies. For example, there are tools like `mvnw` and `gradlew` that are wrapper scripts to install Maven & Gradle, and Rust has `rustup` which is a tool for installing Rust.

I see two kinds of dependencies that are installed: 1.) Directly install dependencies and 2.) Indirectly installed dependencies. The former would be directly installed by the buildpack like runtimes & compilation tools (things that are listed in buildpack.toml). The latter would be things like NPM or Maven installed modules, but could be anything that a tool run by the buildpack installs like `rustup` and the Maven/Gradle wrappers.

I think it might help for clarity to have the title of this section and the `Runtime and Compilation Dependencies` section be more generic, and have the present titles be listed as specific examples under each section.

sophiewigmore: Great idea, thanks for the feedback. I tried to address this comment in https://github.com/paketo-buildpacks/rfcs/pull/57/commits/59e94453a4f0878a26b708553b532e3713324902

text/0028-bill-of-materials.md
sophiewigmore: Address this in https://github.com/paketo-buildpacks/rfcs/pull/57/commits/59e94453a4f0878a26b708553b532e3713324902. Let me know if this wording is strong enough

text/0028-bill-of-materials.md
```diff
@@ -0,0 +1,106 @@
...
+  "arch" : "<compatible architecture>",
+  "source" : {
+    "name" : "<package source name>",
+    "version" : "<package source version>",
+    "upstreamVersion" : "<package source upstream version>"
+  },
+  "summary" : "<package summary>"
+}
+```
+This closely resembles the metadata that is already available on stacks.
```
fg-j, line 51: Where can we find an example of the metadata already attached to Paketo stacks?

sophiewigmore: Stack metadata is listed under the `io.paketo.stack.packages` label.
Try `docker inspect <stack-image> | jq -r '.[0].Config.Labels."io.paketo.stack.packages"' | jq -r`

dmikusa-pivotal:
@ekcasey @matthewmcnew @fg-j - We reviewed this in the working group meeting. Just a quick reminder to submit feedback or resolve any ongoing conversations. Thanks!

text/0028-bill-of-materials.md
ryanmoran: Given that the upstream RFC to include BOM metadata in the layer was accepted, should we hold off on a requirement to specify the layer in the BOM entry itself?

text/0028-bill-of-materials.md
ekcasey: @matthewmcnew yes I agree that we should. However, I think it is okay for Paketo to essentially incubate this feature while CNB works on improving the BOM. With the understanding that we may need to make adjustments if/when CNB makes this a standard.

text/0028-bill-of-materials.md
```diff
@@ -0,0 +1,105 @@
...
+{
+  "name" : "<package name>",
+  "version" : "<package version>",
+  "arch" : "<compatible architecture>",
+  "source" : {
+    "name" : "<package source name>",
+    "version" : "<package source version>",
+    "upstreamVersion" : "<package source upstream version>"
+  },
+  "summary" : "<package summary>"
```
ekcasey, line 48: We should probably include a `licenses` key.

text/0028-bill-of-materials.md
```diff
@@ -0,0 +1,105 @@
...
+Stacks (such as those found in the [Paketo Stacks repository](https://github.com/paketo-buildpacks/stacks))
+should have BOM metadata that includes in-depth information on all of the OS
+level packages installed as part of the stack.
+
+The minimal set of metadata for stacks should be:
+```
+{
+  "name" : "<package name>",
+  "version" : "<package version>",
+  "arch" : "<compatible architecture>",
```
ekcasey, line 42: What is the utility of `arch`?

text/0028-bill-of-materials.md
```diff
@@ -0,0 +1,105 @@
...
+
+[bom.metadata]
+  sha256 = "<hash of dependency artifact from uri>"
+  uri = "<uri to dependency>"
+  version = "<dependency version>"
+```
+The optional set of keys may include:
+```
+  # Optional parameters
+  cpe = "<version-specific common platfrom enumeration>"
```
ekcasey, line 76: Seems like maybe we should include `purl` and `swid` as optional fields as well - https://cyclonedx.org/use-cases/#known-vulnerabilities

text/0028-bill-of-materials.md
```diff
@@ -0,0 +1,105 @@
...
+The following Paketo components should have BOM metadata attached to them:
+
+### Stacks
+Stacks (such as those found in the [Paketo Stacks repository](https://github.com/paketo-buildpacks/stacks))
+should have BOM metadata that includes in-depth information on all of the OS
+level packages installed as part of the stack.
+
+The minimal set of metadata for stacks should be:
+```
+{
```
ekcasey, line 39: In my mind it would make sense to use the same schema for the stack BOM as we use for the regular BOM. Eventually CNB should provide a way to combine these into a single BOM and using the same schema in both places will set us up nicely for that future. We may choose to require certain fields be set in certain cases in our buildpacks (e.g. we will set "source.upstreamVersion" for every package entry but not for language modules), but it would be nice if the schema itself was consistent.

text/0028-bill-of-materials.md
sophiewigmore: The processor architecture that the package is compatible, such as `amd64`.
While it's perhaps not the most valuable BoM field, it's information that we already currently surface for the stacks in the `io.paketo.stack.packages` label, so we thought we might as well surface more information than less.

text/0028-bill-of-materials.md
sophiewigmore: @ForestEckhardt thoughts here? My thought was that between the `cpe` notation, and the source URL/SHA256, that would be plenty of information about vulnerabilities/where the package came from. 

Adding `swid`/`purl` would be quite involved to generate/get this for every dependency in the dep-server which to me might not be worth the tradeoff of the value add. It's also worth noting that tools like `syft` that generate CycloneDX can easily get this information

text/0028-bill-of-materials.md
sophiewigmore: The reason we left this out is because we were investigating tools to collect licenses for the stacks, and the best option we found[dpkg-licenses](https://github.com/daald/dpkg-licenses/)  which only guarantees up to 90% of all license information, due to the lack of standardization in licenses. The tradeoff between the work involved getting that last 10% of license metadata, and user need wasn't worth it at the time we investigated this, so we left it out as something in the minimal set of fields.
Issue where we mentioned this: https://github.com/paketo-buildpacks/stacks/issues/70

I do think it would be good to include license information in the future, but I would be okay leaving it out of this RFC for now.

text/0028-bill-of-materials.md
sophiewigmore: I agree with this.

text/0028-bill-of-materials.md
sophiewigmore: Ah looking at the most recent comments on the upstream RFC, I see that CPEs aren't as encouraged as SWID/pURLs since it's going to be deprecated soon...I am rethinking my original comment now and can see that there is a value add with SWID/pURL.

text/0028-bill-of-materials.md
sophiewigmore: addressed in https://github.com/paketo-buildpacks/rfcs/pull/57/commits/077a268c0465f665bff252635d084806264349c9

text/0028-bill-of-materials.md
```diff
@@ -0,0 +1,114 @@
...
+[bom.metadata.source]
+  name = "<package source name>"
+  version = "<package source version>",
+  upstream-version = "<package source upstream version>"
+```
+The optional set of keys may include:
+```
+  # Optional parameters
+  arch = "<compatible architecture>"
+  summary = "<package summary>"
```
ryanmoran, line 54:
```suggestion
# Optional parameters
[bom.metadata]
  arch = "<compatible architecture>"
  summary = "<package summary>"
```

ryanmoran (changes requested):
I can see that each of the types of BOM entries seem to be conforming to a standardized schema. Could we show that complete schema at some point in the RFC?

text/0028-bill-of-materials.md
ForestEckhardt: After some investigation it is unclear to me whether or not we are on the hook for generating the `SWIDs`. My current research leads me to believe that the dependency maintainer is responsible for creating a `SWID` so our ability to add this information is out of our control.

sophiewigmore: We have decided to move forward with support package URLs following [some investigation](https://github.com/paketo-buildpacks/dep-server/issues/42#issuecomment-880961646).

sophiewigmore: Check out https://github.com/paketo-buildpacks/rfcs/pull/57/commits/5fc173d286d60e685ed81876e1f4a1765cd39b46

text/0028-bill-of-materials.md
```diff
@@ -0,0 +1,166 @@
...
+license results to ensure that individuals can still trust that all relevant
+licenses are present. This level of confidence and verification may change in
+the future, but that change should be communicated in later documentation.
+
+### Package URLs
+
+[Package URLs](https://github.com/package-url/purl-spec) will be provided for
+our directly installed dependencies. These types of dependencies (runtime
+dependencies) do not have explicitly supported types in the [Package URL
+Types](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst)
```
ekcasey, line 162:
```suggestion
our directly installed dependencies. Some of these dependencies (runtime
dependencies) do not have explicitly supported types in the [Package URL
Types](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst)
```

There are some cases where dependencies have defined PURL types. For example, the [spring-cloud-bindings dependency](https://github.com/paketo-buildpacks/spring-boot/blob/main/buildpack.toml#L48) installed by paketo-buildpacks/spring-boot, would be type `maven`.

text/0028-bill-of-materials.md
```diff
@@ -0,0 +1,166 @@
...
+document, so we will use the `generic` or `debian` types as instructed by
```
ekcasey, line 163:
```suggestion
document, so we will either use the `generic` type or another type as instructed by
```
Given these aren't deb packages, the `deb` type is likely inappropriate

text/0028-bill-of-materials.md
```diff
@@ -0,0 +1,166 @@
...
+The minimal set of keys to include in these type of BOM entries are:
+```
+[[bom]]
+name = "<dependency name>"
+
+[bom.metadata]
+  version = "<dependency version>"
+
+[bom.metadata.source]
+  name = "<dependency source name>"
```
ekcasey, line 72:
```suggestion
  name = "<dependency source name>"
  purl = "<package URL>"
```

Given all of the stack dependencies are ubuntu LTS packages it should be easy to provide valid purls for all of these. e.g.
```
pkg:deb/ubuntu/curl@7.50.3-1?arch=amd64&distro=bionic
```

text/0028-bill-of-materials.md
```diff
@@ -0,0 +1,166 @@
...
+  version = "<dependency source version>",
+  upstream-version = "<dependency source upstream version>"
+```
+The optional set of keys may include:
```
ekcasey, line 76: Is there any reason to call out optional keys? I was assuming anything in the schema that wasn't required was optional.

sophiewigmore: Good call. Addressed in https://github.com/paketo-buildpacks/rfcs/pull/57/commits/7575872a10d60d0017ce9d052ca34a534bcde553

ekcasey: Sorry for nitpicking on this but I am still confused after the above linked changes. In my mind there are only two things we need to specify. The complete schema (any field in which could optionally be filled out, for any type of entry) and any required fields for specific entry types. I don't understand why we re repeating subsets of the schema if to indicate required fields.

Are these the suggested fields (in which case maybe we could teak the wording to make that clear)? Or does this imply that other fields in the schema are disallowed for these dependency types (I don't think we should do this)?

text/0028-bill-of-materials.md
```diff
@@ -0,0 +1,162 @@
...
+The bill of materials for each of type of entry should have some subset of
+fields that conform to the overall schema below.
+```
+[[bom]]
+name = "<dependency name>"
+
+[bom.metadata]
+  arch = "<compatible architecture>"
+  cpe = "<version-specific common platfrom enumeration>"
+  deprecation_date = "<dependency EOS date formatted in using RFC 3339>"
```
ryanmoran, line 42:
```suggestion
  deprecation-date = "<dependency EOS date formatted in using RFC 3339>"
```

Looks like we are already using kebob-casing elsewhere. We should stick with that for consistency.

text/0028-bill-of-materials.md
```diff
@@ -0,0 +1,162 @@
...
+[`buildpack.toml` file](https://github.com/paketo-buildpacks/node-engine/blob/main/buildpack.toml).
+
+The set of keys to include in these type of BOM entries are:
+```
+[[bom]]
+name = "<dependency name>"
+
+[bom.metadata]
+  cpe = "<version-specific common platfrom enumeration>"
+  deprecation_date = "<dependency EOS date formatted in using RFC 3339>"
```
ryanmoran, line 101:
```suggestion
  deprecation-date = "<dependency EOS date formatted in using RFC 3339>"
```

ryanmoran (changes requested):

text/0028-bill-of-materials.md
sophiewigmore: Following this comment/our discussion around this at the 7-27 Working Group meeting, we have gone ahead and streamlined the RFC to solely include the top-level schema to show possible fields, and removed the dependency-type schemas. They are "suggested fields", the only required fields are "name" and "version".

text/0028-bill-of-materials.md
```diff
@@ -0,0 +1,113 @@
...
+fields that conform to the overall schema below.
+```
+[[bom]]
+name = "<dependency name>"
+
+[bom.metadata]
+  arch = "<compatible architecture>"
+  cpe = "<version-specific common platfrom enumeration>"
+  deprecation-date = "<dependency EOS date formatted in using RFC 3339>"
+  licenses = [<licenses that the dependency has>]
```
ryanmoran, line 43: Can we specify what format these license identifiers should appear in? Are they SPDX format?

sophiewigmore: Done, and yes they are

ryanmoran (approved):

PR #71, opened 2021-06-02
RFC: Reloadable process types RFC

paulcwarren:
Reloadable process types RFC

## Summary
Provide a utility buildpack to support language buildpacks in creating reloadable process types for their run containers.
## Use Cases

Usage of dev orchestration tools like Tilt and Skaffold are on the rise. These tools aim to make  developing distributed software that runs on Kubernetes easier.  Offering the capability to develop inside the container via live reload features.  With this, changes are streamed from the developer’s workstation into the running container whereupon the init process is restarted.  The dev orchestrator, of course, handles most of this by building the changes and copying them in the container but ultimately the running container needs the ability to restart its entrypoint process for the changes to take effect.

Today buildpacks are unable to participate in these live reload development workflows but, moving forward, we would very much like them to be able to.

## Checklist


text/0000-reloadable-process-types.md
```diff
@@ -0,0 +1,36 @@
...
+
+When enabled by a BP_LIVE_RELOAD_ENABLED=true environment variable, the language buildpacks may request the inclusion of watchexec, via the build plan.
+
+This affords language buildpacks the ability to use watchexec for creating a reloadable process type when no better options are available.  For example, the java buildpack may prefer to use spring-boot-devtools if that is available but can always fallback to watchexec otherwise.  
+
+As a result, a language buildpack may produce a reloadable process type such as:
+
+`watchexec -r java org.springframework.boot.loader.JarLauncher`
+
+and in doing so, for the reloadable process type, the buildpack makes watchexec the stable init process for the container that monitors the workspace directory for file changes, restarting the app process when they occur.
```
arjun024, line 25: > that monitors the workspace directory for file changes, restarting the app process when they occur.

Many buildpacks may also require that you rebuild the source code again before it is run. E.g. Changes to C# source code when used with dotnet-core buildpack. Is this buildpack intended just for Java buildpacks?

dmikusa: Correct me if I'm wrong Paul, but I believe this is the tool's responsibility (Tilt/Skaffold).

>The dev orchestrator, of course, handles most of this by building the changes and copying them in the container

text/0000-reloadable-process-types.md
```diff
@@ -0,0 +1,36 @@
...
+
+Usage of dev orchestration tools like Tilt and Skaffold are on the rise. These tools aim to make  developing distributed software that runs on Kubernetes easier.  Offering the capability to develop inside the container via live reload features.  With this, changes are streamed from the developer’s workstation into the running container whereupon the init process is restarted.  The dev orchestrator, of course, handles most of this by building the changes and copying them in the container but ultimately the running container needs the ability to restart its entrypoint process for the changes to take effect.
+
+Today buildpacks are unable to participate in these live reload development workflows but, moving forward, we would very much like them to be able to.
+
+## Detailed Explanation
+
+We propose creating a utility buildpack that packages the watchexec command line utility for restarting arbitrary processes.
+
+When enabled by a BP_LIVE_RELOAD_ENABLED=true environment variable, the language buildpacks may request the inclusion of watchexec, via the build plan.
```
dmikusa, line 17: Here's what I'm taking away from this, please let me know if I'm missing anything:

This buildpack will participate (i.e. detection passes) if all of the following conditions are met:

- BP_LIVE_RELOAD_ENABLED=true is set
- A buildpack requests `watchexec` (or some other key) in the build plan.

The buildpack will do the following:

- Contributes `watchexec` to a layer with command on $PATH

What I'm unclear about is:

1. Will this buildpack be contributing any process types? or is that going to be the responsibility of another buildpack to do? For example, in Java the executable-jar buildpack contributes process types for running the executable JAR. Would the executable-jar buildpack have to do something different if `watchexec` is present? 

2. If this buildpack is responsible for providing `watchexec` in the build plan, what buildpack is responsible for requiring `watchexec` in the buildplan? Will this buildpack both provide & require it, which is fine? Or is some other buildpack supposed to require it?

If other buildpacks need to be changed to opt-in and participate with this buildpack, I think you need to add some more detail here. Specifically, what buildpacks will need to be modified, what changes will need to be made to those and what do the order groups need to be in the composite buildpacks so this all works correctly.

text/0000-reloadable-process-types.md
```diff
@@ -0,0 +1,36 @@
...
+
+`watchexec -r java org.springframework.boot.loader.JarLauncher`
+
+and in doing so, for the reloadable process type, the buildpack makes watchexec the stable init process for the container that monitors the workspace directory for file changes, restarting the app process when they occur.
+
+Note that watchexec watches the current directory (and below). The working directory is the app directory so all buildpacks wishing to support reloadable process types would need to contribute their entrypoint processes to the app directory.
+
+Note also that watchexec supports shell-less invocation and therefore will be able to support the tiny builder that produces a run image with no shell.  
+
+It also supports windows through both Cmd and Powershell shells.
```
dmikusa, line 31: Is this more of an FYI, like `watchexec` supports Windows so could potentially be used with Windows buildpacks? Or is there something specific that this proposal would be doing to enable Windows support? In either case, it might be good to clarify a little more the intent of this statement.

text/0000-reloadable-process-types.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Motivation
+
+Usage of dev orchestration tools like Tilt and Skaffold are on the rise. These tools aim to make  developing distributed software that runs on Kubernetes easier.  Offering the capability to develop inside the container via live reload features.  With this, changes are streamed from the developer’s workstation into the running container whereupon the init process is restarted.  The dev orchestrator, of course, handles most of this by building the changes and copying them in the container but ultimately the running container needs the ability to restart its entrypoint process for the changes to take effect.
+
+Today buildpacks are unable to participate in these live reload development workflows but, moving forward, we would very much like them to be able to.
+
+## Detailed Explanation
+
+We propose creating a utility buildpack that packages the watchexec command line utility for restarting arbitrary processes.
```
fg-j, line 15: So, the proposal is simply to create a `watchexec` buildpack, similar to other dependency-providing buildpacks like [Paketo Yarn](https://github.com/paketo-buildpacks/yarn) or [Paketo Bundler](https://github.com/paketo-buildpacks/bundler)?

text/0000-reloadable-process-types.md
fg-j: When I was exploring the potential to use Tilt + buildpacks, one thing I noticed was that a user generally determines the build command(s) that are used to produce the contents of their image. (Often via a Dockerfile). When doing a hot reload, Tilt rebuilds the app image contents using (a subset of) those same build command(s) each time. Then, the resulting artifacts are copied into the running app container.

Buildpack builds have a key difference: buildpacks - not users - determine the build commands that produce the image. And these commands aren't checked into a file – they're implicit in the buildpacks' source code. If rebuilds are delegated to the dev tool (e.g. Tilt), if it doesn't follow the same process as the buildpack's build, you could easily end up with a built artifact that is incompatible with the app container (e.g. built for wrong operating system), or just substantively different from the contents produced by the buildpack (e.g. built with a different set of dependencies).

This could lead to failing (re)builds, containers whose behaviour in development diverges from their behaviour in prod, and other issues.

Are there any ideas about how to mitigate this problem / create an "it just works" experience here?

text/0000-reloadable-process-types.md
dmikusa-pivotal: I'm not sure it's that simple. The existing utility buildpacks are basically self-sufficient. They get invoke, do something and the other buildpacks don't care. Think Procfile or env-variables. They are setting process types or env variables on the image, other buildpacks don't need to be aware this is happening.

I think this proposal is a little more involved because the language buildpack would have to opt-in and modify its process types if watchexec is enabled, since there's no way for `watchexec` to emit the necessary process types (at least I don't think it's possible).

fg-j: 🤔  I see. Seems like language buildpacks would be able to "opt in" to adding the reload feature once the watchexec buildpack exists.

I'm wondering if, for the sake of simplicity, this RFC should just propose and outline the functionality of the watchexec buildpack. Or is part of the effort here to align the Paketo buildpacks to _all_ have reloadable processes support that works roughly the same? That's probably a question for @paulcwarren.

natalieparellano: There has been some discussion in the past about whether it should be possible for buildpacks to see what process types have been defined by other buildpacks (technically, it's possible to read other buildpacks' launch.toml but not defined by the spec). I wonder if having such a mechanism would make this proposal easier.

text/0000-reloadable-process-types.md
ekcasey: > This buildpack will participate (i.e. detection passes) if all of the following conditions are met:
>
> BP_LIVE_RELOAD_ENABLED=true is set
> A buildpack requests watchexec (or some other key) in the build plan.

@dmikusa The way I am thinking about this the watchexec buildpack does not require `BP_LIVE_RELOAD_ENABLED=true` to be true to pass detect.  The buildpack will always provide watchexec and thus participate if a downstream buildpack requests watchexec via the build plan. The downstream buildpack on the other hand, makes a decision about whether to request and use watchexec based on the value of `BP_LIVE_RELOAD_ENABLED`.

ekcasey (commented):
@pau This is almost certainly the result of misleading instructions in the repo README, but we shouldn't delete the RFC template in this PR.

text/0000-reloadable-process-types.md
paulcwarren: The intention here, I think, is that pack (and this utility buildpack) would be used with `tilt` and `skaffold` custom builds, something like this (tilt example):

```
custom_build('localhost:5000/apps/demo', 
  'pack build -e BP_LIVE_RELOAD_ENABLED=true --tag $EXPECTED_REF --publish,
  ['pom.xml'],
  live_update = [
    sync('./target/classes', '/workspace/BOOT-INF/classes')
  ]
)
k8s_yaml('./k8s/deployment.yaml'))
k8s_resource('spring-petclinic', port_forwards="8080:8080”)
````

When `tilt up` is run this will perform an initial build and deployment, using pack, that publishes an image using the tag TIlt's wants.  It then live sync's a limited set of changes, in this case changes within the existing Java classpath compiled by the IDE, directly into the application container.  Other changes outside of this, to the pom for example, would trigger a full build and deployment again that would again delegate to pack.

Does that clarify things?

text/0000-reloadable-process-types.md
paulcwarren: The former I think.  If windows-related language buildpacks could, indeed, take advantage of a reloadable process type in the same/similar way to Java then I was noting here that watchexec supports windows operating system.

Please let me know if you would like this clarified and I will commit a change to do so.

text/0000-reloadable-process-types.md
paulcwarren: What @dmikusa-pivotal says basically.  And also see Emily's [comment above](https://github.com/paketo-buildpacks/rfcs/pull/71#discussion_r647666558)

paulcwarren:
> @paulcwarren This is almost certainly the result of misleading instructions in the repo README, but we shouldn't delete the RFC template in this PR.

OK.  I'll add it back in.

text/0000-reloadable-process-types.md
dmikusa-pivotal: Yes, if you could. Maybe just say that the potential is there, but it's out of the scope of this proposal.

text/0000-reloadable-process-types.md
dmikusa-pivotal: @paulcwarren can you add a bit to clarify this? I see three points that need to be described.

1. Clarify that the watchexec buildpack is just providing the binary, nothing more. That makes it very simple.
2. Clarify what another buildpack would need to do to opt-in. Something to the effect of watch for a variable like Emily mentions in her comment and then modify the start command to use watchexec.
3. I think you'd like to start with Java support for this, so indicate that as well in the proposal.

text/0000-reloadable-process-types.md
paulcwarren: Done.  See latest commit.

ryanmoran: The intent here is to make some progress towards the idea of "hot-reloadable" applications using buildpacks, while not requiring that we undertake the entire workstream that would include significant changes upstream to the buildpack spec/lifecycle. The tradeoffs to correctness or ease-of-use are understood, but could maybe be explicitly outlined here so that others reading this can understand this RFCs shortcomings. @paulcwarren, do you think it would be appropriate to include a bit about some of these tradeoffs in the design of this RFC?

ekcasey (approved):

text/0000-reloadable-process-types.md
paulcwarren: Done

text/0000-reloadable-process-types.md
paulcwarren: The tradeoffs being?

ryanmoran: I think we only have that one concern. It feels worth noting that once the container is initially built, the only thing it can do is swap some files on the filesystem and restart the process. So, if your application needs to be recompiled, this must happen outside of the built container and in so doing must happen in a way that remains compatible with the built container environment.

paulcwarren: Noted.  Thanks.  I'll update the RFC tomorrow AM.  Thankyou.

paulcwarren: Done.  See [3f6b870](https://github.com/paketo-buildpacks/rfcs/pull/71/commits/3f6b870691c500a8d1e6201a2f990ea9761485e5)

text/0000-reloadable-process-types.md
paulcwarren: Done.  See [3f6b870](https://github.com/paketo-buildpacks/rfcs/pull/71/commits/3f6b870691c500a8d1e6201a2f990ea9761485e5)

dmikusa-pivotal (approved):

text/0000-reloadable-process-types.md
fg-j: Thanks for fleshing that out. This does clarify things.

fg-j (approved):

ryanmoran (approved):

ekcasey:
Merged in https://github.com/paketo-buildpacks/rfcs/pull/90

PR #96, opened 2021-07-19
RFC: Add pURLs to the dependency server

ForestEckhardt:
[Readable](https://github.com/ForestEckhardt/rfcs-1/blob/add-purl-to-dep-server/text/dependencies/rfcs/0002-dependency-server-purl.md)

Resolves https://github.com/paketo-buildpacks/dep-server/issues/83

### DIFF
--- a/text/dependencies/rfcs/0002-dependency-server-purl.md
+++ b/text/dependencies/rfcs/0002-dependency-server-purl.md
@@ -18,7 +18,11 @@ identified by the end user. Package URLs are also a standard supported by the Cy
 The following field should be added to the dependency server output:
 
 * `purl`: a string which is a pURL for that dependency that conforms to the
-  [pURL Specification](https://github.com/package-url/purl-spec).
+  [pURL Specification](https://github.com/package-url/purl-spec). For
+  consistency, we will be using the `generic` type because it allows to
+  consistently generate the pURLs. Because we are using a `generic` type we
+  will also be including the `source` and `source_sha256` as qualifiers on the
+  pURL to give end users as much specific information as necessary.
 
 An example of the new output:
 ```
--- a/text/dependencies/rfcs/0002-dependency-server-purl.md
+++ b/text/dependencies/rfcs/0002-dependency-server-purl.md
@@ -1,0 +1,48 @@
+# Add pURL Field to the Dependency Server Metadata
+
+## Summary
+
+Add a field for pURLs into the dependency server schema.
+This RFC is an addendum to [RFC #0009](https://github.com/paketo-buildpacks/rfcs/blob/main/text/0009-dep-server.md)
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


text/dependencies/rfcs/0002-dependency-server-purl.md
```diff
@@ -0,0 +1,48 @@
...
+      {
+        "id": "io.paketo.stacks.tiny",
+        "mixins": ["some-required-mixin"]
+      }
+    ],
+    "source": "https://dl.google.com/go/go1.15.src.tar.gz",
+    "source_sha256": "69438f7ed4f532154ffaf878f3dfd83747e7a00b70b3556eddabf7aaee28ac3a",
+    "deprecation_date": "",
+    "licenses": ["BSD-3-Clause"],
+    "purl": "pkg:generic/go@1.15?download_url=https://dl.google.com/go/go1.15.src.tar.gz&checksum=sha256:69438f7ed4f532154ffaf878f3dfd83747e7a00b70b3556eddabf7aaee28ac3a"
```
sophiewigmore, line 45: This example is good, but it might be helpful to go into more detail in the implementation section that we are specifically going to be using the `generic` type, rather than something else (like the `debian` type) and what the other bells and whistles here are (the checksum, the download URL)

This is helpful so we have a standard we can apply to all of our dependencies consistently.

text/dependencies/rfcs/0002-dependency-server-purl.md
```diff
@@ -0,0 +1,48 @@
...
+Add a field for pURLs into the dependency server schema.
+This RFC is an addendum to [RFC #0009](https://github.com/paketo-buildpacks/rfcs/blob/main/text/0009-dep-server.md)
+
+## Motivation
+
+Currently, the dependency server offers a lot of valuable information about the
+files it hosts. However, it currently does not have the capacity to supply
+pURLs. This field should be added to the dependency server because pURLs are an
+industry standard package identifier that allow dependencies to be accurately
+identified by the end user.
```
sophiewigmore, line 14:
```suggestion
pURLs. This field should be added to the dependency server because pURLs are an
industry standard package identifier that allow dependencies to be accurately
identified by the end user. Package URLs are also a standard supported by the CycloneDX Bill of Materials format, and are [integrated with a number of other tools](https://github.com/package-url/purl-spec#users-adopters-and-links). Overall, package URLs can provide rich information beyond what's available via CPEs alone, making them a strong choice to include on our dependency metadata.
```

sophiewigmore (changes requested):

text/dependencies/rfcs/0002-dependency-server-purl.md
ForestEckhardt: I have just pushed up a change. Give it a read and see if it satisfies this comment.

sophiewigmore (approved):

PR #98, opened 2021-07-28
RFC: Node.js Installation Buildpacks BOM

sophiewigmore:
## Summary

Resolves https://github.com/paketo-buildpacks/nodejs/issues/425

[Readable](https://github.com/paketo-buildpacks/rfcs/blob/install-bom/text/nodejs/0013-install-bom.md)

## Checklist

### DIFF
--- a/text/nodejs/0013-install-bom.md
+++ b/text/nodejs/0013-install-bom.md
@@ -81,11 +81,6 @@ Node.JS language family buildpack order groups.
   * directly before the `npm-start` buildpack in the NPM order group, and
   * directly before the `node-start` buildpack in the no-package-manager order group.
 
-Users of the Node.JS buildpack can opt out of this behaviour by creating a
-custom order grouping that does not include the BOM Generation buildpack, or
-they can set a `BP_ENABLE_MODULE_BOM` flag during container build-time to `false`.
-The default for this flag will be `true`.
-
 #### Detection
 
 The Node Module BOM Generator CNB always detects with the following contract:
@@ -122,6 +117,13 @@ The build phase will perform a few tasks as mentioned above.
   command.
 
 6. Clean up the generated BOM file
+
+#### Potential Future Optimizations
+
+Given future user interest in a way to opt out of this buildpack beyond a
+custom order grouping, an optimization that could be made is the inclusion of a
+`BP_ENABLE_MODULE_BOM` environment variable that can be set during container
+build-time.
 
 ## Rationale
 
--- a/text/nodejs/0013-install-bom.md
+++ b/text/nodejs/0013-install-bom.md
@@ -1,0 +1,48 @@
+#  Installation Buildpack BOM Generation
+
+## Summary
+
+Add support for the generation of the Bill of Materials in the NPM Install and
+Yarn Install buildpacks in order to provide entries for modules installed
+during the build process.
+
+## Motivation
+
+As a part of an initiative to provide a Bill of Materials (BOM) in the Node.JS
+Language Family, we have to support generation in both buildpacks that directly
+install dependencies (Node Engine and Yarn), as well as in buildpacks that run
+an installion command that results in a set of installed dependencies (NPM
+Install and Yarn Install)
+
+For the directly installed dependencies in the Node Engine and Yarn buildpacks,
+we control the metadata around the dependencies and largely implemented this on
+the dependency server side. The availability of this metadata made it
+straightforward to consume the metadata and generate the BOM.
+
+For the indirectly installed modules that are installed by the NPM Install
+and Yarn Install buildpacks, this is not the case because we delegate the
+installation process to the package managers. Due to the nature of these
+dependencies and how they are installed, the process of generating a Bill of
+Materials will require changes to the installation buildpacks themselves.
+
+This is an important piece of the Bill of Materials, because it gives a
+complete picture of what dependencies are inside of the final application
+image. Without this metadata, a major value-add of a Bill of Materials is lost
+and users may have to use third-party tools to generate the rest of the BOM,
+which is a less than ideal user experience.
+
+## Proposal
+
+TODO: Use https://github.com/CycloneDX/cyclonedx-node-module
+
+## Implementation
+
+TODO: describe how this will work
+
+## Alternatives
+
+TODO: do it by hand
+
+## Prior Art
+
+TODO: RFC
--- a/text/nodejs/0013-install-bom.md
+++ b/text/nodejs/0013-install-bom.md
@@ -2,9 +2,9 @@
 
 ## Summary
 
-Add support for the generation of the Bill of Materials in the NPM Install and
-Yarn Install buildpacks in order to provide entries for modules installed
-during the build process.
+Add support for the generation of the Bill of Materials in the Node Language
+Family in order to provide entries for modules installed
+during the build process that can be found on the final app image.
 
 ## Motivation
 
@@ -21,9 +21,10 @@ metadata made it straightforward to consume the metadata and generate the BOM.
 
 For the indirectly installed modules that are installed by the NPM Install and
 Yarn Install buildpacks, this is not the case because we delegate the
-installation process to the package managers. Due to the nature of these
-dependencies and how they are installed, the process of generating a Bill of
-Materials will require changes to the installation buildpacks themselves.
+installation process to the package managers. The case in which we have no
+package manager and use vendored node modules also falls under this case. Due
+to the nature of these dependencies and how they are installed, the process of
+generating a Bill of Materials will require changes to the language family itself.
 
 This is an important piece of the Bill of Materials, because it gives a
 complete picture of what dependencies are inside of the final application
@@ -36,9 +37,8 @@ which is a less than ideal user experience.
 
 ### Tool details
 
-The Bill of Materials that will result from the NPM/Yarn Install buildpacks
-should be an accurate accounting of all of the modules that get installed into
-the `node_modules` directory.
+The Bill of Materials that will result should be an accurate accounting of all
+of the modules that get installed into the `node_modules` directory.
 
 There already exists a tool called [CycloneDX Node.js
 Module](https://github.com/CycloneDX/cyclonedx-node-module) which is an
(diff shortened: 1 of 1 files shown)


text/nodejs/0013-install-bom.md
```diff
@@ -0,0 +1,116 @@
...
+All of these are fields that we aim to support per [the Paketo BOM
+RFC](https://github.com/sophiewigmore/rfcs/blob/8b1e8c9ed6201313f47c5897223cbffb265e96ed/text/0028-bill-of-materials.md)
+
+### Buildpack Additions
+
+Using the CycloneDX tool to generate the BOM will involve some changes to the
+NPM Install and Yarn Install Buildpacks.
+
+* After the `npm install` or `yarn install` command is run during the build
+  phase, we should run `npm install -g @cyclonedx/bom` to install the tool on
```
arjun024, line 70: NPM Install/Yarn Install buildpacks are designed such that they can be used offline in Internet-less environments. This installation of cyclonedx/bom doesn't seem to work in offline mode. Can this be packaged as a dependency? Or even, do you think this merits a buildpack of its own or be in the build image? Regardless, I think we should maintain install buildpack's ability to work fully in offline mode.

sophiewigmore: @arjun024 that's a good point. I'm more inclined to have this run as a part of this buildpack since the goal is just to collect insights around the install process, and it doesn't really have a role in adding something to the final app image (besides metadata). 

I can look into packaging this as a dependency and update you here.

sophiewigmore: I think @ForestEckhardt and I have two options we are considering after looking into this:

A. Extract the tool as a tarball with its dependencies, and make it available as a standalone dep-server dependency. Install it with it’s own buildpack in both an online and offline scenario. Run BOM creation using the tool in its own buildpack after NPM Install/Yarn Install/ or before Node Start in the no package manager scenario. 

B. Extract the tool as a tarball with its dependencies, and then run BOM tool installation and BOM creation together in one buildpack after NPM Install/Yarn Install/ before Node Start in the no package manager scenario.

We are leaning towards having this process be a part of its own buildpack(s) for a couple reasons. First, it keeps buildpack modularity intact. While it seems intuitive to run BOM creation alongside the NPM Install and Yarn Install buildpacks, the issue is that we need to support the Node Start case, and it seems unintuitive to create the BOM in this buildpack, since it's a start-command buildpack.
 @arjun024 do you have any thoughts?
The other reason we are leaning to one of these two set ups is because it feels slightly sneaky to slip the dependency into the image within another buildpack. We could add it as a part of the node-engine dependency or install it in the node-engine buildpack, but this doesn't seem that transparent especially since the inclusion of this dependency opens the door for additional CVEs.

sophiewigmore: We are ideally going with option B here. We will host the CycloneDX BOM tool as a dependency in the dep-server, and then introduce a buildpack that runs before start commands to install the tool and run BOM generation together.

arjun024: SGTM

text/nodejs/0013-install-bom.md
```diff
@@ -61,36 +61,70 @@ the following fields for each module:
...
+### Language Family Additions
 
 Using the CycloneDX tool to generate the BOM will involve some changes to the
-NPM Install and Yarn Install Buildpacks.
+Node.JS language family buildpack order groups.
+
+* A single BOM generation buildpack will be added. It's primary functions are:
+  * perform the BOM dependency tool retrieval from the dep-server,
+  * execute BOM generation with the tool (`cyclonedx-bom -o bom.json`), and
+  * contribute the BOM to the layer metadata.
```
arjun024, line 76: Nitpick: You are not writing to the "layer metadata" right? You're gonna write to the {build,launch}.toml? (build/launch environment metadata - I don't know the technical term?) I think of layer metadata as `<layers>/<layer>.toml`. I think spelling it out would be clearer.

text/nodejs/0013-install-bom.md
```diff
@@ -61,36 +61,70 @@ the following fields for each module:
...
+Users of the Node.JS buildpack can opt out of this behaviour by creating a
+custom order grouping that does not include the BOM Generation buildpack, or
+they can set a `ENABLE_MODULE_BOM` flag during container build-time to `false`.
+The default for this flag will be `true`.
+
+#### Detection
+
+The buildpack will pass detection in two different cases:
+
+* Detection will pass if there is a `node_modules` directory in the source directory.
```
arjun024, line 93: I don't think apps would have a `node_modules` dir at detect time unless they are vendored. The x-install buildpacks generate node_modules dir during the build phase

text/nodejs/0013-install-bom.md
```diff
@@ -61,36 +61,70 @@ the following fields for each module:
...
 
-* Implement some code to unmarshal the JSON Bill of Materials and transform
-  each BOM entry into a
-  [`postal.Dependency`](https://github.com/paketo-buildpacks/packit/blob/c5a40518f2c6bd913ade999b9e2d58d6892d2ea9/postal/buildpack.go#L12)
-  type.
+The build phase will perform a few tasks as mentioned above.
 
-* (Optionally) generate `CPE` and `pURL` from the data and add it to the
+1. Perform the BOM dependency tool retrieval from the dep-server. It will get
+   added to a build-time layer. It will contribute a BOM entry for the tool itself
```
arjun024, line 106: >  It will get added to a build-time layer

Why is this tool added to a layer (& a BOM entry for the tool itself added) instead of just using the tool to generate the BOM and throwing away the tool?

text/nodejs/0013-install-bom.md
sophiewigmore: Yea the point here was that. If the modules are vendored, we don't have to require `node_modules`. My wording is a bit misleading I can fix it up

text/nodejs/0013-install-bom.md
sophiewigmore: We wanted to cache the tool so that if any node modules change, we can still generate the BOM again without having to re-download the tool.

A BOM entry for the tool is necessary because it's something we are installing onto the image, and we need to record of everything we install via our buildpacks, same as we do for node engine and. yarn

text/nodejs/0013-install-bom.md
sophiewigmore: tried to fix the wording in https://github.com/paketo-buildpacks/rfcs/pull/98/commits/7f6e30ec1eedcbb1221f9472bb97fa3464887a54

text/nodejs/0013-install-bom.md
sophiewigmore: See https://github.com/paketo-buildpacks/rfcs/pull/98/commits/7f6e30ec1eedcbb1221f9472bb97fa3464887a54

text/nodejs/0013-install-bom.md
arjun024: Can we fail detection if BP_*_BOM is set to false? - so we can use it as an optional buildpack. If that's too detailed for the RFC, that's fine too.

text/nodejs/0013-install-bom.md
```diff
@@ -0,0 +1,159 @@
...
+
+* For each of the three Node.JS buildpack order groups, the BOM generation
+  buildpack will run directly before the start command buildpack. This means:
+  * directly before the `yarn-start` buildpack in the Yarn order group,
+  * directly before the `npm-start` buildpack in the NPM order group, and
+  * directly before the `node-start` buildpack in the no-package-manager order group.
+
+Users of the Node.JS buildpack can opt out of this behaviour by creating a
+custom order grouping that does not include the BOM Generation buildpack, or
+they can set a `BP_ENABLE_MODULE_BOM` flag during container build-time to `false`.
```
arjun024, line 86: Are you planning to use this as a generic flag across language families? If it's nodejs-specific, I suggest we stick to the convention of the BP_NODE prefix

text/nodejs/0013-install-bom.md
sophiewigmore: Yea!! That's what we're proposing. Is that unclear in our explanation?

text/nodejs/0013-install-bom.md
ryanmoran: It may make sense to have both. Imagine that a user might want to disable the BOM for just a single buildpack, or all BOM generation across the entire build.

ForestEckhardt: How would y'all feel if for this RFC and the initial implementation of the buildpack we removed the ability to skip detection via and environment variable. I think that we could hold off for now and survey the landscape of language module BOM generating buildpacks and make a more informed decision on what kind of control mechanism we would like. I think that this is a relatively safe approach because if we get user feedback in the interim for such a feature this control structure is easily added and will help us solidify a pattern for similar types of buildpacks that may come in the future.

arjun024: If the bom generation do not eat up a lot time, I'm fine with it. You can also propose the skip-detection in the RFC and not do it in the initial implementation if you want.

sophiewigmore: @arjun024 we have updated the RFC to mention this environment variable as a future optimization pending user interest. We are leaving it out of our initial implementation

text/nodejs/0013-install-bom.md
sophiewigmore: Since we are no longer implementing this environment variable, I don't think this is necessary now?

arjun024 (approved):

text/nodejs/0013-install-bom.md
```diff
@@ -0,0 +1,161 @@
...
+
+Using the CycloneDX tool to generate the BOM will involve some changes to the
+Node.JS language family buildpack order groups.
+
+* A single BOM generation buildpack will be added. It's primary functions are:
+  * perform the BOM dependency tool retrieval from the dep-server,
+  * execute BOM generation with the tool (`cyclonedx-bom -o bom.json`), and
+  * contribute the BOM to the `launch.toml` `[bom]` label.
+
+* For each of the three Node.JS buildpack order groups, the BOM generation
```
ekcasey, line 78: I wonder if instead of generating the BOM during the build phase of the BOM generation buildpack, we should instead have the buildpack that produces installs the modules generate the BOM. There are a couple reasons for this:

Once CNB introduces the ability [associate BOM entries with a specific layer](https://github.com/buildpacks/rfcs/blob/main/text/0087-bom-in-layer-metadata.md), we will want to:
1. e.g. associate these BOM entries with a layer like paketo-buildpacks/yarn-install:modules
2. save time by avoiding rerunning the tool if the layer has not changed (the yarn-install buildpack should be responsible for making this determination)

In the case where there is no package manage we will still need the new buildpack to generate the BOM, or I guess node-start could potentially do it?

text/nodejs/0013-install-bom.md
```diff
@@ -0,0 +1,161 @@
...
+  buildpack will run directly before the start command buildpack. This means:
+  * directly before the `yarn-start` buildpack in the Yarn order group,
+  * directly before the `npm-start` buildpack in the NPM order group, and
+  * directly before the `node-start` buildpack in the no-package-manager order group.
+
+#### Detection
+
+The Node Module BOM Generator CNB always detects with the following contract:
+  * Requires {`node`, `node_modules`} during `build`
+  * Provides none
```
ekcasey, line 88: Seems like it should provide cyclonedx-bom. If we take my sugestion above regarding, shifting responsibility for executing the tool to the buildpack providing the described layer (when appropriate), then
1. if `node_modules` is vendored, it will also require cyclonedx-bom
2. yarn-install and npm-install can require cyclonedx-bom in their first build plan entry (and fall back to passing without generating a bom in a second entry).

text/nodejs/0013-install-bom.md
ForestEckhardt: I think that we need a new buildpack regardless and as it stands we can get away with having one buildpack do all of the BOM generation. I would prefer to implement the smallest change first and then potentially expand the functionality of other buildpacks or add additional buildpacks as the need arises. I have some concerns about having to duplicate logic across three different buildpacks just so I have the BOM on the right layer and so I would like to dive into this deeper when the feature is ready to play.

Is the lack of consideration for layer specific BOM a blocker for this initial implementation in your eyes?

text/nodejs/0013-install-bom.md
ForestEckhardt: I think this is correct if we do divide up the generation amongst buildpacks, however I am hesitant to do this fo the reasons mentioned above.

text/nodejs/0013-install-bom.md
sophiewigmore: We just added a performance section, but when we checked the time it took to run the tool against an app with 1300 node modules, it only took around 3 seconds. It's pretty quick, so we're not overly worried about the re-running of the tool

ekcasey: > it only took around 3 seconds. It's pretty quick, so we're not overly worried about the re-running of the tool

While I don't think 3 seconds is a deal-breaker, it's not completely insignificant either. For example a no-op rebuild of the paketo yarn sample app currently takes about 4.6 seconds on my workstation.

>  I have some concerns about having to duplicate logic across three different buildpacks just so I have the BOM on the right layer and so I would like to dive into this deeper when the feature is ready to play.

My assumption would be that the vast majority of the logic lives in the bom generation tool itself, leaving very little logic that would need to be duplicated?

> Is the lack of consideration for layer specific BOM a blocker for this initial implementation in your eyes?

No. In general I don't see my suggestions here as a blocker. I think associating the BOM with the correct layer and shaving a few seconds off the build whenever possible are both good long term goals, but if the node maintainers believe that accomplishing those goals would significantly slow down our ability to deliver ther feature or make the buildpacks harder to maintain, I trust the maintainers weigh those tradeoffs. This also seems like something that could be refactored later if need be without impacting end users, which lowers the stakes.

sophiewigmore: > While I don't think 3 seconds is a deal-breaker, it's not completely insignificant either. For example a no-op rebuild of the paketo yarn sample app currently takes about 4.6 seconds on my workstation.

This is fair. The app we used is quite large so I'd expect that to be more of a worst-case time. Regardless, you are right though that it's significant enough that we should probably consider optimizing.
 
> My assumption would be that the vast majority of the logic lives in the bom generation tool itself, leaving very little logic that would need to be duplicated?

Right now the logic includes the installation of the tool (/cache reuse of the tool), running the tool, and converting the BOM JSON file into `packit.BOMEntry` types (for now). 

It's not terrible to duplicate, but the fact of the matter is that we can get the basic functionality for the BOM via a single buildpack, rather than adding a separate buildpack for one case, and making changes to NPM Install and Yarn Install in the other cases. The optimizations for build time and layer inclusion are definitely worthwhile, but I am more inclined to go with the basic implementation that we have outlined first (pending maintainer approval) and then work towards these optimizations down the line as you said.

@paketo-buildpacks/nodejs-maintainers can you weigh in?

arjun024: SGTM for the first implementation. If the `node_modules` hasn't changed, you should still be able to save time by avoiding downloading/rerunning the tool, right?

ForestEckhardt: Correct

thitch97: I'm also on board with the single-buildpack proposal as a first pass

arjun024 (approved):

PR #100, opened 2021-08-12
Proposal to add support for rustup

dmikusa-pivotal:
## Summary

Rustup is the official and primary way to install Rust. This proposal is to add support for Rustup to the Rust buildpacks.

[Readable version](https://github.com/paketo-buildpacks/rfcs/blob/a76c816c42cf865b6d79b963fa41275f68f27d1b/text/rust/0000-rustup.md)

## Use Cases

This will unlock using nightly Rust, as well as provide the `rustup` binary for other potential uses.

## Checklist

### DIFF
--- a/text/rust/0000-rustup.md
+++ b/text/rust/0000-rustup.md
@@ -1,0 +1,43 @@
+# Paketo Community Rustup Buildpack
+
+## Summary
+
+The primary way of installing Rust is through a tool called `rustup`. It would be nice if we can support using `rustup` as another means of installing Rust.
+
+## Motivation
+
+We have bypassed using `rustup` in the rust-dist CNB by directly downloading a full Rust distribution and installing that. This has limitations such as the file is large, it's specific to an architecture & toolchain, and it's not feasible to do nightly builds that way. Using the `rustup` tool provides us with additional flexibility, plus it gives users the primary means through which they are used to having Rust installed. If a user needs to customize or make some adjustments, they are much more likely to be familiar with doing so through `rustup`.
+
+## Detailed Explanation
+
+Right now we have two buildpacks: rust-dist and cargo install. The former provides the Rust installation and the latter uses Cargo to compile source code. This proposal would add a third buildpack called rustup. Users would pick to either use rust-dist or rustup. The default would be to use rustup, given that is what most users will expect and it's more flexible. We'll continue to retain rust-dist as this may be easier for some users in offline environments.
+
+## Rationale and Alternatives
+
+- Do nothing. We'll miss out on use cases we could service.
+- Update rust-dist to cover some of the gaps, such as having nightlies and being able to install additional toolkits. This would be technically possible but would require some extreme automation, given that nightlies come out, well, nightly (which would imply we'd do the same with rust-dist). It would also require more entries in buildpack.toml if we're to support more toolkits.
+
+## Implementation
+
+The idea is to have a build plan like this:
+
+1. The rustup buildpack will provide both rust and rustup.
+2. The rust-dist buildpack will not change and will continue to provide rust.
+3. The cargo-install buildpack will change and will now provide two buildplans. Both will provide and require cargo. The first will require rust. The second will require rust and rustup.
+
+The idea is that we have a buildplan which will allow for either rust-dist and cargo-install to be used together or we'll have a buildplan that allows for rustup and cargo-install to be used together. When the rustup buildpack runs, it will consume the rustup and rust requirements from the buildplan and so the rust-dist buildpack while it still runs will effectively do nothing (i.e. it only installs rust if the rust requirement has not been yet met). Optionally, a builder could remove rust-dist or rustup if the builder author had a preverence for one over the other.
+
+Since a user cannot manipulate the buildplan directly, we are exposing two ways that a user can control which method is used to install Rust.
+
+1. Through an environment variable. We will add `BP_RUSTUP_ENABLED` which the rustup buildpack will examine. If it's true, which is the default, then the buildpack will proceed to use Rustup to install Rust. If it's set to false, the buildpack will do nothing, which will allow for some other buildpack, presently that's just rust-dist, to install Rust.
+2. Through a custom builder. A user could compose their own builder without the rustup buildpack, which would allow them to select a different buildpack for installing Rust.
+
+## Prior Art
+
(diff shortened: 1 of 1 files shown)


text/rust/0000-rustup.md
```diff
@@ -0,0 +1,43 @@
...
+
+The idea is that we have a buildplan which will allow for either rust-dist and cargo-install to be used together or we'll have a buildplan that allows for rustup and cargo-install to be used together. When the rustup buildpack runs, it will consume the rustup and rust requirements from the buildplan and so the rust-dist buildpack while it still runs will effectively do nothing (i.e. it only installs rust if the rust requirement has not been yet met). Optionally, a builder could remove rust-dist or rustup if the builder author had a preverence for one over the other.
+
+Since a user cannot manipulate the buildplan directly, we are exposing two ways that a user can control which method is used to install Rust.
+
+1. Through an environment variable. We will add `BP_RUSTUP_ENABLED` which the rustup buildpack will examine. If it's true, which is the default, then the buildpack will proceed to use Rustup to install Rust. If it's set to false, the buildpack will do nothing, which will allow for some other buildpack, presently that's just rust-dist, to install Rust.
```
ekcasey, line 32: Does using rustup by default mean that offline builds will not work by default even if the rust-dist buildpack has a vendored copy of the desired version?

text/rust/0000-rustup.md
```diff
@@ -0,0 +1,43 @@
...
+- Do nothing. We'll miss out on use cases we could service.
+- Update rust-dist to cover some of the gaps, such as having nightlies and being able to install additional toolkits. This would be technically possible but would require some extreme automation, given that nightlies come out, well, nightly (which would imply we'd do the same with rust-dist). It would also require more entries in buildpack.toml if we're to support more toolkits.
+
+## Implementation
+
+The idea is to have a build plan like this:
+
+1. The rustup buildpack will provide both rust and rustup.
+2. The rust-dist buildpack will not change and will continue to provide rust.
+3. The cargo-install buildpack will change and will now provide two buildplans. Both will provide and require cargo. The first will require rust. The second will require rust and rustup.
```
ekcasey, line 26: Is there any reason a user would prefer a `rustup` install if the rust-dist buildpack has the correct version? If not I wonder if we could do something like:

1. Have cargo-install just require rust via the build plan
1. Place rust-dist buildpack before the rustup buildpack in the order
1. Have rust-dist detect and provide rust if the required version/os/arch is supported
2. Have rustup always detect and provide rust

Therefore, if rust-dist can satisfy the requirements it can run first and claim the plan entry, such that the rustup buildpack is essentially a no-op during build. If rust-dist cannot provide the required rust, rustup will take over.

The benefits of this would be
1. if the correct rust version is vendored in an offline version of the rust-dist buildpack users will get that version resulting in faster builds that work in an airgapped environment
1. users don't need to think about where rust is coming from, they just specify the version
1. cargo-install doesn't need to be aware of the different ways upstream buildpacks provide rust

text/rust/0000-rustup.md
dmikusa-pivotal: It depends. Handling offline support for Rust could be done in a few different ways. Primarily `rustup` will pull downloads from a repository and by default that requires the Internet. You can host those binaries on your Intranet though, so you could just point `rustup` to your own server with `RUSTUP_DIST_SERVER` that doesn't require Internet and continue using `rustup`.

I think that would be the best way to support an offline environment because you can a.) still use `rustup` and b.) it allows full control of what versions are available to rustup, including nightly.

If you wanted the more traditional offline approach we support, where you can bundle rust into the image then you could disable `rustup` and use the rust-dist buildpack. I haven't tested to see what happens if `rustup` fails to install Rust but my guess is that the `rustup` CNB would fail and the build would then fail.

An option would be to not fail the build if `rustup` fails. If `rustup` exit's non-zero, we could log a warning and continue instead. So long as the `rustup` CNB doesn't consume the `rust` build plan entry, I'd expect `rust-dist` to run and supply rust.

Do you have a preference on how that behavior would work?

text/rust/0000-rustup.md
dmikusa-pivotal: I think the main reason to use `rustup` is for nightly support. We can't do that with `rust-dist`. There's no way to specify nightly with `rust-dist` at the moment, so I'm not sure how we'd detect if someone wanted nightly. I suppose we could overload the `BP_RUST_VERSION`, allow it to be set to `nightly`. Splitting hairs a bit but nightly isn't technically a version, it's a different toolchain (stable vs nightly). We could also add `BP_RUST_TOOLCHAIN` but it wouldn't make a lot of sense for `rust-dist` which would only ever have `stable`. 

Ancillary benefits of using `rustup` by default. It's the typical way that people install Rust, so most are already familiar with it and probably what they expect to be happening. It has a lot of additional functionality too like multiple toolchain support, additional target support, additional component support, and offline support. The `rustup` buildpack allows setting additional flags to `rustup` and env variables so a lot of additional functionality is available for power users. If we made rust-dist first, then power users would have to opt-out of it before unlocking additional functionality.

What about something like this:

- Order:  rustup, rust-dist, cargo-install
- Rustup will contribute rust
- Rust-dist will contribute rust
- Cargo-install will require rust

If Cargo should not care how Rust is installed, which seems reasonable cause it really doesn't care, then we can leave out the `rustup` build plan entry. Rustup could directly examine `BP_RUSTUP_ENABLED` for opt-out support, if false it would just omit the `rust` build plan entry which would result in `rust-dist` running. Then cargo-install wouldn't need to change. That would simplify a bit.

ekcasey: Your rationale here is convincing to me. I like your suggestion for the simplified build plan which removes the `rustup` entry.

> We could also add BP_RUST_TOOLCHAIN but it wouldn't make a lot of sense for rust-dist which would only ever have stable

Sounds like we may want to add this configuration option (in another RFC?) and have it switch the toolcahin (in the `rustup` case) or cause the buildpack to opt-out during detection (in the `rust-dist` case).

dmikusa-pivotal: For `rustup` you can change the default toolchain it installs via cli args, which are exposed through an env variable already, so I don't think it would need an env specific to the toolchain. At least not from the beginning. Down the road we might want to offer more configurability around what toolchains are installed, like if someone want's multiple toolchains.

dmikusa-pivotal: I made the suggested changes here.

ForestEckhardt:
I am on board with this RFC :+1:

ForestEckhardt (approved):

PR #102, opened 2021-08-17
Clojure tools buildpack

eddumelendez:
[Readable](https://github.com/paketo-buildpacks/rfcs/blob/6f8164a215f6c1e14b170b16dd457796cb91784f/text/java/0000-clojure-tools.md)


text/java/0003-clojure-tools.md
```diff
@@ -0,0 +1,33 @@
+# Clojure Tools Buildpack
```
dmikusa-pivotal, line 1: Can you rename this file to `0000-clojure-tools.md`? We usually keep the number as 0000 until it's approved, then it gets a number assigned at approval time.

text/java/0003-clojure-tools.md
```diff
@@ -0,0 +1,33 @@
...
+[t]: https://clojure.org/guides/tools_build
+[b]: https://github.com/eddumelendez/clojure
+
+## Motivation
+
+This is an additional buildpack, for building JVM-based applications based on Clojure, and offers yet another workload that can be built by Paketo buildpacks.
+
+## Detailed Explanation
+
+N/A
```
dmikusa-pivotal, line 17: Can you expand on what needs to be changed to transfer ownership to Paketo? Either in Details Explanation or under Implementation. See [here](https://github.com/paketo-buildpacks/rfcs/blob/4ad0d65ef1662fb5de2871e64013deddc4f529f4/text/java/0001-microsoft-jdk.md#implementation). It should be very similar to that section.

text/java/0003-clojure-tools.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Rationale and Alternatives
+
+N/A
+
+## Implementation
+
+This buildpack will be adopted by the Java Buildpacks team and transferred to the `paketo` GitHub organization.  The original author will retain full commit rights to the repository and be entitled to a full say in the direction and features of the buildpack.
```
dmikusa-pivotal, line 25: Would you prefer to be added as a contributor? We had this discussion previously, but with the number of buildpacks you've contributed it probably makes sense to just add you as a contributor to the team. Thoughts?

dmikusa-pivotal (changes requested):

text/java/0003-clojure-tools.md
dmikusa-pivotal: Cool. This change looks good. We'll have to follow governance and call for a vote to add you as a contributor, but I don't think there should be a problem with that. Thanks again for all your contributions!

dmikusa-pivotal:
Looks good to me.

dmikusa-pivotal (approved):

ekcasey:
Hey @eddumelendez, thanks for another great contribution! FYI I think the "Readable" link in your comment is pointing to an old draft of the RFC.

text/java/0000-clojure-tools.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Detailed Explanation
+
+The Clojure Tools Buildpack will be moved into the paketo-buildpacks Github organization and added to the Java sub-team. Maintenance for this buildpack will be provided by the Java subteam.
+
+The following changes will be made:
+
+* [Clojure Tools Buildpack][b] moved to Paketo Buildpacks org
+* Buildpack will have paketo-buildpacks/clojure ID
+* Buildpack will be published to gcr.io/paketo-buildpacks/clojure
```
ekcasey, line 23:
```suggestion
* Buildpack will have paketo-buildpacks/clojure-tools ID
* Buildpack will be published to gcr.io/paketo-buildpacks/clojure-tools
```

Maybe `clojure-tools` is more explicit? It, doesn't seem like a common use case but one could theoretically [build a clojure project with maven](https://github.com/talios/clojure-maven-plugin) or another tool chain.

ekcasey (approved):

eddumelendez:
@ekcasey thanks! I have accepted your suggestion and also the link in the PR description is updated.

eddumelendez:
As previously discussed with @dmikusa-pivotal , I have added the site to scrap the latest version for clojure tools and a note about using the real version `1.10.3.943` but getting an error when using the buildpack.

text/java/0000-clojure-tools.md
```diff
@@ -0,0 +1,51 @@
...
+## Detailed Explanation
+
+The Clojure Tools Buildpack will be moved into the paketo-buildpacks Github organization and added to the Java sub-team. Maintenance for this buildpack will be provided by the Java subteam.
+
+The following changes will be made:
+
+* [Clojure Tools Buildpack][b] moved to Paketo Buildpacks org
+* Buildpack will have paketo-buildpacks/clojure-tools ID
+* Buildpack will be published to gcr.io/paketo-buildpacks/clojure-tools
+* Buildpack will have go module of github.com/paketo-buildpacks/clojure
```
dmikusa-pivotal, line 24:
```suggestion
* Buildpack will have go module of github.com/paketo-buildpacks/clojure-tools
```

This should also be clojure-tools

text/java/0000-clojure-tools.md
```diff
@@ -0,0 +1,51 @@
...
+N/A
+
+## Unresolved Questions and Bikeshedding
+
+* During the implementation version `1.10.3.943` was not allowed, the message said 
+ 
+```
+[builder]   unable to parse version 1.10.3.943
+[builder]   Invalid Semantic Version
+```
```
dmikusa-pivotal, line 51: @ekcasey - What are your thoughts on this? I know that our tools expect a major.minor.maint version number, which is why this was a problem. I think we could probably have the CI compress the version into that format by doing `1.10.3-943`. I think the dash would be OK and there would still be a difference in the versions.

What do you think? I know you've had to deal with this before. Thanks

text/java/0000-clojure-tools.md
```diff
@@ -0,0 +1,51 @@
...
+
+The following changes will be made:
+
+* [Clojure Tools Buildpack][b] moved to Paketo Buildpacks org
+* Buildpack will have paketo-buildpacks/clojure-tools ID
+* Buildpack will be published to gcr.io/paketo-buildpacks/clojure-tools
+* Buildpack will have go module of github.com/paketo-buildpacks/clojure-tools
+* Github Actions will be added through [pipeline-builder][p]
+* An action will be added to [pipeline-builder][p] with the purpose of checking for new versions of [Clojure Tools][c]
+
```
dmikusa-pivotal, line 27: @eddumelendez Can you submit an example order group for the [paketo-buildpacks/java composite buildpack](https://github.com/paketo-buildpacks/java/blob/main/buildpack.toml#L30-L57)?

Where does this buildpack need to fit? I'm guessing in a spot similar to maven/gradle, but would like to confirm. Thanks

eddumelendez: between leiningen and gradle would be a good fit

```
{ id = "paketo-buildpacks/leiningen",                  version="3.2.2", optional = true },
{ id = "paketo-buildpacks/clojure-tools",            version="1.0.0", optional = true },
{ id = "paketo-buildpacks/gradle",                      version="5.4.0", optional = true },
```

dmikusa-pivotal: OK, sounds good to me.

dmikusa-pivotal:
I tested with the [two sample apps](https://github.com/eddumelendez/samples/tree/clojure_samples/java/tools-build) that were provided. Both worked for me.

I'm 👍 for this.

text/java/0000-clojure-tools.md
ekcasey: Our current practice is to truncate versions. For example Bellsoft Liberica `11.0.12+7` becomes `11.0.2` in the `paketo-buildpacks/bellsoft-liberica` [buildpack.toml](https://github.com/paketo-buildpacks/bellsoft-liberica/blob/main/buildpack.toml#L100-L101). 

I kind of hate that we do this... but I think we can stick with this strategy while we accept this buildpack and then, subsequently,  fix our tooling to accommodate the real versions in all of the java buildpacks.

Issue #103, opened 2021-08-17
Update hash algorithm in BOM schema


ForestEckhardt:
Currently the only hash algorithm that can be used in the BOM is SHA256, as it has a named field. However, this is extremely limiting especially if you are trying to use a hash from metadata to populate your BOM which could be any number of hashing algorithms. We should update the schema to allow you to specify both a hash as well as it's algorithm.

PR #105, opened 2021-08-18
RFC: Rework Hash Fields in BOM

ForestEckhardt:
[Readable](https://github.com/paketo-buildpacks/rfcs/blob/hashes-in-bom/text/0034-hash-field-bom.md)

Resolves https://github.com/paketo-buildpacks/rfcs/issues/103

### DIFF
--- a/text/0034-hash-field-bom.md
+++ b/text/0034-hash-field-bom.md
@@ -38,7 +38,7 @@ name = "<dependency name>"
   version = "<dependency version>"
 
 [bom.metadata.checksum]
-  algorithm = "<hashing algorithm>"
+  algorithm = "<CycloneDX supported hashing algorithm>"
   hash = "<dependency artifact hash from URI made using the specified algorithm>"
 
 [bom.metadata.source]
@@ -48,8 +48,14 @@ name = "<dependency name>"
   version = "<dependency source version>"
 
 [bom.metadata.source.checksum]
-  algorithm = "<hashing algorithm>"
+  algorithm = "<CycloneDX supported hashing algorithm>"
   hash = "<dependency source artifact hash from source URI made using the specified algorithm>"
+```
+
+As pointed out in the schema, the checksum algorithm should be a
+CycloneDX-supported hashing algorithm. The supported options are:
+```
+{ 'MD5'| 'SHA-1'| 'SHA-256'| 'SHA-384'| 'SHA-512'| 'SHA3-256'| 'SHA3-384'| 'SHA3-512'| 'BLAKE2b-256'| 'BLAKE2b-384'| 'BLAKE2b-512'| 'BLAKE3'}
 ```
 
 This style closely reflects what both
--- a/text/0034-hash-field-bom.md
+++ b/text/0034-hash-field-bom.md
@@ -1,0 +1,72 @@
+# Update Hash Field in Bill of Materials
+
+## Summary
+
+Currently, the only hashing algorithm that we support in the BOM (bill of
+materials) is `SHA256`, as the only checksum fields are explicitly named
+`sha256`. This shoehorns the hash algorithm to `SHA256` an alienates any hashes
+we can collect in other formats. This RFC proposes that the Paketo BOM supports
+hashes in any algorithm.
+
+## Motivation
+
+In the original BOM RFC it was proposed to just have fields named `sha256` as
+that was the only hashing algorithm used by dependencies hosted by Paketo.
+However, in the pursuit of enhancing our BOM by adding information about things
+such as node module information, we have started to use third party tools to
+help generate the BOM. These third party tools return industry-standard BOM
+formats that include a wide and varying array of hashes using varying hashing
+algorithms. This has led us to conclude that our BOM specification needs to
+update to be inclusive of a wide range of hashing algorithms to people that are
+not hosting the packages they use and the metadata attached.
+
+## Implementation
+
+We purpose the following change to the BOM schema:
+```
+[[bom]]
+name = "<dependency name>"
+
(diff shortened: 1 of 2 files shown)


ekcasey (approved):

text/0034-hash-field-bom.md
```diff
@@ -0,0 +1,72 @@
...
+  cpe = "<version-specific common platform enumeration>"
+  deprecation-date = "<dependency EOS date formatted in using RFC 3339>"
+  licenses = [<dependency license IDs(s) in SPDX Format>]
+  purl = "<package URL per github.com/package-url>"
+  summary = "<package summary>"
+  uri = "<compiled dependency URI>"
+  version = "<dependency version>"
+
+[bom.metadata.checksum]
+  algorithm = "<hashing algorithm>"
```
ryanmoran, line 41: Should we have a definition of what the values of the algorithm field are allowed to be? For example, CycloneDX specifies them like [this](https://cyclonedx.org/docs/1.3/#type_hashAlg).

text/0034-hash-field-bom.md
```diff
@@ -0,0 +1,72 @@
...
+such as node module information, we have started to use third party tools to
+help generate the BOM. These third party tools return industry-standard BOM
+formats that include a wide and varying array of hashes using varying hashing
+algorithms. This has led us to conclude that our BOM specification needs to
+update to be inclusive of a wide range of hashing algorithms to people that are
+not hosting the packages they use and the metadata attached.
+
+## Implementation
+
+We purpose the following change to the BOM schema:
```
ryanmoran, line 25:
```suggestion
We propose the following change to the BOM schema:
```

text/0034-hash-field-bom.md
```diff
@@ -0,0 +1,72 @@
+# Update Hash Field in Bill of Materials
+
+## Summary
+
+Currently, the only hashing algorithm that we support in the BOM (bill of
+materials) is `SHA256`, as the only checksum fields are explicitly named
+`sha256`. This shoehorns the hash algorithm to `SHA256` an alienates any hashes
```
ryanmoran, line 7:
```suggestion
`sha256`. This shoehorns the hash algorithm to `SHA256` and alienates any hashes
```

text/0034-hash-field-bom.md
```diff
@@ -0,0 +1,72 @@
...
+  arch = "<compatible processor architecture>"
+  cpe = "<version-specific common platform enumeration>"
+  deprecation-date = "<dependency EOS date formatted in using RFC 3339>"
+  licenses = [<dependency license IDs(s) in SPDX Format>]
+  purl = "<package URL per github.com/package-url>"
+  summary = "<package summary>"
+  uri = "<compiled dependency URI>"
+  version = "<dependency version>"
+
+[bom.metadata.checksum]
```
ryanmoran, line 40: Should we support multiple checksums? CycloneDX seems to specify that a single package might have multiple checksum values in their [spec documentation](https://cyclonedx.org/use-cases/#packaging-and-distribution).

text/0034-hash-field-bom.md
ForestEckhardt: Back to my old habits

text/0034-hash-field-bom.md
ForestEckhardt: I thought about this for a while. I am not sure that it is totally necessary as this is just metadata to be audited so if someone is using a more obscure hashing algorithm I would rather just support it out of the boxes then have to add support for it.

I am open to having a strictly defined list but don't know there is enough of a use case at the moment to make this a chief concern.

text/0034-hash-field-bom.md
ForestEckhardt: Again this is something I thought about but it would just be a representation of hashes for the same artifact using different algorithms if my understanding is correct. That kind of structure just feels kind of like overkill at the current moment and muddies the schema. I would be willing to go back and revise this at a later date though.

text/0034-hash-field-bom.md
ryanmoran: I guess what I am trying to avoid is having some BOMs say `sha256` and others say `SHA-256` to mean the same thing.

ekcasey: > I guess what I am trying to avoid is having some BOMs say sha256 and others say SHA-256 to mean the same thing.

+1

ForestEckhardt: In that case does it just make sense too pull directly from CycloneDX which is:
```
{ 'MD5'| 'SHA-1'| 'SHA-256'| 'SHA-384'| 'SHA-512'| 'SHA3-256'| 'SHA3-384'| 'SHA3-512'| 'BLAKE2b-256'| 'BLAKE2b-384'| 'BLAKE2b-512'| 'BLAKE3'}
```

sophiewigmore: SPDX options: ` SHA1, SHA224, SHA256, SHA384, SHA512, MD2, MD4, MD5, MD6` from https://spdx.github.io/spdx-spec/3-package-information/#310-package-checksum

ForestEckhardt: I think that it probably make more sense to go with the CycloneDX specification as it is more comprehensive and the upstream is doing more of a CycloneDX first policy.

sophiewigmore: @ryanmoran @ekcasey we have gone with CycloneDX's specification as Forest pointed out. The RFC now includes this specification.

ryanmoran (approved):

ekcasey (approved):

PR #107, opened 2021-08-25
Add initial version of rfc for semantic version tags

c0d1ngm0nk3y:
## Summary

Add a `rfc` to use tags in `buildpacks` to use semantic versioning.

## Use Cases

Adding support for using automatic security updates.

### DIFF
--- a/text/0000-semantic-version-tags.md
+++ b/text/0000-semantic-version-tags.md
@@ -6,7 +6,7 @@ To allow users to use the semantic versioning of buildpacks (see [rfc 0029](./00
 
 ## Motivation
 
-Semantic versioning for buildpacks was introduced with [rfc 0029](./0029-semantic-versioning.md), but the buildpacks have only a tag for the complete version (`MAJOR.MINOR.PATCH`). So any user can either use this tag and got no updates at all or the user omits the tag and receives all updates including breaking changes automatically.
+Semantic versioning for buildpacks was introduced with [rfc 0029](./0029-semantic-versioning.md), but the buildpacks have only a tag for the complete version (`MAJOR.MINOR.PATCH`). So any user can either use this tag and get no updates at all or the user omits the tag and receives all updates including breaking changes automatically.
 
 With the use of the semantic versioning, it would be good to allow the user to specify which updates should be taken automatically.
 
--- a/text/0000-semantic-version-tags.md
+++ b/text/0000-semantic-version-tags.md
@@ -30,7 +30,7 @@ When a minor version is updated `1.2.3` -> `1.3.0`
 * Tag `1` should address the version `1.3.0`
 * Tag `1.3` should address the version `1.3.0`
 * Tag `1.3.0` should address the version `1.3.0`
-* Tag `latest` should address the version `1.3.0`
+* Tag `latest` should address the version `1.3.0` (if v1 is still the latest one)
 
 This allows users to safeguard from incompatible changes and even minor changes, but still receive security updates.
 
@@ -41,7 +41,7 @@ When a patch version is updated `1.2.3` -> `1.2.4`:
 * Tag `1` should address the version `1.2.4`
 * Tag `1.2` should address the version `1.2.4`
 * Tag `1.2.4` should address the version `1.2.4`
-* Tag `latest` should address the version `1.2.4`
+* Tag `latest` should address the version `1.2.4` (if v1.2 is still the latest one)
 
 This allows users to pin the version and to be sure that the image is the same.
 
@@ -64,4 +64,4 @@ This is a common pattern and often seen for docker images. A good example is the
 
 ## Unresolved Questions and Bikeshedding
 
-N/A
+* Backporting security or bug fixes to older releases is not addressed by this rfc.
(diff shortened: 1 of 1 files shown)


ryanmoran:
Thanks for the RFC. Quick question: it feels like some parts of the RFC are maybe hinting at this, but do you expect that users pinned to an older major or minor of a buildpack would still receive security patches or feature updates, even once a newer major or minor has been released? I just want to make sure we aren't implying any sort of backporting that we don't intend to support.

sophiewigmore:
Reading through this RFC as a maintainer of a couple buildpacks, I generally support doing this especially now that we have the semantic versioning rules for buildpacks firmly set in  https://github.com/paketo-buildpacks/rfcs/blob/main/text/0029-semantic-versioning.md. 

My only slight concern is that I recommend users use the latest versions of our buildpacks, even when there are breaking changes in order to make the most of new feature and patches we ship. I worry that doing this will encourage users not to use the full breadth of buildpack offerings, and stay pinned to old versions. @ryanmoran / anyone else have thoughts here?

fg-j:
@sophiewigmore I think it's probably fine if users choose to pin to a major or minor and update on a cadence of their choosing. We can encourage being on `:latest`, and certainly tell people to upgrade if they're encountering issues, but I can understand why a dev team might want to pin to a minor and update it on a predictable cadence.

If we implement this, we should certainly document that tags won't get backported dependency updates.

text/0000-semantic-version-tags.md
```diff
@@ -0,0 +1,67 @@
...
+This is a common pattern and often seen for docker images. A good example is the [Official Python Images](https://hub.docker.com/_/python?tab=tags).
+
+`python` is referring to the overall latest image available.
+`python:3` is pinning the major version to 3, but the latest minor and patch versions are used.
+`python:3.6` is pinning the major version to 3, the minor version to 6 but the latest patch version is used.
+`python:3.6.14` is pinning the exact version.
+
+## Unresolved Questions and Bikeshedding
+
+* Backporting security or bug fixes to older releases is not addressed by this rfc.
```
ryanmoran, line 67:
```suggestion
```

text/0000-semantic-version-tags.md
```diff
@@ -0,0 +1,67 @@
+# Semantic Versioning in Tags for Buildpacks
+
+## Summary
+
+To allow users to use the semantic versioning of buildpacks (see [rfc 0029](./0029-semantic-versioning.md)) in a convenient way, a buildpacks must have tags not only in the format `MAJOR.MINOR.PATCH`, but also `MAJOR.MINOR` and `MAJOR`.
```
ryanmoran, line 5:
```suggestion
To allow users to use the semantic versioning of buildpacks (see [rfc 0029](./0029-semantic-versioning.md)) in a convenient way, a buildpacks must have tags not only in the format `MAJOR.MINOR.PATCH`, but also `MAJOR.MINOR` and `MAJOR`.

NOTE: Backporting security or bug fixes to older releases is not addressed by this RFC.
```

text/0000-semantic-version-tags.md
```diff
@@ -0,0 +1,67 @@
...
+
+This allows users to pin the version and to be sure that the image is the same.
+
+## Rationale and Alternatives
+
+The only alternative is to keep everything as is. In such case we will miss the points described in the [Motivation](#motivation)
+
+## Implementation
+
+TBD
```
ryanmoran, line 54:
```suggestion
The current workflows that push buildpack images to their registry locations already tag those images with the values `latest` and `MAJOR.MINOR.PATCH`. These workflows should be updated to also include `MAJOR.MINOR` and `MAJOR` tags.
```

ryanmoran (changes requested):

dmikusa-pivotal (commented):
I'm +1 for the tag format. It seems totally reasonable. I don't think the implementation would be that difficult either.

I'm wondering if anyone has tried this out to see how it would work with pack and other platforms? 

I haven't tried it, but some thoughts.

- I suspect it will work OK as arguments to `pack build`, if you use the `-b` flag and prefix your image with the repo name, like `docker.io/paketo-buildpacks/bellsoft-liberica:8`. When you use that format, pack will pull an image and add it to a dynamically created builder.
- I'm skeptical this format would work when creating composite images or builder images. My experience is that pack is very picky about the version in buildpack.toml and package.toml matching up. While you might be able to say something like `docker.io/paketo-buildpacks/bellsoft-liberica:8` in package.toml, if the corresponding version in buildpack.toml doesn't match up, I think it's going to fail.

I also wonder what the intent is regarding this PR & builders (I didn't see it specifically mentioned). I think a lot of users do not actually consume buildpacks directly, they are consuming a builder and the buildpacks it contains (this is the default with `pack build` if you don't add any `-b` flags). Are builders in scope or out of scope for this RFC? Is there a desire to have similar tagging with builders? 

Continuing on this thread, I wonder if we really need these labels on all buildpack images? or if it would make more sense to have them just on builders & composite CNBs, the things that users consume directly. Any thoughts on that?

loewenstein:
We haven't tried this indeed for creating builders, but had consumer facing usage in mind primarily.

I do think though, that at least some component buildpacks are consumer facing, if only the different JRE provinding buildpacks. If I understood the documentation correctly, they are supposed to be "injected" before the Java composite buildpack in order to replace the Bellsoft Liberica provided JREs.

dmikusa-pivotal (approved):

PR #110, opened 2021-09-28
Volume mount buildpack initial RFC

menehune23:
## Checklist

### DIFF
--- a/text/utilities/0000-volume-mount-buildpack.md
+++ b/text/utilities/0000-volume-mount-buildpack.md
@@ -1,0 +1,84 @@
+# Volume Mount Buildpack
+
+## Summary
+
+Implement a new buildpack that can inject volume-mounted files into the application directory.
+
+## Motivation
+
+Sometimes users have pieces of their application that, for one reason or another, they do not want to commit into their source code repository ([example](https://github.com/paketo-buildpacks/nginx/issues/281)).
+
+## Detailed Explanation
+
+Implementing this RFC would amount to creating a new repo within the Paketo Buildpacks GitHub organization to house the buildpack. The buildpack would then be eventually added to all groups (as optional) within Paketo builders to make it available for use.
+
+A user could then use the buildpack by supplying a volume mount during a build (e.g. `--volume` flag in `pack`, or a k8s volume mount). The user could then expect the contents of the volume to be copied into the application directory during the build phase of the proposed buildpack.
+
+For example, if a user has an application directory of the following structure:
+
+```
+<app-dir>
+├── package.json
+├── node_modules
+│   ├── ...
+```
+
+and they supplied a volume mount of the following structure:
+
+```
+/my-volume
+├── some-dir
+│   ├── sub-dir
+│   │   ├── some-file
+├── other-file
+```
+
+they could expect that their app directory would contain the following after the buildpack executes:
+
+```
+<app-dir>
+├── package.json
+├── node_modules
+│   ├── ...
+├── some-dir
+│   ├── sub-dir
+│   │   ├── some-file
+├── other-file
+```
+
+Note:
+- `my-volume` itself was not copied, just its contents.
+- If any file collisions occur, the volume-mounted files would take precedence.
+
+## Rationale and Alternatives
+
+We could leverage the [service binding spec](https://github.com/servicebinding/spec#workload-projection) was considered, especially since this is [now implemented in packit](https://github.com/paketo-buildpacks/packit/pull/228), and update buildpacks on a case-by-case basis (such as updating the nginx buildpack to search for a `nginx.conf` binding), but this would amount to individual buildpack complexity when a modular and more generic approach is available.
+
+## Implementation
+
+Related to the mention of service bindings above, we could also consider using service bindings for the implementation of the proposed buildpack, especially since `packit` supports resolving bindings already. However, there are some drawbacks to this:
+
+- Service bindings are not intended to become content of the built image, simply data to support integration of external services.
+- Service bindings happen to use volume mounts as an implementation detail, and therefore have strict requirements as to the folder structure within a binding. It would NOT allow generic folder structures to be injected into the application directory because of this.
+
+Instead, a custom utility buildpack could operate on environment variables:
+
+- `$BP_VOLUME_MOUNT_SOURCE`: specifies the source location of the volume, as an absolute path.
+- `$BP_VOLUME_MOUNT_DEST`: specifies the target directory -- relative to the application directory -- into which the volume contents will be copied.
+  - Refers to a directory, not a file.
+  - If any intermediate directories do not already exist within the application directory, they will be created (e.g. setting `$BP_VOLUME_MOUNT_DEST = foo/bar` will create `<app-dir>/foo/bar` then copy contents into `bar`).
+  - An empty or `.` value implies the application directory itself.
+
+### Detect
+
+Detection will pass if `$BP_VOLUME_MOUNT_SOURCE` is set. Otherwise, detection will fail.
+
+### Build
+
+During build, the glob `$BP_VOLUME_MOUNT_SOURCE/**` will be recursively copied to `$BP_VOLUME_MOUNT_DEST`.
+
+## Unresolved Questions and Bikeshedding
+
+- Should `$BP_VOLUME_MOUNT_SOURCE` have a default?
+
+- Should detection be based on the presence of the volume rather than the environment variable? (Note, would require `$BP_VOLUME_MOUNT_SOURCE` to have a default).


text/utilities/0000-volume-mount-buildpack.md
```diff
@@ -0,0 +1,86 @@
...
+
+During build, the glob `$BP_VOLUME_MOUNT_SOURCE/**` will be recursively copied to `$BP_VOLUME_MOUNT_DEST`.
+
+## Unresolved Questions and Bikeshedding
+
+- Should `$BP_VOLUME_MOUNT_SOURCE` have a default?
+
+- Should detection be based on the presence of the volume rather than the environment variable? (Note, would require `$BP_VOLUME_MOUNT_SOURCE` to have a default).
+
+- Detection: What if a later buildpack relies on a file that will be volume mounted? It wouldn't be available during detection as the build phase would not have executed yet.
```
menehune23, line 86: @ForestEckhardt I added this question based on your feedback today.

In thinking about it further, I think there's a simple solution: rely on the fact that mounted files will overwrite their source-provided counterparts. For instance, if the nginx CNB looks for `nginx.conf`, that file would still need to be included in the app source. It could be a dummy or perhaps use an include statement to include the to-be-mounted file.

That's specific to nginx, but in general, what if the original app dir provided by the user is simply still required to adhere to buildpack detection logic in isolation (i.e. without making any assumptions about other buildpacks dumping information into the app dir)?

ForestEckhardt: What is the advantage to allowing you to overwrite a file like that as opposed to just having a separate app? I am struggling to see why I would want this to be volume mounted in.

menehune23: That's outlined in the motivation section of this RFC, but in short it's about changes that users may want to "patch in" on a build-by-build basis. In that section, there's a [concrete example for nginx](https://github.com/paketo-buildpacks/nginx/issues/281).

menehune23: Also, addressing the detection concern, we could get around needing dummy files in the source by having buildpacks also check the volume during detection. I'm not sure I love that idea though as it sort of couples the buildpacks together, albeit loosely. If we wanted to go this route, we could perhaps add an API in `packit` to facilitate running detection in the volume if detection in the app dir fails (if `$BP_VOLUME_MOUNT_SOURCE` is set).

voor: It's also a question of personas.  As the platform team providing Cloud Native Buildpack capabilities to its tenants, we understand the requirements for running a static webpage inside of the environment, things like the nginx config file needs to not respond to OPTION queries or other nonsensical things that a person just creating a Hugo or Jekyll site wouldn't want to be concerned with.

So Team A, the platform team, can inject the nginx config from a centralized location and manage it.
Team B, the website team, can write their markdown files or other somesuch, and their builds work great.

ForestEckhardt: > Also, addressing the detection concern, we could get around needing dummy files in the source by having buildpacks also check the volume during detection. I'm not sure I love that idea though as it sort of couples the buildpacks together, albeit loosely. If we wanted to go this route, we could perhaps add an API in `packit` to facilitate running detection in the volume if detection in the app dir fails (if `$BP_VOLUME_MOUNT_SOURCE` is set).

This sounds like way more work than it would be worth and I think would add a lot of complexity to the detection process as a whole. I would much rather a dummy file than anything else but I still think that the API is clunky which makes it hard for me to get fully on board with the idea.

dmikusa-pivotal: I think there's another concern here. Even if you get past detection, a buildpack would be copying the files out of the volume mount at *build* time. If the volume mount changes at runtime, which is possible given the Nginx use case, that would not copy the files, unless the buildpack is adding a helper to copy the files before the app starts (it wouldn't for the nginx use case of a single small conf file, but in theory copy large amounts of files before the app starts could delay the app start and cause problems).

Is copying the files really necessary? Could a user just volume mount the file into the image directly? Assuming we publish the expected location for that file in our docs.

voor: I'm concerned the requirement to have a dummy file is effectively the equivalent of asking them to put the nginx.conf file into their code repository.  This is the scenario we're hoping to avoid entirely, they shouldn't need to create these files.

We could also look at something like copying the nginx.conf into the code repository elsewhere (like in GitLab CI), and instead of using a git repository as the source use the resulting "overlay" of files, but again this seems like a common problem better solved in buildpacks.

arjun024: @dmikusa-pivotal 
> One side effect of this is that the buildpack doesn't rewrite the file, so {{port}} doesn't get rewritten to the assigned port

Nginx uses a `profile.d` [here](https://github.com/paketo-buildpacks/nginx/blob/v0.3.3/build.go#L140) to run the templating at launch and and does rewrite the config [here](https://github.com/paketo-buildpacks/nginx/blob/v0.3.3/cmd/configure/main.go#L64). I tried with listen `{{env "NEWPORT"}}` in the `test.conf` in your command and passed `$NEWPORT` in the docker-run. It worked fine. Not sure I misinterpreted your comment.

If I'm not wrong, then a simple run-time volume-mount could solve the problem of overriding config files for nginx. Combined with Procfiles, users can also change the start command if they wish so. 

> You just need to have a directory with some nginx.conf file so it passes detection. The buildpack could ease this requirement if it chose to.

I'm hesitant about the idea of easing this requirement because nginx buildpack is also designed to be used in a builder directly, and this could block the detection of other buildpacks that come later in the order like php.

dmikusa-pivotal: >If I'm not wrong, then a simple run-time volume-mount could solve the problem of overriding config files for nginx. Combined with Procfiles, users can also change the start command if they wish so.

No, that's exactly what I was doing. Maybe the error I was getting, `failed to create nginx.conf: open /workspace/nginx.conf: permission denied` was coming from the `profile.d` script. That would explain why the template fields were not populated for me.

Did you get that error @arjun024? Are you using Docker Desktop for Mac?

I think this is my point in general. It doesn't seem like we need a buildpack to copy the files. A user can just volume mount them in directly.

>I'm hesitant about the idea of easing this requirement because nginx buildpack is also designed to be used in a builder directly, and this could block the detection of other buildpacks that come later in the order like PHP.

Yes, I think that's a concern. What does changing this requirement mean to the PHP CNBs? I haven't looked at them in a while so I don't really know.

It maybe that further logic is require in the nginx CNB such that it checks the buildplan or checks for the presence of PHP in some way. That way if it's running solo, it doesn't need to require nginx.conf.

I suppose other possibilities would be an env variable to bypass that or requiring the volume mount to also be present at build time. If you volume map in the nginx.conf at build time, that would pass detection. I don't think it would but would have to double check, if that will end up with the `nginx.conf` embedded in the image. I think that would be the question. If volume mounting into `/workspace` would result in the file from that mount being included in the produced image.

menehune23: @dmikusa-pivotal 
> I think there's another concern here. Even if you get past detection, a buildpack would be copying the files out of the volume mount at _build_ time. If the volume mount changes at runtime, which is possible given the Nginx use case, that would not copy the files

Over in the nginx issue, I was trying to differentiate the runtime vs build-time needs. What I was hearing is that the need was primarily at build-time. It's true that this solution is only for build-time, as the Procfile buildpack can help with runtime.

menehune23: @voor 

> We could also look at something like copying the nginx.conf into the code repository elsewhere (like in GitLab CI), and instead of using a git repository as the source use the resulting "overlay" of files, but again this seems like a common problem better solved in buildpacks.

I considered this as well. Honestly, I think that would be best, as patching in extra files before a build is really outside of the buildpack lifecycle/domain. I think we should fully investigate how hard something like that would be (simple in GitLab and GitHub Actions, but not sure how simple it is in k8s/kpack) before committing to implement this RFC.

arjun024: @dmikusa-pivotal 
I do not get that error. Docker Server 20.10.6 on Ubuntu

menehune23: @voor 

> I'm concerned the requirement to have a dummy file is effectively the equivalent of asking them to put the nginx.conf file into their code repository. This is the scenario we're hoping to avoid entirely, they shouldn't need to create these files.

I don't see how this need can be avoided without either:

- Altering the official lifecycle/CNB spec to support multiple volume-mounted app dirs, or
- Making buildpacks look in volumes during detection

menehune23: @dmikusa-pivotal 

> It doesn't seem like we need a buildpack to copy the files. A user can just volume mount them in directly.

I think I'm understanding this better now. You're saying that you literally volume mount individual files **into the workspace dir**? Seems like a workable solution to me, at least for the build-time case. (We really should talk about build-time separately from run-time, as this buildpack is focused on build-time -- nginx in specific might have desired behavior at runtime as well, but that might be better discussed in the [nginx issue](https://github.com/paketo-buildpacks/nginx/issues/281))

voor: Let me see if I'm following this then: if the buildpack was explicitly invoked like with https://github.com/voor/static-build-service-template/blob/create-react-app/manifests/create-react-app.yaml#L49-L53 then that wouldn't be a problem, and if someone wanted to automatically detect nginx they'd need the dummy file there?

I'm comfortable with that result if that's the case.

menehune23: @voor That seems right to me, though you might see if you could even volume mount the needed nginx file(s) directly into the workspace dir via your yml config, as @dmikusa-pivotal was suggesting.

menehune23: @arjun024 

> Nginx uses a profile.d here to run the templating at launch and and does rewrite the config here.

Thanks for that clarification! For some reason I thought the templating was occurring at build-time. That's great new. It sounds like you were also able to combine this with a volume mount into the app dir at runtime?

menehune23: Guess I broke my own suggestion by talking about nginx and runtime. Perhaps we should move the nginx-specific convo into the [nginx issue](https://github.com/paketo-buildpacks/nginx/issues/281)?

dmikusa-pivotal: What use cases would there be besides the Nginx one? Given that it seems like the Nginx one can be solved without this proposed buildpack.

menehune23: I'd say that the nginx case was the primary motivator for this -- we just didn't want to build something into that buildpack if it could be used more widely.

That said, if volume mounting into the app dir works for nginx, I think we should opt to close this RFC and not implement this right now.

ForestEckhardt: This might be dumb question so forgive my ignorance but what are devs using to test and view their projects before pushing them if there is no `nginx.conf` present in the app dir? Are the using other software than `nginx` or does `nginx` just have defaults that just kinda work? I was under the impression that you needed some form of `nginx.conf` to host a static site using `nginx` in any capacity.

PR #112, opened 2021-10-01
Adopt Oracle JDK Buildpack

eddumelendez:
[Readable](https://github.com/paketo-buildpacks/rfcs/blob/bd2807d48ba9b5e12ff0a370b831a3e3c07e5241/text/java/0000-oracle-jdk.md)

### DIFF
--- a/text/java/0000-oracle-jdk.md
+++ b/text/java/0000-oracle-jdk.md
@@ -35,7 +35,7 @@ The following changes will be made:
 - Buildpack will have `paketo-buildpacks/oracle-jdk` ID
 - Buildpack will be published to `gcr.io/paketo-buildpacks/oracle-jdk`
 - Buildpack will have go module of `github.com/paketo-buildpacks/oracle-jdk`
-- An action will be added to [pipeline-builder](https://github.com/paketo-buildpacks/pipeline-builder/tree/main/actions) with the purpose of checking for new versions of Oracle JDK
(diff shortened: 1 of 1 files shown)


text/java/0000-oracle-jdk.md
```diff
@@ -0,0 +1,38 @@
+# Paketo Buildpacks Oracle JDK Buildpack
+
+## Summary
+
+A [Oracle JDK Buildpack](https://github.com/eddumelendez/oracle-jdk) for supplying Oracle's distribution of OpenJDK has been created by [@eddumelendez][@eddumelendez]. It is built using common tooling (libpak and libjvm) used by the Paketo Java sub-team and should be adopted into the Paketo Buildpacks Github Org.
```
dmikusa-pivotal, line 5: I'm curious about the name. What would you think about just `oracle` or `oracle-java` for the name?

I don't know if they ever will, but my concern with the present name is that if Oracle starts releasing JREs, the name would be a little misleading.

dmikusa-pivotal (changes requested):

text/java/0000-oracle-jdk.md
eddumelendez: I would choose just `oracle`

dmikusa-pivotal: OK, let's go with `paketo-buildpacks/oracle` then.

dmikusa-pivotal (approved):

PR #113, opened 2021-10-07
Go BOM RFC

brayanhenao:
## Summary
This RFC contains all the information about the Golang BOM initiave.

## Checklist

### DIFF
--- a/text/go/0001-go-bom.md
+++ b/text/go/0001-go-bom.md
@@ -1,61 +1,80 @@
 # Go BOM Generation
 
+_NOTE: We may wish to move this RFC under the go-build directory if/when it is
+approved. We are keeping it here while the discussion is open in order to
+preserve history._
+
 ## Summary
 
-Add support for the generation of the Bill of Materials in the Go Language Family in order to provide entries for
-modules installed during the build process that can be found on the final app image.
+Add support for the generation of the Bill of Materials in the Go Language
+Family in order to provide entries for modules installed during the build
+process that can be found on the final application image.
 
 ## Motivation
 
-As a part of an initiative to provide a Bill of Materials (BOM) for all the Paketo Language Family Buildpacks, in
-previous months
-the [Node.JS BOM Module](https://github.com/paketo-buildpacks/rfcs/blob/main/text/nodejs/0013-install-bom.md) was
-developed.
+Previously, the
+[Node.JS Module BOM buildpack](https://github.com/paketo-buildpacks/rfcs/blob/main/text/nodejs/0013-install-bom.md)
+was developed to create a BOM for the nodejs language family. We would now like
+to add similar support for BOM generation in the Go language family.
 
-We would like to add support for BOM generation in the Go language family. We would like to create a new buildpack to
-provide this. We do not anticipate the need for changes to other buildpacks in the language family.
+We propose modifying the [`go-build` buildpack](https://github.com/paketo-buildpacks/go-build)
+to add BOM generation as an optional feature during the build process.
+Additionally, we propose to add a new buildpack called `cyclonedx-gomod` which
+provides the `cyclonedx-gomod` executable for this process to run.
 
-We propose to support generating BOMs only for projects which utilize vendoring via `go mod`. At this time we do not
-propose to support any other vendoring solutions
-(e.g. `dep` or files copied directly into the `vendor` directory without a
-`go.mod` file). Extending the new buildpack to support to other vendoring solutions could be a future RFC, if we
-discover a clear requirement for this.
+We propose to support generating BOMs only for projects which utilize vendoring
+via `go mod`. At this time we do not propose to support any other vendoring
+solutions (e.g. `dep` or files copied directly into the `vendor` directory
+without a `go.mod` file). Supporting other vendoring solutions could be a future
+RFC, if we discover a clear requirement for this. The primary reason for this is
+that the `cyclonedx-gomod` tool (see the Implementation section below) does not
+support generating BOMs for these alternative vendoring mechanisms.
 
 ## Implementation
 
 ### Tool details
 
-The Bill of Materials that will result should be an accurate accounting of all the modules that are defined and
-installed by `go modules`.
+The Bill of Materials that will result should be an accurate accounting of all
+the modules that are defined and installed by `go mod`.
 
-There already exists a tool called [CycloneDX Gomod](https://github.com/CycloneDX/cyclonedx-gomod) which is an official
-CycloneDX-supported tool that creates valid CycloneDX BOMs for Go applications. CycloneDX is a BOM format supported by
-the OWASP Foundation, and is a serious contender (at the time of this RFC creation) as a format we will eventually
-support.
+There already exists a tool called [CycloneDX Gomod](https://github.com/CycloneDX/cyclonedx-gomod)
+which is an official CycloneDX-supported tool that creates valid CycloneDX BOMs
+for Go applications. CycloneDX is a BOM format supported by the OWASP
+Foundation, and is a serious contender (at the time of this RFC creation) as a
+format we will eventually support.
 
-The CycloneDX Go Module tool is available as a pre-built binary for multiple platforms, and also it can easily be
-installed via `go install github.com/CycloneDX/cyclonedx-gomod`.
+The CycloneDX Go Module tool is available as a pre-built binary for multiple
+platforms, and also it can easily be installed via `go install
+github.com/CycloneDX/cyclonedx-gomod`.
 
 This tool supports three different cases of BOM generation:
 
-1. `app`: Include only those modules that the target application actually depends on. Modules required by tests or
-   packages that are not imported by the application are not included. Build constraints are evaluated, which enables a
-   very detailed view of what's really compiled into an application's binary.
-2. `mod`: Include the aggregate of modules required by all packages in the target module. This optionally includes
-   modules required by tests and test packages. Build constraints are NOT evaluated, allowing for a "whole picture" view
-   on the target module's dependencies.
-3. `bin`: Offers support of generating rudimentary SBOMs from binaries built with Go modules.
+1. `app`: Include only those modules that the target application actually
+   depends on. Modules required by tests or packages that are not imported by
+   the application are not included. Build constraints are evaluated, which
+   enables a very detailed view of what's really compiled into an application's
+   binary.
+2. `mod`: Include the aggregate of modules required by all packages in the
+   target module. This optionally includes modules required by tests and test
+   packages. Build constraints are NOT evaluated, allowing for a "whole picture"
+   view on the target module's dependencies.
+3. `bin`: Offers support of generating rudimentary SBOMs from binaries built
+   with Go modules.
 
-Of these, we are only interested in `app`. There are a few reasons why `mod` and `bin` are not suitable:
+Of these, we are only interested in `app`. There are a few reasons why `mod` and
+`bin` are not suitable:
 
 * The buildpacks do not currently support building modules without an associated
-executable (i.e. libraries) which is the primary use-case for `mod`.
-* The BOM produced by the `mod` command contains less information than the BOM produced by the `app` command.
-(e.g. the `mod` command does not include license info, or detailed info for each file).
-* The `mod` command does not respect flags provided to the `go build`
-  process at build time e.g. `CGO_ENABLED`, `GOFLAGS`.
-* The `bin` command is for pre-built executables, which is unnecessary for the buildpack as it has access to the source code
-* The `bin` command produces less detailed output than `app` as it does utilize the source code if it is present.
+  executable (i.e. libraries) which is the primary use-case for `mod`.
+* The BOM produced by the `mod` command contains less information than the BOM
+  produced by the `app` command.  (e.g. the `mod` command does not include
+  license info, or detailed info for each file).
+* The `mod` command does not respect flags provided to the `go build` process at
+  build time e.g. `CGO_ENABLED`, `GOFLAGS`.
+* The `bin` command is for pre-built executables, which is unnecessary for the
+  buildpack as it has access to the source code
+* The `bin` command produces less detailed output than `app` as it does utilize
+  the source code if it is present.
 
 The output BOM contains the following fields for each module:
 
@@ -68,24 +87,30 @@ The output BOM contains the following fields for each module:
 
 Note: The `license` field can be retrieved when the flag `-licenses` is passed to the command.
 
-Almost all of fields that we aim to support
-per [the Paketo BOM RFC](https://github.com/paketo-buildpacks/rfcs/blob/main/text/0033-bill-of-materials.md#overall-schema)
-are available. The only one that is not available is `description`, and this is because there is no standard way to
-capture this information in Go projects (as compared with other languages like nodejs and rust, where the package
-description files contain a field for this metadata). See unresolved questions below.
+Almost all of fields that we aim to support per
+[the Paketo BOM RFC](https://github.com/paketo-buildpacks/rfcs/blob/main/text/0033-bill-of-materials.md#overall-schema)
+are available. The only one that is not available is `description`, and this is
+because there is no standard way to capture this information in Go projects (as
+compared with other languages like nodejs and rust, where the package
+description files contain a field for this metadata). See unresolved questions
+below.
 
-In order to support this tool in both the online and offline cases, we will host the binary tool on
-the [dep-server](https://github.com/paketo-buildpacks/dep-server).
+In order to support this tool in both the online and offline cases, we will host
+the binary tool on the
+[dep-server](https://github.com/paketo-buildpacks/dep-server).
 
 ### Performance
 
-The addition of this tool to the language family order groupings might have an impactful change to the overall
-performance of the buildpacks.
+The addition of this tool to the language family order groupings might have an
+impactful change to the overall performance of the buildpacks.
 
-Performance metrics are still TBD, but for a typical project (e.g. `pack`) we observe that it takes about 10-30 seconds
-to construct a BOM.
+Performance metrics are still TBD, but for a typical project (e.g.
+[`pack`](https://github.com/buildpacks/pack)) we observe that it takes about
+10-30 seconds to construct a BOM.
 
-We can investigate ways to improve this performance.
+We can investigate ways to improve this performance. In the meantime, we propose
+including an environment variable: `BP_DISABLE_MODULE_BOM` (which defaults to
+`false`) to skip BOM generation if it takes an unacceptable duration.
 
 ### Limitations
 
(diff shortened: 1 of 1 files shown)


text/go/0001-go-bom.md
```diff
@@ -0,0 +1,164 @@
...
+
+As a part of an initiative to provide a Bill of Materials (BOM) for all the Paketo Language Family Buildpacks, in
+previous months
+the [Node.JS BOM Module](https://github.com/paketo-buildpacks/rfcs/blob/main/text/nodejs/0013-install-bom.md) was
+developed.
+
+We would like to add support for BOM generation in the Go language family. We would like to create a new buildpack
+to provide this. We do not anticipate the need for changes to other buildpacks in the language family.
+
+We propose to support generating BOMs only for projects which utilse vendoring via `go mod`.
```
arjun024, line 18:
```suggestion
We propose to support generating BOMs only for projects which utilize vendoring via `go mod`.
```

text/go/0001-go-bom.md
```diff
@@ -0,0 +1,164 @@
...
+`go.mod` file).
+Extending the new buildpack to support to other
+vendoring solutions could be a future RFC, if we discover a clear requirement for this.
+
+## Implementation
+
+### Tool details
+
+The Bill of Materials that will result should be an accurate accounting of all the modules that are defined and
+installed by `go modudules`.
```
arjun024, line 30:
```suggestion
installed by `go modules`.
```

text/go/0001-go-bom.md
```diff
@@ -0,0 +1,164 @@
...
+* hash,
+* source URI,
+* license ID,
+* and package URL
+
+Note: The `license` field can be retrieved when the flag `-licenses` is passed to the command.
+
+Almost all of fields that we aim to support
+per [the Paketo BOM RFC](https://github.com/paketo-buildpacks/rfcs/blob/main/text/0033-bill-of-materials.md#overall-schema)
+are available. The only one that is not available is `description`, and this is because there is no standard way to
```
arjun024, line 68: I do not see `description` even as a recommended field in the said RFC

text/go/0001-go-bom.md
```diff
@@ -0,0 +1,164 @@
...
+
+Using the CycloneDX tool to generate the BOM will involve some changes to the Go language family buildpack order
+groups as follows:
+
+* A single BOM generation buildpack will be added. It's primary functions are:
+    * perform the BOM dependency tool retrieval from the dep-server,
+    * execute BOM generation with the tool (`cyclonedx-gomod app -o bom.json`), and
+    * contribute the BOM to the `launch.toml` `[bom]` label.
+
+* For the Go Mod buildpack order group, the BOM generation buildpack will run directly before the `go-build` buildpack.
```
arjun024, line 96: Is there any particular reason this buildpack runs before go-build and not after? In my head, I think of doing accounting and writing a Bill of Materials as a last step one would do after all the build commands are completed.

Just for clarity, it would be nice if you can print out the entire order group here

text/go/0001-go-bom.md
```diff
@@ -0,0 +1,164 @@
...
+data. This is a viable option if installing a tool is not a reasonable choice.
+
+An alternative to performing all of these steps in a separate buildpack would be to add the CycloneDX tool installation
+as a step in the Go Build buildpack, and then run BOM creation alone in a buildpack before the start commands. This is
+less than ideal because it's not transparent or intuitive to install it during the go specific build process. It makes
+more sense to separate all module BOM creation logic into its own buildpack, so that users may easily opt out.
+
+## Unresolved Questions and Bikeshedding
+
+* How do we know which of the main commands we are building and hence which main package to generate the BOM?
```
arjun024, line 157: I think an idea would be to make use of `$BP_GO_TARGETS` ([link](BP_GO_TARGETS)) just like how `go-build` buildpack does. I'd say that every "main" target that will be subject to compilation should have its bom present as the user can dynamically set any of them as the container start-command during run.

text/go/0001-go-bom.md
brayanhenao: Fixed!

text/go/0001-go-bom.md
robdimsdale: No particular reason, except the BOM process doesn't require the build process to have run. We modeled this ordering off the nodejs BOM buildpack, and this is the analogous ordering to what that language family does.

We're happy to change the ordering if folks prefer that.

text/go/0001-go-bom.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Tool details
+
+The Bill of Materials that will result should be an accurate accounting of all the modules that are defined and
+installed by `go modules`.
+
+There already exists a tool called [CycloneDX Gomod](https://github.com/CycloneDX/cyclonedx-gomod) which is an official
+CycloneDX supported-tool that creates valid CycloneDX BOMs for Go applications. CycloneDX is a BOM format supported by
+the OWASP Foundation, and is a serious contender (at the time of this RFC creation) as a format we will eventually
+support.
```
sophiewigmore, line 34:
```suggestion
the OWASP Foundation, and is a format that we plan to support per the recently merged [upstream CNB RFC](https://github.com/buildpacks/rfcs/blob/main/text/0095-sbom.md).
```

text/go/0001-go-bom.md
```diff
@@ -0,0 +1,163 @@
...
+   very detailed view of what's really compiled into an application's binary.
+2. `mod`: Include the aggregate of modules required by all packages in the target module. This optionally includes
+   modules required by tests and test packages. Build constraints are NOT evaluated, allowing for a "whole picture" view
+   on the target module's dependencies.
+3. `bin`: Offers support of generating rudimentary SBOMs from binaries built with Go modules.
+
+Of these, we are only interested in `app`. The buildpacks do not currently support building modules, or libraries, which
+is the primary use-case for `mod`. The `bin` command is for pre-built executables, which is unnecessary for the
+buildpack as it has access to the source code. The BOM created by the `bin` command also contains less information than
+the one created via the `app` command.
```
sophiewigmore, line 52: Can you explain more why we are not interested in the `mod` case? Why wouldn't we want to include as much information about the dependency tree (and anything that comes into contact with the app) as possible?

text/go/0001-go-bom.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Language Family Additions
+
+Using the CycloneDX tool to generate the BOM will involve some changes to the Go language family buildpack order groups
+as follows:
+
+* A single BOM generation buildpack will be added. It's primary functions are:
+    * perform the BOM dependency tool retrieval from the dep-server,
+    * execute BOM generation with the tool (`cyclonedx-gomod app -o bom.json`), and
+    * contribute the BOM to the `launch.toml` `[bom]` label.
```
sophiewigmore, line 92: Might be worth noting that an extra conversion step will need to be performed between these two steps to convert the output JSON file to the legacy CNB TOML label format.

text/go/0001-go-bom.md
sophiewigmore: Side note, now that the upstream CNB RFC has been merged in, do we need to add onto this initiative to include transitioning the BOM format from the legacy CNB format to SPDX/CycloneDX in a top-level RFC addendum? Or is this something we will tackle later? cc @ForestEckhardt @simonjjones

text/go/0001-go-bom.md
sophiewigmore: This order grouping makes sense to me logically since the BOM creation directly will need the Go modules. +1 to getting the full order group in the RFC.

robdimsdale: @sophiewigmore which ordering do you prefer? The one that is written, or the proposed change? Either is fine by us.

Happy to add the full ordering here for clarity.

text/go/0001-go-bom.md
robdimsdale: Sure, we can add that.

text/go/0001-go-bom.md
ryanmoran: The RFC was approved, but there is no current support in the lifecycle for the new SBOM format. We can move forward with an RFC, but implementations would need to wait until there was lifecycle support.

text/go/0001-go-bom.md
ryanmoran: You would need to generate the BOM before the `go-build` buildpack runs because it removes all of the source from the working directory: https://github.com/paketo-buildpacks/go-build/blob/c40e20b61d57bed68095055814b10b74f4b1e617/build.go#L87-L90

After it executes, there won't be any `go.mod` or `go.sum` files to help with generating the BOM.

text/go/0001-go-bom.md
```diff
@@ -0,0 +1,163 @@
...
+## Alternatives
+
+An alternative to the option to using the CycloneDX BOM tool would be to hand-create the BOM as described. We could go
+through each go module directory, pull information out of each `go.mod` file and then build out our own BOM with that
+data. This is a viable option if installing a tool is not a reasonable choice.
+
+An alternative to performing all of these steps in a separate buildpack would be to add the CycloneDX tool installation
+as a step in the Go Build buildpack, and then run BOM creation alone in a buildpack before the start commands. This is
+less than ideal because it's not transparent or intuitive to install it during the go specific build process. It makes
+more sense to separate all module BOM creation logic into its own buildpack, so that users may easily opt out.
```
ryanmoran, line 151: The more I read of the RFC, the more I'm asking myself why we aren't pursuing this alternative. In this paragraph it appears that we have 2 concerns: 1) we need to install a tool, and 2) we want users to be able to opt-out.

For 1, I don't see why we couldn't introduce a buildpack that just does the tool installation and makes it available during build.
For 2, I also don't see why we need a separate buildpack to make this possible.

Can we elaborate a bit more on these concerns?

ryanmoran (commented):
The technical implementation proposed in the RFC seems sound. Still, I think the biggest concern I have with this, and really also with the previous Node.js BOM work, is that creating a separate buildpack that is responsible for the BOM decouples the artifact from its receipt. It makes it harder to understand what the BOM is actually keeping an account of. I'd love to see a more thorough discussion of the tradeoffs we are making before we start cementing this as a pattern for all the buildpacks we maintain going forward.

text/go/0001-go-bom.md
sophiewigmore: This is fair Ryan. Part of the decision behind the pattern we used in the Node Module BOM was because while we could've done it the way you've described in point # 1, except for the fact we'd still need a separate buildpack for the Node Start buildpack case. Because of that outlier, we decided to go with the pattern we have now to be consistent across the language family.

I don't think that same concern is relevant here, since the BOM only needs to be created for one order group scenario. Therefore, the plan you've described is viable as far as I know

text/go/0001-go-bom.md
```diff
@@ -0,0 +1,163 @@
...
+   type. This step and step 5 will eventually go away if we support CycloneDX as a BOM format.
+
+4. (Optionally) generate `CPE` and `pURL` from the data and add it to the
+   `postal.Dependency`.
+
+5. Run the
+   [`postal.GenerateBillOfMaterials`](https://github.com/paketo-buildpacks/packit/blob/c5a40518f2c6bd913ade999b9e2d58d6892d2ea9/postal/service.go#L186)
+   command.
+
+6. Clean up the generated BOM file
```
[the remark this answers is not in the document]
robdimsdale: Yeah, to be totally honest we are looking more at the implementation of the buildpack than the RFC, so we should update this RFC and also the other one for consistency.

text/go/0001-go-bom.md
ryanmoran: I think I am much less concerned about the duplication of logic than I am about the decoupling of the BOM from the artifact and also the implicit coupling of this BOM buildpack to the implementations of those buildpacks doing the "work" of the build process.

Additionally, recording the BOM into a single place makes it more awkward for us to support other BOM use-cases going forward. For example, I'd like to see us also record parts of the BOM that are only used during the build phase. This separate buildpack will need to do quite a bit of reverse-engineering to discover how a particular artifact was used in the build process. This is obviously more difficult than just having each buildpack report what it used directly.

text/go/0001-go-bom.md
```diff
@@ -0,0 +1,163 @@
...
+
+#### Potential Future Optimizations
+
+Given future user interest in a way to opt out of this buildpack beyond a custom order grouping, an optimization that
+could be made is the inclusion of a
+`BP_ENABLE_MODULE_BOM` environment variable that can be set during container build-time.
```
fg-j, line 131: Is this intentionally left out of scope for now? A penalty of 10-30 seconds per build seems steep without an off switch.

text/go/0001-go-bom.md
```diff
@@ -0,0 +1,163 @@
...
+
+The addition of this tool to the language family order groupings might have an impactful change to the overall
+performance of the buildpacks.
+
+Performance metrics are still TBD, but for a typical project (e.g. `pack`) we observe that it takes about 10-30 seconds
+to construct a BOM.
+
+We can investigate ways to improve this performance.
+
+### Language Family Additions
```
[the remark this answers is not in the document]
fg-j: @sophiewigmore was the analogous pre-populated `node_modules` case handled in the Nodejs BOM buildpack? What did y'all decide to do?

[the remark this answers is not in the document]
robdimsdale: @fg-j Sorry, I thought your previous comment was asking what the desired behavior is in the case where there is a populated `vendor` directory but no `go.mod` file. In that case `cyclonedx-gomod` fails to execute, so we propose that we do not support that case.

In the case where there is a valid `go.mod` file _and_ a pre-populated `vendor` directory (and vendoring is enabled), then if the `go.mod` file and the `vendor` directory are consistent, this is supported with some limitations. According to the `cyclonedx-gomod` README:

>Modules that use vendoring are, although in a limited manner, supported.
Limitations are as follows:
>
>* **No hashes.** Go doesn't copy all module files to vendor, only those that are required to build and test the main module. Because module checksums consider almost all files in a module's directory though, calculating accurate hashes from the vendor directory is not possible. As a consequence, SBOMs for modules that use vendoring do not include component hashes.
>* **License detection may fail.** Go doesn't always copy license files when vendoring modules, which may cause license detection to fail.

If the `go.mod` file is inconsistent with the `vendor` directory, is this even a valid situation for a go project?

Am I addressing your question, or am I still missing something?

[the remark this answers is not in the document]
robdimsdale: and yes, in the `node` case, the `cyclonedx-node-module` supports the pre-populated vendor directory without a `package.json`. That is essentially a difference between the `cyclonedx-node-module` and `cyclonedx-gomod` tools.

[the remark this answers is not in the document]
fg-j: > If the go.mod file is inconsistent with the vendor directory, is this even a valid situation for a go project?

If you're asking "Will the Go buildpack build it, as it's currently implemented?", I'm pretty sure the answer is yes. If you're asking "Is this a well-formed app?", IMO the answer is no.

ryanmoran (dismissed):

text/go/0001-go-bom.md
```diff
@@ -0,0 +1,274 @@
...
+
+There already exists a tool called [CycloneDX Gomod](https://github.com/CycloneDX/cyclonedx-gomod)
+which is an official CycloneDX-supported tool that creates valid CycloneDX BOMs
+for Go applications. CycloneDX is a BOM format supported by the OWASP
+Foundation, and is a serious contender (at the time of this RFC creation) as a
+format we will eventually support.
+
+The CycloneDX Go Module tool is available as a pre-built binary for multiple
+platforms, and also it can easily be installed via `go install
+github.com/CycloneDX/cyclonedx-gomod`.
```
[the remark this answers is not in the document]
fg-j: Great point. Thanks for the clarification.

ForestEckhardt:
> One last nit to pick 😬 On re-reading this, I got myself confused a few times reading `BP_DISABLE_MODULE_BOM`, which defaults to false.
> 
> For me, more intuitive to have an env var `BP_MODULE_BOM_ENABLED` that defaults to true. This mirrors the naming of `BP_LIVE_RELOAD_ENABLED` and avoids the confusion of a double negative.

If I might counter with it feels a little weird to have unset == `true` because I feel the adding the BOM should be considered the default behavior of the buildpack as opposed to an additional toggle.

ForestEckhardt:
The only other thing to call out is that y'all might want to remove the `Unresolved Questions` section as they both appear to be resolved .

ryanmoran:
I am a proponent of having unset align with false, but @fg-j is right that it appears the Java folks have been making the opposite choice for a while now. The question then is which is better: having a consistent "micro"-interface by aligning unset with false, or having a consistent "macro"-interface by aligning this variable with the choices previously made.

fg-j:
@robdimsdale 

> Assuming we don't want to change the java buildpack (which seems reasonable, as it would result in breaking changes) then I'd advocate for consistency at the macro level across the buildpack ecosystem.

I'm with you here. If this is the chosen direction, the RFC needs to be updated to `BP_MODULE_BOM_ENABLED`. Will approve once that's changed (or explicitly abandoned!)

robdimsdale:
@fg-j

> the RFC needs to be updated to `BP_MODULE_BOM_ENABLED`. Will approve once that's changed (or explicitly abandoned!)

Great, thanks. Did we decide that the unset value for this environment variable defaults to `true`? If so, I'll add a note in the performance section detailing that buildpack consumers can set this variable to `false` to disable BOM generation.

fg-j:
@robdimsdale I think yes. Seems like the behaviour that folks approved in this RFC is BOM-on-by-default (i.e. `BP_MODULE_BOM_ENABLED` defaults true).

text/go/0001-go-bom.md
```diff
@@ -0,0 +1,314 @@
...
+The addition of this tool to the language family order groupings might have an
+impactful change to the overall performance of the buildpacks.
+
+Performance metrics are still TBD, but for a typical project (e.g.
+[`pack`](https://github.com/buildpacks/pack)) we observe that it takes about
+10-30 seconds to construct a BOM.
+
+We can investigate ways to improve this performance. In the meantime, we propose
+including an environment variable: `BP_DISABLE_MODULE_BOM` (which defaults to
+`false`) to skip BOM generation if it takes an unacceptable duration.
```
fg-j, line 130:
```suggestion
including an environment variable: `BP_MODULE_BOM_ENABLED` (which defaults to
`true`) to skip BOM generation if it takes an unacceptable duration.
```

text/go/0001-go-bom.md
```diff
@@ -0,0 +1,314 @@
...
+
+The `go-build` buildpack will optionally `require` this `cyclonedx-gomod`
+dependency, depending on whether it will run BOM generation. Pseudo-code for
+this optional `require` looks as follows:
+
+```
+if BP_DISABLE_MODULE_BOM is unset/false and go.mod file exists:
+    require: go-dist, cyclonedx-gomod
+else:
+    require: go-dist
```
fg-j, line 171:
```suggestion
if BP_MODULE_BOM_ENABLED is false and go.mod file exists:
    require: go-dist, cyclonedx-gomod
else:
    require: go-dist
```

text/go/0001-go-bom.md
```diff
@@ -0,0 +1,314 @@
...
+
+1. Perform the BOM dependency tool retrieval from the dep-server. It will get
+   added to a build-time layer. It will contribute a BOM entry for the tool
+   itself on the `build.toml` `[bom]` section. This will behave similarly to the
+   `dep` buildpack.
+
+The build phase in the `go-build` buildpack will perform the tasks highlighted
+below. If there is no `go.mod` file present, or the `BP_DISABLE_MODULE_BOM`
+environment variable is set to `true`, the `go-build` build phase will skip BOM
+generation entirely.
```
fg-j, line 241:
```suggestion
below. If there is no `go.mod` file present, or the `BP_MODULE_BOM_ENABLED`
environment variable is set to `false`, the `go-build` build phase will skip BOM
generation entirely.
```

text/go/0001-go-bom.md
```diff
@@ -0,0 +1,314 @@
...
+
+1. Parse the `BP_DISABLE_MODULE_BOM` environment variable, skip BOM generation
```
fg-j, line 243:
```suggestion
1. Parse the `BP_MODULE_BOM_ENABLED` environment variable, skip BOM generation
```

fg-j (changes requested):

robdimsdale:
As discussed in today's working group meeting, I'm going to close this in favor of https://github.com/paketo-buildpacks/rfcs/pull/126 as that RFC covers the implementation of SBOM - at least for golang.

We will use the `syft` tool provided by: https://github.com/paketo-buildpacks/rfcs/pull/124 to generate the SBOMs.

robdimsdale:
Actually I can't close this RFC - I'll need @brayanhenao to do that.

brayanhenao:
@robdimsdale Closed

PR #114, opened 2021-10-08
RFC: Auto-generate Reference Documentation

ForestEckhardt:
[Readable](https://github.com/paketo-buildpacks/rfcs/blob/auto-reference-docs/text/0036-auto-reference-docs.md)


text/0036-auto-reference-docs.md
```diff
@@ -0,0 +1,188 @@
...
+...
+```
+
+We propose that all maintainers add this configurations table array to their
+buildpacks as a way of automatically generating information about the
+environment variables that can be configured to modify the buildpacks'
+behavior.
+
+Additional documentation can be generated using an optional file included in
+the project's root named `reference-doc.md`. The `reference-doc.md` file would
```
dmikusa-pivotal, line 104: I feel like this location should be flexible. I might want to keep this under `docs/reference-doc.md` or `docs/reference.md`. I think the proposal should allow for that.

text/0036-auto-reference-docs.md
```diff
@@ -0,0 +1,188 @@
...
+This environment variable is used during launch
+
+...
+
+### BP_JVM_TYPE
+the JVM type - JDK or JRE
+Default Value: `JRE`
+This environment variable is used during build
+
+## Build Plan
```
dmikusa-pivotal, line 130: Where is this section and the next section sourced from in the buildpack.toml?

text/0036-auto-reference-docs.md
```diff
@@ -0,0 +1,188 @@
...
+| X | ✓ | ✓ | `es` |
+| ✓ | X | X | `est` |
+| ✓ | X | ✓ | `est` |
+| ✓ | ✓ | X | `es` |
+| ✓ | ✓ | ✓ | `est` |
+
+---
+You can see the whole sample document generated [here](https://github.com/ForestEckhardt/spikes/blob/main/auto-doc/sample-output.md).
+
+### Alternative To `reference-doc.md`
```
dmikusa-pivotal, line 157: I could see another potential source for additional reference docs being the project's README.md. My main complaint with `reference-doc.md` is that I'll probably have some of the same info here as I will in README.md.

I wonder if there's some approach we could come up with that sources information out of the README.md? 

Perhaps having some front matter / comments in the README.md that wouldn't show up when reading it, but the tool could use to know which sections to extract? or perhaps a hidden config file that would instruct the tool on what sections of README.md it should pull into the docs?

text/0036-auto-reference-docs.md
ForestEckhardt: I am more than happy to allow for alternate location for the `reference-doc.md` I am also not necessarily tied to its currently location.

text/0036-auto-reference-docs.md
```diff
@@ -0,0 +1,188 @@
...
+
+Additional documentation can be generated using an optional file included in
+the project's root named `reference-doc.md`. The `reference-doc.md` file would
+provide maintainers an opportunity to include supplemental documentation. For
+example, maintainers might want to describe the build plan contract or caching
+strategy, etc. We think that having an additional document that is clearly
+named will make the process of adding and updating supplemental reference
+documentation clear and approachable to new contributors.
+
+Below is an example snippet that was generated using a [POC CLI](https://github.com/ForestEckhardt/spikes/tree/main/auto-doc) and a fork of
```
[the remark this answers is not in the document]
ForestEckhardt: Currently it is inserted at the end but I can mark that for you in the example to help you differentiate.

text/0036-auto-reference-docs.md
dmikusa-pivotal: OK. My vote would be for `docs/reference.md`, or something like `.docs.yml` that is like `reference-doc-path: other/reference-path.md`.

text/0036-auto-reference-docs.md
dmikusa-pivotal: First, I think we could come up with some reasonable defaults. For example, without any configuration, the tool would pull in all H2 level sections (and below) as additional reference sections. That would get you a reasonable amount of additional configuration. It might result in some duplicates and we could possibly try to weed those out, or just require configuration to strip them out.

Second, I think we can make the configuration for this pretty simple. Maybe the config file has include / exclude sections where you can list the name of the section to include or exclude. 

```.docs.yml
readme:
  include:
  - "Foo"
  - "Bar Section"
```

or 

```.docs.yml
readme:
  exclude:
  - "Foo"
  - "Bar Section"
```

Where the difference would be that include starts with noting and pulls in just the sections you list and exclude pulls in all sections and excludes what you list.

We don't need to care about the header level, just if the name matches. If it matches, all sub-headers would be included.

You could get more complicated, like with regex or ways to include/exclude sub-headers under an included section, but I'm not sure it needs to be that complicated. At least not to get started with this.

Lastly, I don't think someone contributing would need to care too much about this. If there's a large doc addition to the README, a maintainer would need to review and see what should be synced up with the docs (or possibly not if they are using excludes, then by default new sections would be added). In practice though, I'm not seeing large docs additions to the README. It's small changes as an env option gets added or the detect logic changes slightly. This would allow for the commit to update the README.md & buildpack.toml, which is the standard process now, and ensure the docs are in sync (no other changes necessary). 

I also feel like on a project-by-project basis you could look at automating the update of README sections from buildpack.toml content, which could potentially allow a committer to only need to update buildpack.toml. I think that's another issue though, out of scope here.

text/0036-auto-reference-docs.md
```diff
@@ -0,0 +1,188 @@
...
+buildpack. With the second option, maintainers can update the docs within the
+same commit as the changes.
+
+## Implementation
+
+In order to achieve automation we are proposing the creation of a new CLI tool
+or the extension of an existing CLI tool (i.e. `jam` or `libpak/cmd`). This
+tool will consume the structures in order to generate reference documentation
+in the form of markdown. The first structure we are proposing to consume is the
+`buildpack.toml` with inclusion of the `metadata.configurations` table array.
```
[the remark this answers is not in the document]
dmikusa-pivotal: I feel like that is what I would expect. If a section can't be generated then skip the section. No errors. No empty sections.

text/0036-auto-reference-docs.md
dmikusa-pivotal: I can say for now and the foreseeable future that for the Java-related CNBs our README is basically the reference documentation. If we had a separate doc for reference, I don't think there would be much of anything left in the README. That's what concerns me about duplication. I suppose if we had to have a separate file, we could have an automation to copy README.md to the reference doc page. It's just a duplication for us though.

ryanmoran: One thing that we have in our READMEs that may not make sense for reference docs is some information around how to package up the buildpack to use it if you are building from source. These types of development details are useful to contributors coming into the repo, but not to users just consuming the built artifact.

Relatedly, from what I can tell, the Java buildpacks have never included this type of information in their READMEs. I wonder if that has inadvertently raised the barrier for contribution.

dmikusa-pivotal: @ryanmoran I think the key to making this work would be having some include/exclude capability in the tool, where you could exclude a section with a given header or include only certain section headers. That would let you exclude things that are not relevant for reference docs. You need some level of control to influence what the tool will take from the README and turn into docs.

sophiewigmore: Reading through this following the 10/19 WG meeting, I think I am on board with the README idea. I'd love to see a POC or example of what this would look like for a Java/non Java buildpack. I believe that the issue Ryan has pointed can be solved by Dan's include/exclude structure (however that works in practice).  As a maintainer of buildpacks, I feel like a lot of the docs are written/updated by myself or other maintainers, so my main concern is more around information de-duplication rather than ease of contribution in this case. That being said, even with the README case, I believe that the ease of contribution is still greatly improved just by having the documentation live alongside the buildpacks instead of the website doc.

dmikusa-pivotal: That looks good 👍. It encapsulates the idea of what I was thinking.

sophiewigmore: Just tried it out also, this works as I'd expected as well! Awesome @ForestEckhardt

ryanmoran (approved):

sophiewigmore (approved):

dmikusa-pivotal (approved):

TisVictress (approved):

Issue #95, opened 2021-07-16
Implement RFC0030: Buildpackless Builders


fg-j:
[RFC](https://github.com/paketo-buildpacks/rfcs/blob/6bf4ef13f1ad0420dcd0ace37a81680ae8e05a6a/text/0030-buildpackless-builders.md)

### Summary
Paketo should publish buildpackless versions of its three builders.

* [ ] All three builders (Full, Base, and Tiny) have buildpackless versions published on Dockerhub and GCR that automatically receive updates and cut versions when new lifecycle and stack images are available.

fg-j:
This has been implemented for the full builder with one repo and 2 orphaned mainline branches: https://github.com/paketo-buildpacks/full-builder. (See `main` and `buildpackless`. 

However, in the process of implementing with this approach, it's evident that publishing two builders in one repository adds complexity that might not be worth the benefits of keeping them together.

Complexity includes:
- automatically cutting releases that target 2 different branches
- PR-ing into two different branches when `builder.toml` files need to be updated
- a more complicated `dependabot.yml` to keep two different mainline branches' dependencies up to date
- syncing workflows from a centralized github-config repo is more complicated if two mainline branches need to be updated with slightly different sets of files

For these reasons, it might be worth revisiting the one-repo approach to this RFC's implementation. Checking in the buildpackless builders in their own repos can significantly reduce the automation complexity.

PR #124, opened 2021-11-03
Proposes to add a new utilities buildpack for Syft

dmikusa-pivotal:
## Summary

A proposal to add a utility buildpack that will install the [Syft](https://github.com/anchore/syft) CLI.

## Use Cases

This tool can be used to generate SBoM material in standard formats like SPDX & CycloneDX.

## Checklist


text/utilities/0000-syft-buildpack.md
```diff
@@ -0,0 +1,45 @@
...
+
+* Another buildpack requires `syft`
+
+When conditions are met, the buildpack will do the following:
+
+* Contributes Syft to a layer marked `build` and `cache` with command on `$PATH`
+
+Buildpacks that wish to opt-in, simply need to indiate that they require `syft` at detection time. Then `syft` should be present on the path at build time and these buildpacks can run it.
+
+The Syft buildpack would need to be added to the order group for any language family that wishes to use `syft`. It needs to be added prior to the buildpack that is going to require `syft`, which would likely require it to be very early in the order group. For example, the maven buildpack will require Syft, so the Syft buildpack needs to be listed in the order groups prior to maven.
```
robdimsdale, line 37: Based on our explorations with SBOM for Node JS and Golang, we wanted the BOM to be generated at build-time, which was essentially at the end of the buildpack process, not the "fairly early on".

I don't think that changes any of the rationale, just pointing it out.

text/utilities/0000-syft-buildpack.md
```diff
@@ -0,0 +1,45 @@
...
+
+The Syft buildpack would be maintained by the existing [Utilities team](https://github.com/orgs/paketo-buildpacks/teams/utilities). It uses libpak & libcnb.
+
+## Motivation
+
+It can be used by buildpacks to scan and generate both CycloneDX and SPDX listings of layers for compliance with [Buildpacks RFC #95](https://github.com/buildpacks/rfcs/blob/main/text/0095-sbom.md).
+
+## Detailed Explanation
+
+1. A Syft buildpack has been developed. It will be contributed to Paketo.
```
robdimsdale, line 15: Out of curiosity, have you already created the buildpack? Can we try it out?

text/utilities/0000-syft-buildpack.md
```diff
@@ -0,0 +1,45 @@
+# Syft Buildpack
+
+## Summary
+
+Introduce a Syft buildpack. This buildpack will provide the `syft` binary should any buildpacks want to use it. The `syft` CLI is a tool and library for generating a Software Bill of Materials from container images and filesystems.
```
robdimsdale, line 5: I'd love to add a link to syft (https://github.com/anchore/syft) in this section.

text/utilities/0000-syft-buildpack.md
```diff
@@ -0,0 +1,45 @@
...
+
+The new Syft buildpack will participate all the following conditions are met
+
+* Another buildpack requires `syft`
+
+When conditions are met, the buildpack will do the following:
+
+* Contributes Syft to a layer marked `build` and `cache` with command on `$PATH`
+
+Buildpacks that wish to opt-in, simply need to indiate that they require `syft` at detection time. Then `syft` should be present on the path at build time and these buildpacks can run it.
```
robdimsdale, line 35: Would you anticipate also adding in a version to the `provides/requires` relationship? It looks like syft is pre 1.0.0 so i could imagine there will be breaking changes coming at some point.

text/utilities/0000-syft-buildpack.md
dmikusa-pivotal:
```suggestion
Introduce a Syft buildpack. This buildpack will provide the [`syft` binary](https://github.com/anchore/syft) should any buildpacks want to use it. The `syft` CLI is a tool and library for generating a Software Bill of Materials from container images and filesystems.
```

dmikusa-pivotal: How does that look?

text/utilities/0000-syft-buildpack.md
dmikusa-pivotal: yes & yes. I'll DM you about it.

text/utilities/0000-syft-buildpack.md
robdimsdale: perfect!

text/utilities/0000-syft-buildpack.md
dmikusa-pivotal: That's certainly possible. Both breaking changes and having a way to version this. We typically handle this with buildpack by locking the buildpack version, since buildpacks are tightly coupled to the dependency versions they install.

In short, if something breaks we either don't update the Syft buildpack immediately, or we lock the consuming buildpacks to use the last known working version.

We will follow the Paketo RFC on Semver, so if things change upstream and that breaks buildpacks, I'd expect that to occur across major versions only. 

If you feel strongly about it we could explore adding support for multiple major version branches, so like we'd have 0.x and 1.x (when that's released). I could see other buildpacks passing the version through metadata. We're not doing that with other buildpacks like UPX or Watchexec though, so I would rather cross that bridge when we get to it, if it ends up being necessary.

text/utilities/0000-syft-buildpack.md
dmikusa-pivotal: I had originally hoped to have this Syft buildpack run itself. Basically, add it at the end of the change & have it run `syft` once to collect all the metadata.

It doesn't seem feasible though. RFC #95 requires individual buildpacks to provide metadata in buildpack.toml that says the types of SBoM they export. If they rely on another buildpack to export that metadata, it gets a little weird. Stephen also mentioned to me that there could be some visibility issues, where the single buildpack at the end might not be able to see everything that was done.

I'm not 100% sure I understand Stephen's comment, but it did make me think of how we delete application source files as part of our build process, so again, it wouldn't work to scan at the end because the application source is gone and that could cause the scanner to miss some things.

That has essentially left us adding a Syft buildpack to install the CLI then consuming the CLI as necessary in individual buildpacks. That's why I mentioned including it early, because that's the only way to make sure Syft is installed before the consuming buildpacks need it.

Anyway, I'll end my long rambling comment. I'm not sure how much of that is relevant, just figured I'd share a little more of our plans on the Java side.

dmikusa-pivotal:
```suggestion
The Syft buildpack would need to be added to the order group for any language family that wishes to use `syft`. It needs to be added prior to the buildpack that is going to require `syft`, which would likely require it to be very early in the order group. For example, the maven buildpack will require Syft, so the Syft buildpack needs to be listed in the order groups prior to maven. If it's included and the exact order where it's included can be decided by each buildpack language family based on individual need.
```

I added this to just reinforce that it's the buildpack language family's choice if and where to include this.

text/utilities/0000-syft-buildpack.md
robdimsdale: Oh, gotcha. So we would automatically bump the version of this buildpack when there's a new version of syft, and if it happens to break then language-family buildpacks (e.g. Java) lock to the last-known good version of this buildpack, rather than the specific downstream buildpacks (e.g. Maven) locking to a version of the dependency in the `provides`/`requires` relationship?

I'm definitely not looking to add more work and process for the sake of it!

dmikusa-pivotal: Yes, exactly.

sophiewigmore:
@dmikusa-pivotal I haven't been following along with the latest iterations of the BOM work. What was the rationale for adding this buildpack, rather than following creating the BOM alongside buildpacks like we did in the Node.js language family, and now in Go? Is it mostly because of the format versatility (support for SPDX too)? Would this replace the need for BOM creation solutions like the ones I outlined?

dmikusa-pivotal:
> @dmikusa-pivotal I haven't been following along with the latest iterations of the BOM work. What was the rationale for adding this buildpack, rather than following creating the BOM alongside buildpacks like we did in the Node.js language family, and now in Go? Is it mostly because of the format versatility (support for SPDX too)? Would this replace the need for BOM creation solutions like the ones I outlined?

The main rationale for this RFC is to make this tool an option for SBoM management. As you said, there are some benefits of using the same tool across multiple buildpacks. For the Java team, we have several build tools to support, Maven and Gradle being the largest two but there's also SBT, Lein, and Clojure Tools, so rather than integrate different plugins for each tool we can use Syft across them all. It'll give us consistent output across the tools. Plus, as you mentioned it supports a variety of output formats so we don't need different plugins for CycloneDX and SPDX or conversion tools. 

That said, this RFC is not in any way trying to dictate that language families adopt this particular tool. It's just to make it available and something we officially maintain. Language teams can choose to adopt or not adopt it on their own.

sophiewigmore:
@dmikusa-pivotal understood, thanks for outlining that.

samj1912:
I am a bit torn between using syft as a binary vs as a library. Using it as a binary means that each buildpack doesn't need to include the entire syft dependency chain but on the other hand we need some sort of bin wrapper / library. Regardless +1 on the decision to go with syft. It's probably the fastest sbom generation tool that I have come across.

If we do decide to go the syft binary route, it would be really nice to have a reusable go lib to invoke syft bom generation commands instead of relying on pexec/effect directly.

dmikusa-pivotal:
> I am a bit torn between using syft as a binary vs as a library. Using it as a binary means that each buildpack doesn't need to include the entire syft dependency chain but on the other hand we need some sort of bin wrapper / library. 

Yes, that's tricky. There are definitely advantages both ways. I think that since this RFC is primarily for making an official buildpack that would provide `syft` as a CLI, this unlocks the ability for folks to easily consume the CLI while still reserving the right for individual language teams to not go that route. If they would prefer using a library instead, or even some other tool.

That said, I think the CLI does provide some advantages. It's pretty well encapsulated, including things like [reading from standard configuration locations](https://github.com/anchore/syft#configuration) out-of-the-box. We also don't need a lot of customization with the tool, what it does with the standard commands is exactly what we need. I think if we had to customize the operation of it more, then perhaps going the library route could make some sense. Like previously mentioned, this RFC doesn't lock us into anything, we could still pivot and use the library if down the road we end up needing more customization. The nice thing is that CLI or library are both going to give a consistent experience.

Anyway, that is why I'm leaning towards CLI at this point. I'm sure someone could make a compelling argument for the library route too. Fortunately, we don't have to pick exclusively one or the other.
 
> If we do decide to go the syft binary route, it would be really nice to have a reusable go lib to invoke syft bom generation commands instead of relying on pexec/effect directly.

In my initial trials, I have been using libbs for this. It's basically one efffect.Executor call to run the binary and copy the output to a file, and it'll update all of the Paketo "build system" buildpacks (Maven, Gradle, etc..) to have this support. This will essentially give you a SBoM for the application and its dependencies.

I could see moving this into libpak so that it's easy to call from other buildpacks as well. I haven't thought it all through, but we might end up calling that from other CNBs too (like maybe executable-jar or Spring Boot CNB).

It's outside of this RFC, but I'll make sure that happens.

ForestEckhardt:
> I could see moving this into libpak so that it's easy to call from other buildpacks as well. I haven't thought it all through, but we might end up calling that from other CNBs too (like maybe executable-jar or Spring Boot CNB).

Is there a reason why you would want to put this in libpak as opposed to putting it into the buildpack itself. If we would like to make the functionality universal for at least buildpacks written in Go it would make sense to me to attache the functionality to the buildpack as opposed to a implementation library as that would then force you to import two implementation libraries if you are using something that is not libpak or alternatively it would force you to copy and paste code into your implementation.

It would also make sense from a functionality upgrade standpoint (maybe adding common helpers and what not) to have it versioned with the buildpack and its release.

With that being said if we did want a shared interaction library I would argue that it should maybe be at least mentioned in the RFC as future work. I don't think that details need to be completely fleshed out but some sort of mention makes sense in my head.

dmikusa-pivotal:
> > I could see moving this into libpak so that it's easy to call from other buildpacks as well. I haven't thought it all through, but we might end up calling that from other CNBs too (like maybe executable-jar or Spring Boot CNB).
> 
> Is there a reason why you would want to put this in libpak as opposed to putting it into the buildpack itself. If we would like to make the functionality universal for at least buildpacks written in Go it would make sense to me to attache the functionality to the buildpack as opposed to a implementation library as that would then force you to import two implementation libraries if you are using something that is not libpak or alternatively it would force you to copy and paste code into your implementation.
> 
> It would also make sense from a functionality upgrade standpoint (maybe adding common helpers and what not) to have it versioned with the buildpack and its release.
> 
> With that being said if we did want a shared interaction library I would argue that it should maybe be at least mentioned in the RFC as future work. I don't think that details need to be completely fleshed out but some sort of mention makes sense in my head.

@ForestEckhardt I only mentioned libpak because I believe @samj1912 is using it. I wouldn't intend for anyone to use syft through libpak unless they are already using libpak. IMHO, there wouldn't be an advantage to using syft this way unless you're already using libpak.

ryanmoran (approved):

dmikusa-pivotal:
As mentioned in the Paketo WG meeting today, we need to get some additional approvals on this.

@paketo-buildpacks/utilities-maintainers If you could provide your 👍/👎 that would be much appreciated. Thanks

PR #126, opened 2021-11-08
Buildpack Deps SBoM RFC

thitch97:
## Summary
[Readable](https://github.com/paketo-buildpacks/rfcs/blob/sbom-bps/text/0037-cdx-syft-bom.md)

- [x] I have viewed, signed, and submitted the Contributor License Agreement.
- [ ] I have linked issue(s) that this PR should close using keywords or the Github UI (See docs)
- [ ] I have added an integration test, if necessary.
- [x] I have reviewed the styleguide for guidance on my code quality.
- [x] I'm happy with the commit history on this PR (I have rebased/squashed as needed).

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@
 * [0035: Python Paketo Buildpack Promotion](./text/0035-python-promotion.md)
 * [0036: Explorations Repository](./text/0036-explorations.md)
 * [0037: Remote Debug](./text/0037-remote-debug.md)
+* [0038: Support for CycloneDX and Syft SBoM](./text/0038-cdx-syft-sbom.md)
 
 ## Implemented RFCs
 
--- a/text/0038-cdx-syft-sbom.md
+++ b/text/0038-cdx-syft-sbom.md
@@ -1,0 +1,128 @@
+# Support for CycloneDX and Syft SBoM in Paketo
+
+## Summary
+
+Buildpacks which directly provide a dependency or buildpacks which install
+application dependencies will support the generation of SBoM in the format of
+CycloneDX and Syft. Per [CNB
+RFC#95](https://github.com/buildpacks/rfcs/blob/main/text/0095-sbom.md), the
+SBoM documents will live at `<layer>.bom.<ext>.json`, `launch.bom.<ext>.json`
+and `build.bom.<ext>.json` where `<ext>` will be `cdx` (CycloneDX) or
+`syft` (Syft).
+
+## Motivation
+
+[Paketo
+RFC#33](https://github.com/paketo-buildpacks/rfcs/blob/main/text/0033-bill-of-materials.md)
+describes the initial strategy used to add support for SBoM generation in
+Paketo buildpacks. This RFC outlines the process by which Paketo buildpacks
+will build upon that initial strategy by including support for CycloneDX and
+Syft SBoM formats in accordance with the structure proposed in [CNB
+RFC#95:SBOM](https://github.com/buildpacks/rfcs/blob/main/text/0095-sbom.md).
+
+## Detailed Explanation
+
+Paketo buildpacks may produce SBoM for runtime & application dependencies in
+CycloneDX and Syft formats. SBoM information may be provided in three
+locations, namely `<layers>/<layer>.bom.<ext>.json`, `<layers>/launch.bom.<ext>.json` and
+`<layers>/build.bom.<ext>.json`, where <ext> is `cdx` for CycloneDX or `syft` for Syft
+formats. The preferred content of these files is outlined below:
+
+- `<layer>.bom.<ext>.json`: This file should be used only for SBoM information
+  that is specific to a layer. This allows for greater specificity on the
+  origin of the SBoM metadata, including which layer it relates to and which
+  buildpack created the layer. Storing metadata here also potentially reduces
+  overhead incurred by SBoM generation as the file would follow the lifecycle
+  of the layer itself in terms caching and reuse.
+
+- `launch.bom.<ext>.json`: This file should contain SBoM entries for
+  dependencies which are contributed to the app image, but are not necessarily
+  scoped to a specific layer. SBoM information stored here is ephemeral and
+  would need to be regenerated on each build.
+
+- `build.bom.<ext>.json`: This file should contain SBoM entries for
+  dependencies which are contributed to the build environment, but are not
+  necessarily scoped to a specific layer. SBoM information stored here is
+  ephemeral and would need to be regenerated on each build.
+
+**Note:** The above schema is not a hard requirement. What each of these files
+consists of is ultimately the buildpack author's decision. Additional context
+can be found
+[here](https://github.com/buildpacks/rfcs/blob/main/text/0095-sbom.md#what-it-is).
+
+
+Buildpack authors will also need to specify which SBoM formats are supported by
+a given buildpack via `buildpack.toml`. This should look like:
+
+```toml
+api = "0.x"
+
+[buildpack]
+id = "<buildpack ID>"
+name = "<buildpack name>"
+# This can be an array of supported SBOM formats by the buildpack.
+# Valid array values are sbom media types based on https://www.iana.org/assignments/media-types/media-types.xhtml
+sbom = ["application/vnd.cyclonedx+json", "application/vnd.syft+json"]
+```
+
+### SBoM Formats
+
+#### Syft
+
+The minimum set of Syft fields that will be initially supported is represented
+below. It is possible that additional fields will be supported in the future.
+
+```json
+{
+ "artifacts": [
+  {
+   "id": <Syft-generated UUID string for in-file dependency graphing>,
+   "name": <dependency name>,
+   "version": <dependency version>,
+   "licenses": [<dependency license ID(s) in SPDX-format>],
+   "cpes": [<version-specific common platform enumerations>],
+   "purl": [<package URL per github.com/package-url>]
+  }
+ ],
+ "descriptor": {
+  "name": "syft",
+  "version": <Syft version>
+ },
+ "schema": {
+  "version": "1.1.0",
+  "url": "https://raw.githubusercontent.com/anchore/syft/main/schema/json/schema-1.1.0.json"
+ }
+}
+```
+
+#### CycloneDX
+
+The minimum set of CycloneDX fields that will be initially supported is represented
+below. It is possible that additional fields will be supported in the future.
+
+```json
+{
+  "bomFormat": "CycloneDX",
+  "specVersion": "1.3",
+  "version": <BOM revision number>,
+  "components": [
+    {
+      "type": <component type, usually library>,
+      "name": <package name>,
+      "version": <package version>,
+      "licenses": [<package license ID(s) in SPDX-format>],
+      "purl": <package URL per github.com/package-url>,
+    }
+  ]
+}
+```
+
+
+## Prior Art
+
+[Paketo RFC#33](https://github.com/paketo-buildpacks/rfcs/blob/main/text/0033-bill-of-materials.md)
+
+[CNB RFC #95: SBOM](https://github.com/buildpacks/rfcs/blob/main/text/0095-sbom.md)
+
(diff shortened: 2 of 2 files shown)


text/0037-syft-bom.md
```diff
@@ -0,0 +1,103 @@
+# Support for CycloneDX and Syft SBoMs in Paketo
+
+## Summary
+
+Buildpacks which directly provide a dependency or buildpacks which install
+application dependencies will support the generation of SBoM in the format of
+CycloneDX or Syft. Per [CNB
```
ryanmoran, line 7:
```suggestion
CycloneDX and Syft. Per [CNB
```

text/0037-syft-bom.md
```diff
@@ -0,0 +1,103 @@
...
+describes the initial strategy used to add support for SBoM generation in
+Paketo buildpacks. This RFC outlines the process by which Paketo buildpacks
+will build upon that initial strategy by including support for CycloneDX and
+Syft SBoM formats in accordance with the structure proposed in [CNB
+RFC#95:SBOM](https://github.com/buildpacks/rfcs/blob/main/text/0095-sbom.md).
+
+## Detailed Explanation
+
+Paketo buildpacks will now optionally produce SBoM for runtime &
+application dependencies in CycloneDX or Syft formats, or both. Buildpack
```
ryanmoran, line 26:
```suggestion
application dependencies in CycloneDX and Syft formats. Buildpack
```

text/0037-syft-bom.md
```diff
@@ -0,0 +1,103 @@
...
+
+```toml
+api = "0.x"
+
+[buildpack]
+id = "<buildpack ID>"
+name = "<buildpack name>"
+# This can be an array of supported SBOM formats by the buildpack.
+# Valid array values are sbom media types based on https://www.iana.org/assignments/media-types/media-types.xhtml
+sbom = ["application/vnd.cyclonedx+json", "text/spdx+json"]
```
ryanmoran, line 38:
```suggestion
sbom = ["application/vnd.cyclonedx+json", "application/vnd.syft+json"]
```

ryanmoran (changes requested):

text/0037-syft-bom.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Summary
+
+Buildpacks which directly provide a dependency or buildpacks which install
+application dependencies will support the generation of SBoM in the format of
+CycloneDX and Syft. Per [CNB
+RFC#95](https://github.com/buildpacks/rfcs/blob/main/text/0095-sbom.md), the
+SBoM documents will live at `<layer>.bom.<ext>.json`, `launch.bom.<ext>.json`
+and `build.bom.<ext>.json` where `<ext>` will be `cdx.json` (CycloneDX) or
+`syft.json` (Syft).
```
dmikusa-pivotal, line 11:
```suggestion
and `build.bom.<ext>.json` where `<ext>` will be `cdx` (CycloneDX) or
`syft` (Syft).
```

Minor change. The `<ext>` part shouldn't have `.json` which is already part format listed.

text/0037-syft-bom.md
```diff
@@ -0,0 +1,103 @@
...
+describes the initial strategy used to add support for SBoM generation in
+Paketo buildpacks. This RFC outlines the process by which Paketo buildpacks
+will build upon that initial strategy by including support for CycloneDX and
+Syft SBoM formats in accordance with the structure proposed in [CNB
+RFC#95:SBOM](https://github.com/buildpacks/rfcs/blob/main/text/0095-sbom.md).
+
+## Detailed Explanation
+
+Paketo buildpacks will now optionally produce SBoM for runtime &
+application dependencies in CycloneDX and Syft formats. Buildpack
```
dmikusa-pivotal, line 26: There are three different locations where a buildpack can put SBoM information. I don't feel like buildpacks RFC #95, provides any direction as to what should go in each of these files. 

Do we want to specify where the buildpacks should put certain types of information? To provide some consistency across buildpacks.

For example, SBoM information should be broken down like this:

- `<layer>.bom.<ext>.json` This file contains SBoM information specific to a layer. A Buildpack that installs dependencies will install them into a specific layer and should include the SBoM entry for the installed dependency into this layer-specific SBoM file.

- `launch.bom.<ext>.json` This file should contain SBoM information for dependencies not installed into a specific layer, such as application-specific dependencies that have been installed, and those dependencies will exist in the launch image.

- `build.bom.<ext>.json` This file should contain SBoM information for dependencies not installed into a specific layer, such as application-specific dependencies that have been installed, and those dependencies are only used at build time (i.e. they do not end up in the launch image).

I'm not sure if my example above is correct. I'm interested to see how others interpreted the RFC and what plans are for utilizing the different files. 

Regardless of how it's interpreted, it does seem worthwhile to me to provide some guidance, not requirements, on how these files should be used to promote consistency across buildpacks.

thitch97: That seems reasonable to me, I'll update this section

dmikusa-pivotal:
Should we include a provision to make CycloneDX optional? It seems like the primary focus is on Syft JSON, that should be required for all builds. Then, to reduce work/performance impact of scanning, we could have CycloneDX available but only if a certain env variable is toggled on. Maybe `BP_SBOM_CYCLONEDX_ENABLED`?

text/0037-syft-bom.md
```diff
@@ -0,0 +1,103 @@
...
+ "schema": {
+  "version": "1.1.0",
+  "url": "https://raw.githubusercontent.com/anchore/syft/main/schema/json/schema-1.1.0.json"
+ }
+}
+```
+
+#### CycloneDX
+
+The minimum set of Syft fields that will be initially supported is represented
```
dmikusa-pivotal, line 73:
```suggestion
The minimum set of CycloneDX fields that will be initially supported is represented
```

dmikusa-pivotal:
> Should we include a provision to make CycloneDX optional? It seems like the primary focus is on Syft JSON, that should be required for all builds. Then, to reduce work/performance impact of scanning, we could have CycloneDX available but only if a certain env variable is toggled on. Maybe `BP_SBOM_CYCLONEDX_ENABLED`?

Rescinding this comment. 

Based on feedback in the WG meeting today, we'll keep an eye on the impact to build times and upstream tooling with regards to converting formats. If there's a noticeable impact on performance and if upstream tools do not add functionality to allow efficient format conversion or multiple format generation, then we can revisit flags to disable.

text/0037-syft-bom.md
thitch97: This makes sense to me but I wonder whether we can be less strict in the guidance for `launch.bom` and `build.bom`, specifically the bit about the SBoM information being "for dependencies not installed to a layer". In the `node-module-bom` buildpack, for example, CycloneDX [is installed into a specific layer](https://github.com/paketo-buildpacks/node-module-bom/blob/cd06b3d2bee74891acc4135c0fa07c883f177103/build.go#L62) at build time, but its BoM is then appended to the general Build BoM as opposed to a layer-specific BoM.

As these are guidelines, would it be reasonable to you to omit that stipulation and state simply that launch.bom is for deps which end up in the app image and  build.bom is for deps contributed to the build environment?

dmikusa-pivotal: My initial comment was really just my initial interpretation of read through the spec and RFC #95. I'm absolutely not asserting that is correct. It's a hazy topic for me.

I'm totally fine with only using `launch.bom` and `build.bom`. It seems like everyone is on board with that and I think that distinction is clear and pretty easy. The spec also clearly defines this (not updated for RFC#95 but the spirit is the same):  

>SHOULD write BOM (Bill-of-Materials) entries to <layers>/launch.toml describing any contributions to the app image.
>SHOULD write build BOM entries to <layers>/build.toml describing any contributions to the build environment.

Where I get kind of confused is with the third file: `<layer>.bom.<ext>.json`. Given that build things go in `build.bom` and launch things go in `launch.bom`, what's left to go in `<layer>.bom`? Maybe nothing is the answer?

dmikusa-pivotal: OK. I asked for some clarifications in the buildpacks slack. There is some designed overlap in these files. The key benefits to using`<layer>.bom.<ext>.json` are this:

1. The SBoM information in this file is specifically attached to a particular layer, so it's more specific. It's not just saying, this software is in the generated image. It's saying, it's in the generated image and it was put into a specific layer by a specific buildpack, which created the layer.
2. If you put metadata into `<layer>.bom.<ext>.json` then it can be cached. This could potentially reduce the overhead of generating the SBoM metadata.

Given that, it sounds like we should prefer using `<layer>.bom.<ext>.json` for dependencies installed by buildpacks as that would allow the SBoM metadata to be cached with the layer. I think we should prefer or suggest, but not require that. That provides guidance but ultimately leaves it up to the buildpack author.

That would leave `build.bom` and `launch.bom` for the things that don't fit cleanly into layers, like possibly application dependencies or cases where it's not easy to differentiate between tooling and dependencies.

Does that seem reasonable?

fg-j:
@robdimsdale and I have been investigating how this might work for the Go language family buildpack. One question has arisen for us: where on the container filesystem do the `<layer>.bom.<ext>.json`, `launch.bom.<ext>.json` and `build.bom.<ext>.json` live? The upstream RFC (as of when I'm writing this) [states](https://github.com/buildpacks/rfcs/blame/main/text/0095-sbom.md#L103):

> The lifecycle binaries will put the bom files from layers set to launch = true inside the
> `/layers/config/sbom/launch/<buildpack-id>/<layer-id>/bom.<ext>` hierarchy otherwise it will put them in
> `/layers/config/sbom/build/<buildpack-id>/<layer-id>/bom.<ext>`.

Seems to me that it's the responsibility of the lifecycle to create and manipulate those files. Given that, what files should buildpacks produce, and how does the lifecycle find out about them?

text/0037-syft-bom.md
samj1912: Just some background, rfc 95 superceded rfc 87 which introduced this concept. For more information see this section https://github.com/buildpacks/rfcs/blob/main/text/0087-bom-in-layer-metadata.md#what-it-is

thitch97:
> @robdimsdale and I have been investigating how this might work for the Go language family buildpack. One question has arisen for us: where on the container filesystem do the `<layer>.bom.<ext>.json`, `launch.bom.<ext>.json` and `build.bom.<ext>.json` live? The upstream RFC (as of when I'm writing this) [states](https://github.com/buildpacks/rfcs/blame/main/text/0095-sbom.md#L103):
> 
> > The lifecycle binaries will put the bom files from layers set to launch = true inside the
> > `/layers/config/sbom/launch/<buildpack-id>/<layer-id>/bom.<ext>` hierarchy otherwise it will put them in
> > `/layers/config/sbom/build/<buildpack-id>/<layer-id>/bom.<ext>`.
> 
> Seems to me that it's the responsibility of the lifecycle to create and manipulate those files. Given that, what files should buildpacks produce, and how does the lifecycle find out about them?

My interpretation was that each of those `bom` files would live alongside their equivalent metadata files (i.e. `build.toml`, `launch.toml` & `<layer>.toml`) and that the lifecycle would use the information from the metadata files to place `bom` files in the correct location according to the quote you cited. @samj1912 is that assumption correct?

text/0037-syft-bom.md
thitch97: @dmikusa-pivotal That does seem reasonable to me, I'll modify this section.

thitch97: @dmikusa-pivotal Updated the section with additional guidance

dmikusa-pivotal (approved):

fg-j:
@thitch97 That makes sense. Also seems to line up with the behaviour of the function in the lifecycle's `builder` that will move buildpacks' BOMs into the aggregated location (see [this part of the PR implementing this lifecycle feature](https://github.com/buildpacks/lifecycle/blob/5537f2904853ca314775acd5eb70b0127f457085/builder.go#L130))

ryanmoran (approved):

PR #129, opened 2021-11-15
RFC: Source Removal Buildpack

ForestEckhardt:
[Readable](https://github.com/paketo-buildpacks/rfcs/blob/source-removal-buildpack/text/utilities/0004-source-removal-buildpack.md)


text/0039-source-removal-buildpack.md
```diff
@@ -0,0 +1,61 @@
...
+
+
+## Prior Art
+
+- [Paketo Go Build Buildpack](https://github.com/paketo-buildpacks/go-build/blob/main/source_deleter.go)
+- [Google Clear Source](https://github.com/GoogleCloudPlatform/buildpacks/blob/main/pkg/clearsource/clearsource.go)
+
+## Unresolved Questions and Bikeshedding
+
+- Should we promote directly into the `paketo-buildpacks` organization?
```
dmikusa-pivotal, line 57: My vote would be yes if we feel that the buildpack is complete and the implementation is stable. i.e. is it a 1.0.0? It saves time and energy to do it this way.

If we think it still needs some work or is missing features we'd like it to have the community is probably better while we work out the remaining details.

My $0.02.

text/0039-source-removal-buildpack.md
```diff
@@ -0,0 +1,61 @@
+# Source Removal Buildpack
+
+## Summary
+
+Adopt [ForestEckhardt's Source Removal buildpack](https://github.com/ForestEckhardt/source-removal)
+into the Paketo Project to allow for uniform source removal logic across all
+buildpacks that require such functionality.
```
dmikusa-pivotal, line 7: I like the concept & agree that implementing the logic in a consistent way across all buildpacks that require this in Paketo would be helpful.

What's challenging is that the way this is presently implemented for Java-related buildpack (and Rust), is that the delete has to happen at a specific time because there is work that happens after the deletion.

At a high level:

- Do a build, the artifact is stored in a layer so the build is cached
- We delete source code
- Artifact is taken from layer and moved to `/workspace`

I don't think we could change this for the Java buildpacks because buildpacks that run after our build tools, require the build artifact to exist in `/workspace`. This is because a user can push either code or a build artifact. If a user pushes a build artifact, it'll be in `/workspace`. Thus to build from source, we need our build tools buildpacks to compile and place the build artifact in the same location.

I could possibly see this buildpack working for the Rust buildpacks because they don't have the same need as Java.

How would you feel about exposing this buildpack as both a buildpack and as a consumable library? If the actual code for performing the delete were wrapped up in a module, buildpacks like Java could consume the module and integrate it precisely where needed.

dmikusa-pivotal: I suppose another possibility would be to have this RFC document the interface and env variables that should be exposed & to contribute the buildpack as an implementation of this interface. That would free buildpacks like Java to implement it in their own way, but follow the prescribed interface so the user experience is the same.

text/0039-source-removal-buildpack.md
```diff
@@ -0,0 +1,61 @@
...
+Currently ForestEckhardt's Source Removal buildpack always passes detection and
+has three actions.
+
+1. If no configuration is set (i.e. `BP_INCLUDE_FILES` and `BP_EXCLUDE_FILES`
+   are not set), then all the content of the `workspace` will be deleted.
+2. If `BP_INCLUDE_FILES` is set, then all files that **do not** match the
+   pattern globs specified in the environment variables will be deleted.
+3. If `BP_EXCLUDE_FILES` is set, then all files that **do** match the pattern
+   globs specified in the environment variables will be deleted.
+
```
dmikusa-pivotal, line 46: A couple of items that might be interesting for inclusion:

1. Support both `BP_INCLUDE_FILES` and `BP_EXCLUDE_FILES`. I have seen this with some tools, and I think it typically works by applying them in order, so you only include what matches `BP_INCLUDE_FILES` and from that subset of files, you apply the excludes filter.

    It's probably not as common of a use case, but I could potentially see where you might have a folder like `static/` where you only want to include that but then perhaps you want to exclude some content from under that? Maybe exclude hidden files?

2. Regex support. I know, I know. Regex just adds problems, lol, but it is a powerful way to match and customize filtering.

Perhaps #2 is too specific a detail. In which case, we perhaps reword the descriptions here such that they don't prescribe a specific matching implementation like glob or regex, but leave that up to the implementation.

text/0039-source-removal-buildpack.md
```diff
@@ -0,0 +1,61 @@
...
+## Implementation
+
+Currently ForestEckhardt's Source Removal buildpack always passes detection and
+has three actions.
+
+1. If no configuration is set (i.e. `BP_INCLUDE_FILES` and `BP_EXCLUDE_FILES`
+   are not set), then all the content of the `workspace` will be deleted.
+2. If `BP_INCLUDE_FILES` is set, then all files that **do not** match the
+   pattern globs specified in the environment variables will be deleted.
```
dmikusa-pivotal, line 43: Possibly a note that is more for the buildpack docs, but I think we should define and provide some examples of what is meant by "globs". Perhaps some examples of how you can include/exclude various files using globs.

An alternative would be to link to a glob library or shell docs that explain how this would work. If we can say, we use library X or we implement things like Shell Y, then users will have a more concrete idea of what is possible.

text/0039-source-removal-buildpack.md
ForestEckhardt: 1. I think that is not a complete unreasonable request but I would like to see it for a user before complicating the logic more than it already is. I personally feel that this is something that could be very easily added after the buildpack has been adopted and shouldn't be a blocker.
2. I am reluctant to allow for users to user regex primarily because they are not really the convention used by other file control systems. I think I would have to hear a compelling reason to use regex over relatively standard file glob syntax.

text/0039-source-removal-buildpack.md
ForestEckhardt: I am more than happy to reveal that information if it should be necessary. Currently I am just using [`filepath.Match`](https://pkg.go.dev/path/filepath#Match) which is part of the Go standard library.

text/0039-source-removal-buildpack.md
ForestEckhardt: I think the real issue you have highlighted is that having some way for the buildpacks to communicate to the source removal buildpack would be very beneficial. 

I think where I run into problems with that idea is that this buildpack also accepts user input as to what should stay and what should go and it seems like an extremely difficult problem trying to reconcile the user inputs away for the buildpack inputs. This buildpack is more of a way for user to configure what of their own files do they want on the final image. It seems counter intuitive to me at least to make this a buildpack leveraged by other buildpacks.

However, I think you did purpose an interesting having cake and eating it too scenario where we rewrite the buildpack slightly to be a wrapper for a removal library. That way buildpacks can benefit from the logic but we are not blurring the lines as to the purpose of the source removal buildpack.

This means that there is still a buildpack that gives users tight control over the files in the workspace that end up in the final image and buildpack authors get the benefit of shared complex logic.

I would love to hear other peoples thoughts on this!

ForestEckhardt: Again I think that these are changes that could be made after adoption as the overall user facing API for the buildpack would not change if we refactored the buildpack to be a wrapper for a library.

text/0039-source-removal-buildpack.md
dmikusa-pivotal: I'm Ok with suggesting them later.

text/0039-source-removal-buildpack.md
ForestEckhardt: I think that is a fair addition and sorry for confusion!

dmikusa-pivotal:
OK, I've been thinking about this more and it appears like some of my comments yesterday were not in line with the intention of this RFC. Sorry for the confusion. That said, I would like to request some clarification on a couple of points.

1. The summary only talks about contributing a buildpack. The implementation does not talk about how the buildpack will be contributed. If we're going to contribute a buildpack, I feel like this RFC needs to address that in the implementation. See the [Rust RFC](https://github.com/paketo-buildpacks/rfcs/blob/main/text/0014-rust.md) as an example.

2. There is also mention of env variables `BP_EXCLUDE_FILES` and `BP_INCLUDE_FILES`. It is unclear if the intent of this is to describe how the buildpack works or if the intent is to propose these as a standard UI that all Paketo buildpacks should follow. I think this needs to be clarified.

3. The RFC suggests that the Utilities team adopt this buildpack, but this RFC is not being proposed as a Utilities team RFC. Given my uncertainty voiced through #1 and #2, I'm not sure this is wrong, but I'm not sure this is right either. I think this is a point we should consider as I want to make sure that we're targeting the right team so that the correct set of voting members is allowed to weigh in on the RFC.

ForestEckhardt:
This may become an unnecessary buildpack with the merging of this [RFC](https://github.com/buildpacks/rfcs/blob/main/text/0097-official-utility-buildpacks.md#what-it-is), one of the proposed default utility buildpacks in a `purge` buildpack that will allow files to be ignored before export.

dmikusa-pivotal:
> This may become an unnecessary buildpack with the merging of this [RFC](https://github.com/buildpacks/rfcs/blob/main/text/0097-official-utility-buildpacks.md#what-it-is), one of the proposed default utility buildpacks in a `purge` buildpack that will allow files to be ignored before export.

Yes, that's true. I'm not sure how far out that will be, maybe it's still worth having this in Paketo until then. I recommended your buildpack to two or three people since you posted this. It's definitely useful.

When the time comes, you could submit your buildpack as a potential candidate for an official utility buildpack.

fg-j:
The idea of the buildpack wrapping a centralized source-removal package makes sense to me. This seems analogous to the environment variables buildpack, which provides buildpack end-users with functionality that buildpack authors leverage within buildpacks via packit/libcnb.

text/0039-source-removal-buildpack.md
ryanmoran: Can we discuss a bit more the concerns about having a buildpack set or modify the values of `BP_INCLUDE_FILES` and `BP_EXCLUDE_FILES`? It feels reasonable to me that a buildpack would do this so that it could signal downstream to the `source-removal` buildpack what should be kept or removed.

For example, I don't see why you couldn't solve the proposed scenario outlined by @dmikusa-pivotal using this API:

> At a high level:
>
>    1. Do a build, the artifact is stored in a layer so the build is cached
>    2. We delete source code
>    3. Artifact is taken from layer and moved to /workspace

I don't see why that couldn't be the following:

1. Do a build, the artifact is stored in a layer so the build is cached
2. Artifact is taken from layer and moved to /workspace
3. `BP_INCLUDE_FILES` is modified to include the artifact
4. `source-removal` buildpack clears the /workspace of everything except the artifact

I'm probably missing something. What's the concern here?

ForestEckhardt: I guess that is fair. The buildpack could append included files. I think that it gets tricky if a uses wants to use exclude instead there would be so required logic to do both which wouldn't be impossible it is just not implemented.

ForestEckhardt:
I am doing to put this in draft until I have had sometime to reach out to the CNB project about potentially making this one of their default utility buildpacks.

ForestEckhardt:
It seems as if making this a CNB default buildpack is out of scope for the time being so I am going to reopen RFC for discussion!

text/0039-source-removal-buildpack.md
ForestEckhardt: I have modified the buildpack so that it can run while both environment variables are set. It will first run the include files removal logic and then follow up with the exclude files removal logic. I think that this would give a fair pathway for buildpacks to ask for certain files to be included without interfering with the users environment variables

dmikusa-pivotal:
Questions from before: https://github.com/paketo-buildpacks/rfcs/pull/129#issuecomment-970314149

Also, thoughts on exposing the logic/code through a module that could be imported by other buildpacks? This buildpack would then also consume that module.

I don't think the Java buildpacks could use this buildpack directly, it's not going to fit into the workflow, but we currently have a request for more full-featured include/exclude logic like this. it would be nice to reuse the logic/code and have a consistent implementation. If you expose it as a module that we can go import, that would work perfect for us.

ForestEckhardt:
> OK, I've been thinking about this more and it appears like some of my comments yesterday were not in line with the intention of this RFC. Sorry for the confusion. That said, I would like to request some clarification on a couple of points.
> 
> 1. The summary only talks about contributing a buildpack. The implementation does not talk about how the buildpack will be contributed. If we're going to contribute a buildpack, I feel like this RFC needs to address that in the implementation. See the [Rust RFC](https://github.com/paketo-buildpacks/rfcs/blob/main/text/0014-rust.md) as an example.
> 2. There is also mention of env variables `BP_EXCLUDE_FILES` and `BP_INCLUDE_FILES`. It is unclear if the intent of this is to describe how the buildpack works or if the intent is to propose these as a standard UI that all Paketo buildpacks should follow. I think this needs to be clarified.
> 3. The RFC suggests that the Utilities team adopt this buildpack, but this RFC is not being proposed as a Utilities team RFC. Given my uncertainty voiced through [add initial structure #1](https://github.com/paketo-buildpacks/rfcs/pull/1) and [Add license 1 #2](https://github.com/paketo-buildpacks/rfcs/pull/2), I'm not sure this is wrong, but I'm not sure this is right either. I think this is a point we should consider as I want to make sure that we're targeting the right team so that the correct set of voting members is allowed to weigh in on the RFC.

@dmikusa-pivotal 
I think that I have resolved everything in this comment:
1. I have added a section that walks through the promotion process. If more details are needed please let me know.
2. I put the current implementation in a sub-section that hopefully clarifies that this is the current behavior of the buildpack and not an interface proposal.
3. I have re-aimed the RFC so that is applies only to the utilities sub-team.

If you have any other concerns feel free to let me know!

ryanmoran (approved):

dmikusa-pivotal (approved):

text/utilities/0004-source-removal-buildpack.md
```diff
@@ -33,7 +33,16 @@ ecosystem and as needed by particular users.
   benefit from such functionality.
 
 ## Implementation
+The Rust Buildpacks will be moved into the `paketo-buildpacks` Github
```
fg-j: A typo, I think?

ForestEckhardt: I think that is a typo I have already remedied maybe you are looking at an old commit

text/utilities/0004-source-removal-buildpack.md
```diff
@@ -0,0 +1,64 @@
...
+unnecessary source code from the final image as it causes image bloat and poses
+a potential security risk. In order to currently accomplish this, any buildpack
+that wants to remove source code must have logic to remove it from the
+`workspace` so that it does not appear in the final image, however this task is
+complicated because often times there are static files that are required for
+the app to run that should not be deleted from the `workspace`. This leads to
+multiple implementations of non-trivial file removal logic across different
+buildpacks.
+
+This could all be simplified by adding a buildpack that performs all source
```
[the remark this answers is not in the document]
ForestEckhardt: I think that the idea would be for both. Personally as a Go maintainer I would most likely remove the logic that currently exists in the buildpack in favor of this buildpack being appended to the order group. There has also be requests for this kind of behavior from users that are doing some more bespoke buildpack configurations or that want more control over what ends up in their final images.

ekcasey (approved):

arjun024 (approved):

thitch97 (approved):

PR #135, opened 2021-12-10
Add an RFC for secure runtime environments

samj1912:
Signed-off-by: Sambhav Kothari <skothari44@bloomberg.net>

[Readable](https://github.com/samj1912/rfcs-1/blob/runtime-env/text/0000-user-ids.md)

## Checklist


text/0000-user-ids.md
```diff
@@ -0,0 +1,42 @@
...
+## Summary
+
+This RFC recommends that Paketo Buildpacks should in general strive to avoid making run-time modifications to the Buildpacks layers or the application directory by default. Instead, they should consider the output image to be read-only by default and explicitly set the parts that need to be writeable during runtime.
+
+
+## Motivation
+
+As of Buildpacks [RFC 0085](https://github.com/buildpacks/rfcs/blob/main/text/0085-run-uid.md) the Cloud Native Buildpacks project recommends that the run-time and build-time user ids be different. The motivation behind this RFC was to improve the security of output images and prevent arbitrary runtime modifications to built artifacts.
+
+The run-time and build-time user can potentially share the same group id if they wish to make modifications to the output image during run-time, by changing the permissions of the files they wish to modify during run-time to be group writable during the build process. Ideally most buildpacks should use the temporary directories or other known scratch directories for ephemeral runtime files.
```
dmikusa-pivotal, line 12: >Ideally most buildpacks should use the temporary directories

What worries me about relying on temp directories is that these can be small and limited in capability. It's not uncommon for `/tmp` to be mounted with `noexec`, `nosuid`, and `nodev`, with `noexec` being the most likely problem (i.e. you can run something from there).

I don't really know if this is common in K8s though. Is that something you or other's have seen? Perhaps this is something we should investigate, as this would impact how buildpacks are configured to work.

text/0000-user-ids.md
```diff
@@ -0,0 +1,42 @@
...
+
+This RFC recommends that we identify buildpacks that making runtime modifications to the application directory or buildpack layers and modify them to instead use a temporary directory - or have them make these files group writable instead. All the current Paketo stacks' build-time and run-time users belong to a unique group so changing the file permissions to be group writable should have no additional security implications for users using the Paketo stacks.
+
+
+## Rationale and Alternatives
+
+
+The rationale behind this RFC is explained in the upstream RFC - [RFC 0085](https://github.com/buildpacks/rfcs/blob/main/text/0085-run-uid.md)
+
+In the future we could also modify the Paketo stacks to have different run-time/build-time user ids once all the buildpacks follow this practice for a more secure experience.
```
dmikusa-pivotal, line 24: No strong feelings one way or the other, but I'd be curious to know what folks think about doing this as part of this RFC? Why wait and have a second RFC? 

It kind of seems like it would be helpful to coordinate so that we're not making stack changes that break buildpacks.

text/0000-user-ids.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Implementation
+
+At a high level this would involve the following - 
+
+- Use temporary directories where possible to write files
+- Do the equivalent of `chmod g+w` for files that we are expecting to modify during runtime.
+
```
dmikusa-pivotal, line 32: Similar to #136, I think it would be helpful to define a more concrete plan here.

For example:

1. Create an issue to review & catalog our present compliance.
2. Open issues for non-compliant buildpacks to investigate making the change.
3. Open issues to switch where possible.
4. When buildpacks are tested, switch stacks to use different run/build ids.
5. Update Paketo documentation to explain this security precaution and any steps a user might need to take.

text/0000-user-ids.md
ryanmoran: We'd need to think through the impacts this would have on the `watchexec` features we provide. Most of this is relying on seeing updates to files in the `/workspace`.

fg-j: It's also worth thinking about whether there are common app setups that rely on modifying the app's working directory. One example is the .NET app in [this issue](https://github.com/paketo-buildpacks/dotnet-execute/issues/258).
If we made this change to stacks, I think it'd be necessary to provide end-users with a mechanism to reach in and mark some files as group writeable. That way, users have a way around Paketo's security opinion on this.

Perhaps, we create a buildpack that accepts a user-defined set of files through an environment variable and marks all of them group-writeable.

text/0000-user-ids.md
fg-j: The easiest and most comprehensive way to check for compliance might be to create a test stack that has different build and run user IDs. Integration and smoke tests across the repos would probably point out most of the places that will require changes.

text/0000-user-ids.md
ekcasey, line 12: Another reason to do this: in some hardened k8s environments the PodSecurityPolicy requires that the use of a read only root file system. In these environments a writable tmpfs is typically mounted to necessary locations like `/tmp` and `/run`.

However, in the above case we cannot use group level permissions as a workaround.

text/0000-user-ids.md
```diff
@@ -0,0 +1,42 @@
...
+The run-time and build-time user can potentially share the same group id if they wish to make modifications to the output image during run-time, by changing the permissions of the files they wish to modify during run-time to be group writable during the build process. Ideally most buildpacks should use the temporary directories or other known scratch directories for ephemeral runtime files.
+
+## Detailed Explanation
+
+This RFC recommends that we identify buildpacks that making runtime modifications to the application directory or buildpack layers and modify them to instead use a temporary directory - or have them make these files group writable instead. All the current Paketo stacks' build-time and run-time users belong to a unique group so changing the file permissions to be group writable should have no additional security implications for users using the Paketo stacks.
```
ekcasey, line 16: For ephemeral config files and the like I think we should explicitly prefer to write to a temporary directory so that the app container can run with a read only root filesystem. If we must write to the root filesystem we should (in addition to setting the group level permissions) ensure that, whenever possible, the location is configurable with a run-time environment variable so that there is a workaround.

text/0000-user-ids.md
ekcasey: +1 to the idea of creating a more concrete plan

Also, It might be good to include some examples in this RFC including common patterns (i.e. ephemeral config file, log file, etc)

samj1912:
Added a concrete implementation plan for supporting this feature.

text/0000-user-ids.md
dmikusa-pivotal: Sorry, I missed this comment. `noexec` means you can't execute a script from that location. I'm not sure if this is a scenario that comes up, but if you needed to write out a script to run and you can only write to `/tmp`, then you could not execute the script because of `noexec`.

I don't think Java buildpacks are doing this, and it's probably not a concern, but thought I would point it out and see if anyone else things that might affect their buildpacks. If not, feel free to resolve this.

text/0000-user-ids.md
```diff
@@ -0,0 +1,44 @@
...
+
+In the future we could also modify the Paketo stacks to have different run-time/build-time user ids once all the buildpacks follow this practice for a more secure experience.
+
+## Implementation
+
+At a high level this would involve the following - 
+
+- Create base stacks equivalent to the current Paketo stacks but with different runtime and build time user ids.
+- Create empty builders with the above stacks.
+- Run a parallel test pipeline which is optional and allowed to fail for each buildpack. This pipeline should utilize these new builders to gather the level of compliance for the current set of buildpacks. These tests should also provide appropriate directories for writing ephemeral files during runtime such as `/tmp` or `/run`. We can also use docker to mimic read only directories per https://www.thorsten-hans.com/read-only-filesystems-in-docker-and-kubernetes/
```
dmikusa-pivotal, line 32: What is the projected lifetime of this parallel test pipeline? It seems like a reasonable amount of overhead to have parallel pipelines. I'm not sure how others feel about that.

If it's not something we all need to do in the same way, I would prefer to just use this stricter running environment by default with the Java buildpacks. If anything breaks, then we'll fix it. That way we're not having a second pipeline specific to this task and we can continue to validate this is working in the future as well.

We're at around 30m running time for the present set of smoke tests for Java apps and we're talking about adding more tests. I don't really want to duplicate that and double the runtime, if possible.

text/0000-user-ids.md
```diff
@@ -0,0 +1,44 @@
...
+- Track the compliance status and update buildpacks as necessary to comply with this RFC. The majority changes that would be needed would be for runtime executables/scripts provided by buildpacks such as `profile.d` or `exec.d` interfaces. Seperately we would also need to track configuration locations for configurations provided by buildpacks which lead to files being written on the root filesystem, such as log files etc. and configure them to write to ephemeral locations such as `/tmp` or the runtime user's home directory.
+- We should create a utility buildpack that allows users to specific path patterns in `gitignore` format to make such directories group writable.
```
dmikusa-pivotal, line 34: +1 - For clarification, is the intent to do this through Paketo or Utility Buildpacks?

text/0000-user-ids.md
```diff
@@ -0,0 +1,44 @@
...
+- Once all buildpacks comply, we can release the new set of builders with hardened security environments to Paketo users.
```
dmikusa-pivotal, line 35: Am I understanding this correctly? This says that the desired state would be to have a second set of Paketo builders. So there would be the present set plus another that's roughly the same but with these modifications?

text/0000-user-ids.md
```diff
@@ -0,0 +1,44 @@
...
+- At some point in the future we can deprecate the current less-secure builders and stacks.
```
dmikusa-pivotal, line 36: I'm curious what people think would be good for the amount of time these overlap? I think I'd like to see something more specific in this regard. There's overhead in maintaining two sets of builders and without some specified period if time, I worry that this might just turn into project cruft that accumulates.

text/0000-user-ids.md
ekcasey: We could update existing tests to run with a different UID even if that means changing the user at runtime to be different than the default on our existing run images. We can make the change for real in the next major version of the stacks (`focal`??)

text/0000-user-ids.md
ekcasey: Paketo first imo. We can always move it to the CNB project if that is desirable

text/0000-user-ids.md
ekcasey, line 35: Because of the danger of breaking app rebase if we change the default user in our existing stacks. I suggest we introduce this new security posture in new stacks, but leave the bionic stacks as they are.

text/0000-user-ids.md
ekcasey: Can we just introduce in new stacks only and use the end of standard support of bionic in 2023 as a natural phase out?

text/0000-user-ids.md
dmikusa-pivotal: I'm not thinking of a good reason why you'd want to do this. I think if you had an app that was going to shell out and run some script, it would likely contain that script in the app. But let's say that you had an app that generated some commands to run as a script, wrote it out, and run the script. That would not work. Probably not a big deal, so I'm OK closing out this comment thread.

>buildpacks should be able to accomplish anything they need to with exec.d right?

IMHO No. You can't modify start commands from exec.d. It doesn't come up often, but I know that has been limiting at least once. Not worth discussing here. Just me ranting a little.

ryanmoran:
I'm wondering if we could get the same testing coverage by only doing the part where we create a new run image with a user id that differs from our build image. In this scenario, we'd have tests that run `pack build <app> --run-image <run-image-with-diff-user-id>`. So, for each stack, we would have 3 images (build with uid X, run with uid X, and run with uid Y).

Then, like @ekcasey mentions, once we start moving to Ubuntu 22.04 later this spring, we can intentionally switch to only having one run image for these new stacks, but with a user id that is different from the build image.

samj1912:
@ekcasey, @dmikusa-pivotal, @ryanmoran I have updated the RFC as per the comments. I didn't want to presume too much about the focal stack rollout which is why I had left things ambiguous previously, but based on the feedback from the last WG meeting, I have include a line saying that we may target focal based stacks for this RFC.

ryanmoran:
> @ekcasey, @dmikusa-pivotal, @ryanmoran I have updated the RFC as per the comments. I didn't want to presume too much about the focal stack rollout which is why I had left things ambiguous previously, but based on the feedback from the last WG meeting, I have include a line saying that we may target focal based stacks for this RFC.

We're gonna support Jammy (2022.04) first.

text/0000-user-ids.md
ekcasey, line 16:
```suggestion
This RFC recommends that we identify buildpacks that make runtime modifications to the application directory or buildpack layers and modify them to instead use a temporary directory - or have them make these files group writable instead. All the current Paketo stacks' build-time and run-time users belong to a unique group so changing the file permissions to be group writable should have no additional security implications for users using the Paketo stacks.
```

ekcasey (approved):

ryanmoran (approved):

dmikusa-pivotal (approved):

PR #136, opened 2021-12-10
Add an RFC for removing shell specific interfaces

samj1912:
## Summary

Remove shell specific interfaces.
[Readable](https://github.com/samj1912/rfcs-1/blob/direct/text/0000-direct.md)

## Use Cases

See RFC.

## Checklist

### DIFF
--- a/text/0000-direct.md
+++ b/text/0000-direct.md
@@ -2,7 +2,7 @@
 
 ## Summary
 
-This RFC proposes that we move all Paketo buildpacks to use `direct` process types and `exec.d` interface instead of `profile.d` interface and `direct = false` process types.
+This RFC proposes that we move all Paketo buildpacks to use `direct` process types and `exec.d` interface instead of `profile.d` interface and `direct = false` process types wherever possible without breaking backwards compatibility from an end user perspective.
 
 ## Motivation
 
--- a/text/0000-direct.md
+++ b/text/0000-direct.md
@@ -1,0 +1,39 @@
+# Use direct processes and exec.d
+
+## Summary
+
+This RFC proposes that we move all Paketo buildpacks to use `direct` process types and `exec.d` interface instead of `profile.d` interface and `direct = false` process types.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


dmikusa-pivotal:
+1 for the idea.

In regards to the RFC, I would like to see an implementation section. I think my idea would be to perhaps have a bullet list that includes the actions we need to take here. 

1. Create an issue to review & catalog our present compliance.
2. Open issues for non-compliant buildpacks to investigate making the change.
3. Open issues to switch where possible.

It might also be interesting to talk about a process for cases where we need to change APIs. On one hand, we can't just make changes and break users, but on the other, we need a plan to migrate buildpacks to use this new method. Or perhaps we could defer this decision to each individual buildpack that needs changes to be compliant.

fg-j:
Do we have any idea when this is expected to land in the lifecycle?

ekcasey:
@fg-j 

> Do we have any idea when this is expected to land in the lifecycle? Here's the issue to implement the change. I don't see it added to any milestone.

The short answer is that it is currently unscheduled. Since this RFC removes support for `<app>/.profile` scripts from the launcher, some core team members are uncomfortable implementing this change before we approve a version of System Buildpacks which would allow platforms to provide support for `<app>/.profile` script via a buildpack.

ekcasey (approved):

dmikusa-pivotal:
@ekcasey 

> > It might also be interesting to talk about a process for cases where we need to change APIs. On one hand, we can't just make changes and break users
> 
> Are there any cases where we would need to change user facing APIs? I believe we should be able to move to direct processes without affecting end users.

The only thing coming to mind at the moment is `.profile`. It is not something that I think we widely advertise, but it is supported and user-facing. That won't continue to work. Ideally, we'll have a workaround so users hopefully won't have to care.

I think my comment more was to make sure that we are considering the possible implications this might cause. Java buildpacks have been using direct for a while so are unlikely to be impacted, but each language family will probably need to evaluate the question of what might break, if anything.

ryanmoran (approved):

dmikusa-pivotal (approved):

dmikusa-pivotal:
We discussed this in the WG meeting today. In principle, everyone is in agreement that this is a good proposal. We've got enough approvals and this will get merged.

In practice, there are some concerns for specific language families where the impact of losing shell scripting and `.profile` script access might cause problems for users. We discussed leaving the implementation of this up to the individual language family. Allowing them to determine the impact and set a timeline to implement this RFC. If that means waiting until there is a profile buildpack, or some other similar means of solving that problem, that is OK and can be determined at the language team's discretion. 

After merging, I'll create an issue where we can track at the language family level where this has been implemented.

PR #137, opened 2021-12-13
RFC: `buildpack.yml` to Environment Variables for PHP

ForestEckhardt:
[Readable](https://github.com/paketo-buildpacks/rfcs/blob/buildpack-yml-to-env-var-php/text/php/0002-buildpack-yml-to-env-var.md)
https://github.com/paketo-buildpacks/php/issues/472

### DIFF
--- a/text/php/0002-buildpack-yml-to-env-var.md
+++ b/text/php/0002-buildpack-yml-to-env-var.md
@@ -76,6 +76,17 @@ php:
  serveradmin: admin@localhost
 ```
 
+### `BP_PHP_ENABLE_HTTPS_REDIRECT`
+```shell
+$BP_PHP_ENABLE_HTTPS_REDIRECT="true"
+```
+
+This will replace the following structure in `buildpack.yml`:
+```yaml
+php:
+ enable_https_redirect: true
+```
+
 ### Configuration Removal
 The following structure in `buildpack.yml` will not be receiving an environment
 variable configuration option:
--- a/text/php/0002-buildpack-yml-to-env-var.md
+++ b/text/php/0002-buildpack-yml-to-env-var.md
@@ -127,22 +127,6 @@ This will replace the following structure in `buildpack.yml`:
 composer:
   install_options: ["--no-dev", "--prefer-install=auto"]
 ```
-### `BP_COMPOSER_VENDOR_DIR`
-```shell
-$BP_COMPOSER_VENDOR_DIR="vendor"
-```
-
-Note: The value of this environment variable will be used to set
-[`$COMPOSER_VENDOR_DIR`](https://getcomposer.org/doc/03-cli.md#composer-vendor-dir)
-therefore if a user attempts to set this value themselves as well as set the
-value of `$BP_COMPOSER_VENDOR_DIR` the value of `$COMPOSER_VENDOR_DIR` will be
-overwritten resulting in potentially unexpected behavior.
-
-This will replace the following structure in `buildpack.yml`:
-```yaml
-composer:
-  vendor_directory: vendor
-```
 ### `BP_COMPOSER_GLOBAL_INSTALL_OPTIONS`
 ```shell
 $BP_COMPOSER_GLOBAL_INSTALL_OPTIONS="--only-name --type"
@@ -157,8 +141,8 @@ composer:
 ```
 
 ### Configuration Removal
-The following structure in `buildpack.yml` will not be receiving an environment
-variable configuration option:
+The following structure in `buildpack.yml` will not be receiving a buildpack
+specific environment variable configuration option:
 ```yaml
 composer:
   json_path: composer
@@ -167,6 +151,21 @@ This structure is effectively performing the function of the
 [`$COMPOSER`](https://getcomposer.org/doc/03-cli.md#composer) environment
 variable that is natively supported by `composer`. Therefore the structure is
 being removed in favor of the native solution.
+
+The following structure in `buildpack.yml` will not be receiving a buildpack
+specific environment variable configuration option:
+```yaml
+composer:
+  vendor_directory: vendor
+```
+If the user would like to set a custom `composer` vendoring location they can
+use the
+[`$COMPOSER_VENDOR_DIR`](https://getcomposer.org/doc/03-cli.md#composer-vendor-dir)
+environment variable native to `composer`.
+Note:
+The value of `$COMPOSER_VENDOR_DIR` will be modified by the buildpack in order
+to set up efficient caching; these changes will be logged for the user to
+observe and will also be documented.
 
 ---
 
--- a/text/php/0002-buildpack-yml-to-env-var.md
+++ b/text/php/0002-buildpack-yml-to-env-var.md
@@ -189,7 +189,6 @@ migrating to the environment variable configuration.
 
 ## Unresolved Questions and Bikeshedding
 
-- Are there any environment variable names that are out of place?
 - Should we remove the `BP_PHP_SERVER` configuration in favor a multi-buildpack
   build as outlined in [this comment thread](https://github.com/paketo-buildpacks/php/issues/472)
   on the issue for this RFC.
--- a/text/php/0002-buildpack-yml-to-env-var.md
+++ b/text/php/0002-buildpack-yml-to-env-var.md
@@ -76,30 +76,6 @@ php:
  serveradmin: admin@localhost
 ```
 
-### `BP_REDIS_SESSION_STORE_SERVICE_NAME`
-```shell
-$BP_REDIS_SESSION_STORE_SERVICE_NAME="redis-session"
-```
-
-This will replace the following structure in `buildpack.yml`:
-```yaml
-php:
- redis:
-   session_store_service_name: redis-sessions
-```
-
-### `BP_MEMCACHED_SESSION_STORE_SERVICE_NAME`
-```shell
-$BP_MEMCACHED_SESSION_STORE_SERVICE_NAME="memcached-session"
-```
-
-This will replace the following structure in `buildpack.yml`:
-```yaml
-php:
- memcached:
-   session_store_service_name: memcached-sessions
-```
-
 ### Configuration Removal
 The following structure in `buildpack.yml` will not be receiving an environment
 variable configuration option:
@@ -114,6 +90,17 @@ the Procfile buildpack will be added to all order groupings. Because of this
 addition I am proposing the removal of this configuration option in the
 environment variables in favor of encouraging users to use a Procfile if they
 need to set a custom start script.
+
+```yaml
+php:
+ redis:
+   session_store_service_name: redis-sessions
+ memcached:
+   session_store_service_name: memcached-sessions
+```
+Both of these configuration options are going to be replaced with standardized
+service binding types. The types will be `php-redis-session` and
+`php-memcached-session` respectively.
 
 ---
 The proposed environment variables for Composer are as follows:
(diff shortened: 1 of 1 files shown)


text/php/0002-buildpack-yml-to-env-var.md
```diff
@@ -0,0 +1,222 @@
...
+composer:
+  install_global: ["--only-name", "--type"]
+```
+
+---
+
+### Deprecation Strategy
+In order to facilitate a smooth transition from `buildpack.yml`, the buildpack
+should will support both configuration options with environment variables
+taking priority or `buildpack.yml` until the 1.0 release of the buildpack. The
```
sophiewigmore, line 195: A majority of the existent buildpacks in the PHP language family are going to be rewritten following https://github.com/paketo-buildpacks/php/issues/485. Should net-new buildpacks have to implement support for both `buildpack.yml` as well as environment variables to ensure backwards compatibility?

ForestEckhardt: I think that is a pretty valid question. My impression was that at the end of the rewrite while we would have different buildpacks the overall functionality should remain the same. If that is not the case and we are already breaking the high level API so to speak then I think that we should deprecate the `buildpack.yml` at the same time as a large API change.

sophiewigmore: The main goal of the rewrite work is meant to increase modularity and maintainability. I think you are right in saying that the functionality should more or less stay the same. However, in https://github.com/paketo-buildpacks/rfcs/blob/main/text/php/0001-restructure.md I don't see any mention of maintaining `buildpack.yml` support..so it isn't clear cut.

I think it probably makes the most sense to include `buildpack.yml` support for backwards compatibility, BUT it pains me since we are largely moving away from `buildpack.yml` support across all of our buildpacks..

dmikusa-pivotal: The plan discussed yesterday:

1. Add deprecation messages for buildpack.yml to the current version of the buildpack. Something like `Use of 'buildpack.yml' has been deprecated and will be removed in the next major branch of the buildpack.`
2. In the new buildpack, we don't support buildpack.yml at all. Including one will generate an error so that it actively fails and users are forced to address it (i.e. don't silently miss this change).
3. In the new buildpack, we will support config through env variables.
4. We'll have a migration doc explaining how to convert settings from buildpack.yml to env variables.

sophiewigmore: Filing issues across PHP for this now

sophiewigmore: https://github.com/paketo-buildpacks/php-web/issues/216
https://github.com/paketo-buildpacks/php-dist/issues/329
https://github.com/paketo-buildpacks/php-composer/issues/265

thitch97 (approved):

text/php/0002-buildpack-yml-to-env-var.md
```diff
@@ -0,0 +1,222 @@
...
+[`$COMPOSER_VENDOR_DIR`](https://getcomposer.org/doc/03-cli.md#composer-vendor-dir)
+therefore if a user attempts to set this value themselves as well as set the
+value of `$BP_COMPOSER_VENDOR_DIR` the value of `$COMPOSER_VENDOR_DIR` will be
+overwritten resulting in potentially unexpected behavior.
+
+This will replace the following structure in `buildpack.yml`:
+```yaml
+composer:
+  vendor_directory: vendor
+```
```
dmikusa-pivotal, line 158: This seems unnecessary. As the comment mentions, it could result in confusion too. Which one does a user set `COMPOSER_VENDOR_DIR` or `BP_COMPOSER_VENDOR_DIR`. It seems like we could just drop the `BP_` version and if a user wants to set the vendor dir, then they set that specifically. 

I generally like to use existing env variable options when available and since this is just being passed through to composer, that seems reasonable here.

Thoughts?

text/php/0002-buildpack-yml-to-env-var.md
```diff
@@ -0,0 +1,222 @@
...
+[`$COMPOSER_HOME`](https://getcomposer.org/doc/03-cli.md#composer-home)
+therefore if a user attempts to set this value themselves as well as set the
+value of `$BP_COMPOSER_HOME` the value of `$COMPOSER_HOME` will be overwritten
+resulting in potentially unexpected behavior.
+
+This will replace the following structure in `buildpack.yml`:
+```yaml
+composer:
+  json_path: composer
+```
```
dmikusa-pivotal, line 175: Same thoughts here. If all we're doing is setting `COMPOSER_HOME` then we should probably drop this config value and instruct users to set `COMPOSER_HOME` directly for this functionality.

text/php/0002-buildpack-yml-to-env-var.md
dmikusa-pivotal: I think this section needs to be rewritten given the plan discussed here.

text/php/0002-buildpack-yml-to-env-var.md
```diff
@@ -0,0 +1,222 @@
...
+```shell
+$BP_MEMCACHED_SESSION_STORE_SERVICE_NAME="memcached-session"
+```
+
+This will replace the following structure in `buildpack.yml`:
+```yaml
+php:
+ memcached:
+   session_store_service_name: memcached-sessions
+```
```
dmikusa-pivotal, line 101: I suspect that we can do away with these. Right now, it's being used to specify the name of the service binding, but we should probably just switch this to a documented service binding type. Thus you'd have a type like `PHPRedisSessions` and `PHPMemcachedSessions` that would be fixed. The user would then create a binding with this type if they want to enable this functionality. The binding name is then irrelevant and you don't need a configuration option for it.

dmikusa-pivotal (changes requested):

text/php/0002-buildpack-yml-to-env-var.md
ForestEckhardt: I think that the only potential issue is that currently we set these relative the working directory by joining the working directory with the given value. We would lose that functionality of prepending the working directory and I do not know if composer would be able to handle that.

text/php/0002-buildpack-yml-to-env-var.md
ForestEckhardt: I think that this has the same potential drawback as the environment variable above so I think whatever decision we make there we should copy for consistency sake.

text/php/0002-buildpack-yml-to-env-var.md
ForestEckhardt: I am more than happy with that solution! I will make the proper changes.

ForestEckhardt: Type names have been proposed in the removed configuration section

text/php/0002-buildpack-yml-to-env-var.md
dmikusa-pivotal: Actually, for this one I'm not sure that `$COMPOSER_HOME` is the right match. `$COMPOSER_HOME` is changing the global configuration directory. This is where composer loads defaults from. While json_path was meant to allow the user to change where the composer.json file is located.

I think it more closely correlates to the `$COMPOSER` setting, which tells COMPOSER where to find composer.json. We were not using that setting previously though. Previously, we'd "find" the composer.json using the path trick above and then launch `composer` processes from within that directory (i.e. set the working directory). 

It's probably better to drop that and support using `$COMPOSER` for that. 

Setting `$COMPOSER_HOME` is something the buildpack will need to do and something we'll discourage users from doing. I suspect that stuff will go into a layer so it's cached from run to run. The current buildpack is doing that.

text/php/0002-buildpack-yml-to-env-var.md
dmikusa-pivotal: Yes, that's a concern. The other issue here might be with caching. The actual vendor directory should probably be to a layer so we can cache it, while Composer and the app need to see a symlink in the `/workspace` because that is what they think.

I still think we should probably use the default env variable, but we may need to do some special preprocessing on it before we pass it through to Composer. Probably use that as the location for the symlink, since from the end-user perspective that's what they care about.

text/php/0002-buildpack-yml-to-env-var.md
ForestEckhardt: Thanks for this info. I completely missed the `$COMPOSER` environment variable in my reading. I will make the switch and update it accordingly because I am a big fan of using native configuration.

ForestEckhardt: Alright this has been reworked and should be more in line with what you were talking about.

text/php/0002-buildpack-yml-to-env-var.md
ForestEckhardt: I have some concerns using a native configuration option and then doing preprocessing in the buildpack. I think that doing this leads to confusion and unexpected behavior from users that are expecting things to work identically. I think that there are things that we could do to make the processing very transparent to the end user I just want your take on the drawbacks in my opinion before going in and changing it.

dmikusa-pivotal: I think as long as we can guarantee two things it's ok.  1.) whatever additional processing we're doing is transparent to the end user and 2.) the setting works just like the user would typically expect when running locally or on a VM.

My take is that it should. If a user sets `COMPOSER_VENDOR_DIR` locally, that changes the directory where composer will write those files. In the buildpack, I think what we'll need to do is read `COMPOSER_VENDOR_DIR` and add a symlink to our build+cache layer at the `COMPOSER_VENDOR_DIR` location. That way when composer runs, it follows the symlink to the layer but at the same time things just look like the vendor dir already exists. Composer should add stuff to the specified location and it'll "just work". That's my theory on it at least.

The whole thing is tricky though because many PHP apps expect the vendor directory to exist along side the application files, which is in `/workspace`. We can't write them directly there, otherwise we need to run `composer install` every time which is slow or copy files from the cache layer to `/workspace` which is also slow. In this case, we just need to recreate the symlink every time, which is fast.

ForestEckhardt: Alright I have updated this segment to hopefully include everything that would be needed in this RFC. Please take one more look over when you have time.

dmikusa-pivotal (approved):

sophiewigmore:
Hi! I noticed that [php-web has another `buildpack.yml` field](https://github.com/paketo-buildpacks/php-web/blob/5bcb1f838fdbf16964e235d9445920cd8e3fbcc1/config/config.go#L136) for `EnableHTTPSRedirect`. Should that be added here?

ForestEckhardt:
@sophiewigmore This is a good call out! I didn't even know it was a thing cause it is not documented in the README. I think that I would need to hear from @paketo-buildpacks/php-maintainers about whether or not this is still needed. I would be happy to add it if necessary. I think it would need to be supported in the individual server buildpacks.

dmikusa-pivotal:
That setting is used by the template generation so that php-web can emit HTTPD or Nginx config that will redirect a user to the HTTPS.

That is something we should continue to support.

It would go in php-nginx/php-httpd according to the RFC.  https://github.com/paketo-buildpacks/rfcs/blob/main/text/php/0001-restructure.md#buildpacks

ForestEckhardt:
@dmikusa-pivotal I think that my latest change should cover this. I am going to re-request review from you!

dmikusa-pivotal (approved):

thitch97 (approved):

PR #138, opened 2022-01-04
A proposal to alter the builder order

dmikusa-pivotal:
## Summary

A proposal to alter the builder order and move Node.js lower.

## Use Cases

This prevents Node.js buildpack from incorrectly detecting true for applications that also bundle frontend code with backend code.

## Checklist

* [x] I have viewed, signed, and submitted the Contributor License Agreement.
* [x] I have linked issue(s) that this PR should close using keywords or the Github UI (See [docs](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue))
* [x] I have added an integration test, if necessary.
* [x] I have reviewed the [styleguide](https://github.com/paketo-buildpacks/community/blob/main/STYLEGUIDE.md) for guidance on my code quality.
* [x] I'm happy with the commit history on this PR (I have rebased/squashed as needed).

PR #139, opened 2022-01-10
Adds RFC for Javascript Frontend support in Web Servers

ryanmoran:
## Summary

[Readable](https://github.com/paketo-buildpacks/rfcs/blob/javascript-frontends/text/web-servers/0003-javascript-frontend.md)

## Checklist

### DIFF
--- a/text/web-servers/0003-javascript-frontend.md
+++ b/text/web-servers/0003-javascript-frontend.md
@@ -90,6 +90,37 @@ following order groupings:
 
   [[order.group]]
     id = "paketo-buildpacks/nginx"
+
+[[order]]
+
+  [[order.group]]
+    id = "paketo-buildpacks/node-engine"
+
+  [[order.group]]
+    id = "paketo-buildpacks/yarn"
+
+  [[order.group]]
+    id = "paketo-buildpacks/yarn-install"
+
+  [[order.group]]
+    id = "paketo-buildpacks/node-run-script"
+
+  [[order.group]]
+    id = "paketo-buildpacks/httpd"
+
+[[order]]
+
+  [[order.group]]
+    id = "paketo-buildpacks/node-engine"
+
+  [[order.group]]
+    id = "paketo-buildpacks/npm-install"
+
+  [[order.group]]
+    id = "paketo-buildpacks/node-run-script"
+
+  [[order.group]]
+    id = "paketo-buildpacks/httpd"
 ```
 
 In order to detect either of these groups, users will need an application that
@@ -97,8 +128,9 @@ is using either NPM or Yarn to install their packages. These are detected by
 the presence of a `package.json` or `yarn.lock` file. Additionally, users will
 need to specify the `BP_NODE_RUN_SCRIPTS` environment variable indicating how
 to go about building their application. Finally, users will need to include an
-`nginx.conf` file that is configured to serve the build result static files.
+`nginx.conf` or `httpd.conf` file that is configured to serve the build result
+static files.
 
 ## Unresolved Questions and Bikeshedding
 
-* Should we support more than just `nginx` as the web server?
+* What does this mean for this buildpack's position in the builder?
--- a/text/web-servers/0003-javascript-frontend.md
+++ b/text/web-servers/0003-javascript-frontend.md
@@ -1,0 +1,84 @@
+# Support for JavaScript Frontend applications
+
+## Summary
+
+The Web Servers language family should support JavaScript frontend applications
+that required a build process to transform their codebase into a deployable
+artifact. The Web Servers language family will adopt several of the buildpacks
+available in the Node.js language family to create an order grouping capable of
+building these types of frontend applications.
+
+## Motivation
+
+It is a common practice for web applications to be split between a backend that
+serves an API and a frontend that runs in a browser. Today, we have many
+choices to help developers deploy their backend applications in many of the
+most popular programming languages. However, for developers looking to build a
+container capable of serving their frontend, we do not have a solution that
+works well out of the box.
+
+## Detailed Explanation
+
+All of the parts necessary to develop this addition to the Web Server language
+family already exist, but have not yet been pulled together into a consistent
+whole. Specifically, we will be adding the existing `node-engine`,
+`npm-install`, `yarn`, `yarn-install`, and `node-run-script` buildpacks to the
+order groupings list to enable this outcome.
+
+## Rationale and Alternatives
+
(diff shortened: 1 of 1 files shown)


arjun024:
Great proposal! I think it would be great to see a mention of adding the ability to remove the source from the built app image. `node_modules` especially (which in a case like this is useless at runtime) is notorious for its large size.

sophiewigmore:
Finally got a chance to look at this. It makes sense to me that the order groupings to go in the web server language family buildpack. However, I wonder if it would be confusing to a user that this use case is supported in the web servers buildpack rather than the Node.js buildpack? In the RFC, could you enumerate the main rationale for including this here rather than in the Node.js buildpack?

dmikusa-pivotal:
> Great proposal! I think it would be great to see a mention of adding the ability to remove the source from the built app image. `node_modules` especially (which in a case like this is useless at runtime) is notorious for its large size.

Yes, I second that. 

What about Node itself? Is this combination going to result in Node being available in the launch layer? I feel like that would not be ideal and it would bloat the image.

dmikusa-pivotal:
1. I don't have a strong opinion, but I am curious why there are two order groups rather than having one order group w/optional buildpacks (i.e. yarn & yarn-install). In the present state, it seems OK, but if you have more web servers to support, that's going to multiply out and result in 2x order groups per web server.

2. Maybe not relevant yet, but where do we plan for the webservers composite buildpack to fit into builders? I ask because of https://github.com/paketo-buildpacks/rfcs/pull/138 and the fact that this would pull in the same Node.js buildpacks which would match any app that contains a `package.json`, which is a broad match. We're seeing already that it could be a Node.js app, a Java app and now a web servers app.

3. > Should we support more than just nginx as the web server?

We're calling this "Web Servers" buildpack so I would want to see support for multiple web servers. We have an HTTPD buildpack, so that seems like an easy place to start. It seems like it wouldn't be much additional work to support this. I haven't tried it or thought this through a ton, but if you supply `nginx.conf` then detection matches the order group w/Nginx. If you supply an `httpd.conf` file then detection matches the order group w/HTTPD.

ryanmoran:
> Great proposal! I think it would be great to see a mention of adding the ability to remove the source from the built app image. `node_modules` especially (which in a case like this is useless at runtime) is notorious for its large size.

For what it's worth, the `node_modules` directory will be empty already in this configuration. The build plan requirement graph results in a situation where `node` and `node_modules` are only required for the `build` phase. In the worst case, we'd have a dangling `node_modules` symlink that points to a layer that does not exist on the built image. In the vast majority of cases, this should be fine, but we could think about making sure this also gets removed.

I think we should let #129 play out to decide how this should be handled. I'd rather see a generalized solution for this problem than a piecemeal solution that provided differing interfaces or semantics. @ForestEckhardt any idea what is happening on the source removal front?

ryanmoran:
> Finally got a chance to look at this. It makes sense to me that the order groupings to go in the web server language family buildpack. However, I wonder if it would be confusing to a user that this use case is supported in the web servers buildpack rather than the Node.js buildpack. I'm thinking about this from the point of view that we support running an Nginx web server with PHP in the PHP language family as outlined in https://github.com/paketo-buildpacks/rfcs/blob/main/text/php/0001-restructure.md.
> 
> I don't see this possible confusion as a huge issue, as long as the functionality is supported and well-documented. Nevertheless, it would be helpful if in the RFC you could enumerate the main rationale for including this here rather than in the Node.js buildpack?

@sophiewigmore, the primary thinking here is about what the buildpack is producing. To me, the Node.js buildpack is about producing an image containing an application that runs Node.js inside the container. This is commonly a backend server application, but could also be something like a React server that is acting in both a backend and frontend capacity. Either way, the image that gets built contains `node`.

That isn't really the case for what is proposed in this RFC. Instead, `node` is really just needed during the build process to convert some JavaScript into a bundled static file that is then served by a web server. This type of application is a pattern that can be found in a number of language ecosystems beyond just Node.js. For example, Hugo (Go), Jekyll (Ruby), MkDocs (Python), and mdBook (Rust) all provide a similar process to convert some intermediate development files into a set of static assets to be served by a web server. In the future, I could imagine us also including optimized buildpack groups to support these other toolkits.

I've added a new paragraph in the `Rationale and Alternatives` section that outlines my thinking.

ryanmoran:
> 1. I don't have a strong opinion, but I am curious why there are two order groups rather than having one order group w/optional buildpacks (i.e. yarn & yarn-install). In the present state, it seems OK, but if you have more web servers to support, that's going to multiply out and result in 2x order groups per web server.

@dmikusa-pivotal, the reasoning here is that the `npm-install` and `yarn/yarn-install` detection logic have an overlap that precludes them from being grouped together. Specifically, they would both detect in the `yarn-install` case as the workspace would contain a `package.json` as well as a `yarn.lock` file. This would result in a build that first used `npm` to install any packages only to then use `yarn` to reinstall those the packages again (likely different versions given their differing package locking semantics).

ryanmoran:
>     2. Maybe not relevant yet, but where do we plan for the webservers composite buildpack to fit into builders? I ask because of [A proposal to alter the builder order #138](https://github.com/paketo-buildpacks/rfcs/pull/138) and the fact that this would pull in the same Node.js buildpacks which would match any app that contains a `package.json`, which is a broad match. We're seeing already that it could detect a Node.js app, a Java app and now a web servers app.

@dmikusa-pivotal there is some extra detection criteria here. The user would also need to include an `nginx.conf` file on the filesystem. This is also why we shouldn't probably group the web servers (`nginx` and `httpd`) together in a single group with an optional flag. Ensuring that at least one of them detects in this case is important to disambiguate the build plan.

As far as what this means for ordering in the builders, I would assume that this means the Web Server buildpack would need to appear before the Node.js buildpack as the Node.js detection criteria is less specific and would therefore end up detecting when users really want the Web Servers buildpack to detect.

ryanmoran:
> We're calling this "Web Servers" buildpack so I would want to see support for multiple web servers. We have an HTTPD buildpack, so that seems like an easy place to start. It seems like it wouldn't be much additional work to support this. I haven't tried it or thought this through a ton, but if you supply `nginx.conf` then detection matches the order group w/Nginx. If you supply an `httpd.conf` file then detection matches the order group w/HTTPD.

I've added new group to the order that also detect for `httpd`.

dmikusa-pivotal (approved):

arjun024 (approved):

thitch97 (approved):

PR #140, opened 2022-01-19
RFC: Credential Configuration in the Git Buildpack

ForestEckhardt:
[Readable](https://github.com/paketo-buildpacks/rfcs/blob/git-ssh-token/text/utilities/0003-git-credential-configuration.md)


text/utilities/0003-git-ssh-token.md
```diff
@@ -0,0 +1,41 @@
...
+where the token will be provided in a mounted volume through a service binding.
+The `type` for the service binding will be `git-ssh-token`.
+
+That token will then be used to change the global `git` config for that
+container which will persist for the entirety of the build which the [CNB
+specification](https://github.com/buildpacks/spec/blob/main/buildpack.md#requirements)
+states must all happen in the same container for security purposes. The
+following commands will be used to modify the global `git` config:
+```bash
+git config --global url."https://api:{{token}}@github.com/".insteadOf "https://github.com/"
```
samj1912, line 33: This seems very specific to github, how would we handle generic git servers?

samj1912, line 33: This seems very specific to github, how will we handle generic git servers?

text/utilities/0003-git-ssh-token.md
```diff
@@ -0,0 +1,41 @@
...
+variable like most of our other buildpack configuration as this poses a risk of
+exposure by view the build process in some form of CI. Because of this
+restriction, an approach similar to the CA Certificates Buildpack will be used
+where the token will be provided in a mounted volume through a service binding.
+The `type` for the service binding will be `git-ssh-token`.
+
+That token will then be used to change the global `git` config for that
+container which will persist for the entirety of the build which the [CNB
+specification](https://github.com/buildpacks/spec/blob/main/buildpack.md#requirements)
+states must all happen in the same container for security purposes. The
```
samj1912, line 30: Given that the config below will be stored in plain text in git, should we instead opt for https://git-scm.com/docs/git-credential-store?

text/utilities/0003-git-ssh-token.md
```diff
@@ -0,0 +1,41 @@
...
+ability to add SSH authentication to the project universally. 
+
+## Implementation
+
+For security purposes, the token cannot be provided through an environment
+variable like most of our other buildpack configuration as this poses a risk of
+exposure by view the build process in some form of CI. Because of this
+restriction, an approach similar to the CA Certificates Buildpack will be used
+where the token will be provided in a mounted volume through a service binding.
+The `type` for the service binding will be `git-ssh-token`.
```
samj1912, line 25: I believe these are generally git tokens or passwords. Using the term ssh token might be a bit misleading since at the end, the protocol that is used is https.

samj1912:
Some further discussion on the implementation of this rfc took place in paketo slack. Please check out https://paketobuildpacks.slack.com/archives/C011S6EL49L/p1642632942011600

ForestEckhardt:
@samj1912 Has brought up some great points that I think are going to lead to a complete overhaul of the implementation section. I am going to draft this PR and reopen it once I have done an overhaul.

samj1912 (approved):

text/utilities/0003-git-credential-configuration.md
```diff
@@ -0,0 +1,59 @@
...
+
+For security purposes, the credentials cannot be provided through an
+environment variable like most of our other buildpack configuration as this
+poses a risk of exposure by view the build process in some form of CI. Because
+of this restriction, an approach similar to the CA Certificates Buildpack will
+be used where the credentials will be provided in a mounted volume through a
+service binding. Those credentials will then be used to change the global `git`
+config for that container which will persist for the entirety of the build
+which the [CNB specification](https://github.com/buildpacks/spec/blob/main/buildpack.md#requirements)
+states must all happen in the same container for security purposes.
```
dmikusa-pivotal, line 28: I read the referenced link, but I'm not seeing something that backs up this statement. It says:

>The lifecycle MUST be implemented so that the detection and build phases do not have access to OCI image store credentials used in the analysis and export phases. The lifecycle SHOULD be implemented so that each phase may run in a different container.

The first sentence there isn't relevant and I'm not sure I see the connection with the second sentence.

I'm not disagreeing. Do we have a better reference or explanation as to why this is OK?

text/utilities/0003-git-credential-configuration.md
```diff
@@ -0,0 +1,59 @@
...
+- `credentials`: The credentials file should have the following format to
+  conform with the [`git` credential structure](https://git-scm.com/docs/git-credential#IOFMT).
+```
+username=some-username
+password=some-password/token
+```
+
+For every service binding the following `git` command will be run:
+```shell
+git config --global credential.{{context}}.helper "!cat path/to/binding/credentials"
```
dmikusa-pivotal, line 47: Does the tiny builder have a shell in it? or does git run this command in a way that does not require a shell?

ForestEckhardt: The build image for tiny does have a shell. These settings will only be present for build, I did not see much purpose in configuring this to run during launch as well as it feels mostly pertinent to building. Are there applications that would require git credentials during run?

dmikusa-pivotal: That was my assumption too. That we'd only be providing this support for build time.

I guess it's possible that an application would need to use git, but I'd say it's up to the application to configure its credentials. A good 12-factor app shouldn't depend on that being pre-configured at the OS.

My $0.02 is that it's OK to only focus on build so long as we document that.

text/utilities/0003-git-credential-configuration.md
ForestEckhardt: I was looking for a line in the spec that states that all of the build processes happen in one container and this was the best that I could come up with tbh. The point I am trying to get across is that I am making this setting for the build container and not for the final image therefore the credentials setting will only exist during build.

dmikusa-pivotal: What about with untrusted builders? https://buildpacks.io/docs/tools/pack/concepts/trusted_builders/#why-use-a-trusted-builder

Do we care about that scenario?

dmikusa-pivotal: I read that differently. I read it as each buildpack and phase would be separate. So if you had buildpack A and buildpack B, that would be separate containers for each phase for each buildpack for a total of four separate containers.

samj1912: All the build phases of every detected buildpack are run in one container.

ForestEckhardt: @samj1912 do you know if that is spec'd out somewhere

dmikusa-pivotal: Ok, I'm all set here. That clarifies it for me. If there is a link cool, if not, that's Ok too. Thanks

fg-j:
In the [Slack conversation](https://paketobuildpacks.slack.com/archives/C011S6EL49L/p1642629255010000) about this RFC, there's some discussion of using SSH as a protocol instead of HTTPS. Maybe I'm missing something here – do we no longer care about enabling users to clone private packages using SSH as the protocol? Is that case implicitly handled by the implementation outlined in the RFC? Some clarity on this would be appreciated.

ForestEckhardt:
@fg-j In that conversation I had the word token overloaded in my head, I think that the only intention was to allow for authorization tokens and not ssh tokens. I also am not sure that I see a benefit to including ssh cloning protocol when HTTPS should cover all the access needs you would have.

sophiewigmore (approved):

ryanmoran (approved):

PR #142, opened 2022-01-31
Adds RFC for zero-config web server builds

ryanmoran:
## Summary

[Readable](https://github.com/paketo-buildpacks/rfcs/blob/web-server-zero-config/text/web-servers/0003-zero-config.md)

## Checklist


text/web-servers/0003-zero-config.md
```diff
@@ -0,0 +1,75 @@
...
+
+However, many users find themselves needing more than just the simplest
+configuration. They want support for SSI and HSTS, Basic Auth, and enforcement
+of HTTPS connections. This has created a slippery slope in the Staticfile
+buildpack where it is slowly recreating each of the configuration options
+already available in a proper web server configuration file with a more limited
+API. In most cases, users working with the Staticfile buildpack would likely be
+better off using the NGINX or HTTPD buildpacks directly. They'd have better
+control over their deployment configuration and the complete set of
+configuration options provided by that server.
```
dmikusa-pivotal, line 55: I understand the rationale here and have definitely seen users trying to escape the confines of staticfile when they should just supply their own config file. +1 for just pushing those use cases to supply their own config.

What is nice about staticfile buildpack is that there are a number of use cases that just work out-of-the-box, without needing to write config files. IMHO, the primary benefit to staticfile buildpack vs just using the nginx docker image directly is that the buildpack will generate config for you. It's not ever going to replace complicated configurations, but it's a clear advantage when you're in a hurry, like with a PoC or some short-lived project, for common and basic scenarios.

If the default config doesn't cover enough of these "it just works" use cases, there will be a gap, and users might reach for some other tool.

I'm having a little trouble determining what is turned on by the config snippet referenced above, but my $0.02 says we should cover the following scenarios:

- Hosting static files (duh - this is what it's all about)
- Push state for client-side Java apps. Not sure if this is being handled a different way, but this is the most common thing I've seen staticfile buildpack used to do.
- Hosting static files with basic auth. Basic auth isn't great, but I think there's a demand for "host my files, but don't make them public", which when paired with the next item is good enough for a number of cases.
- Force HTTPS. This is a very common requirement for larger companies. It also makes basic auth not totally broken, so it helps if you just want to throw up some files but don't want the whole world to see them.

I have seen these used very commonly with the staticfile buildpack.

text/web-servers/0003-zero-config.md
```diff
@@ -0,0 +1,75 @@
...
+
+This also doesn't take into account that the current implementation is
+completely built upon a concept that we've been trying to remove from all of
+the buildpacks in Paketo, the `buildpack.yml` file. We could convert the
+buildpack to read its configuration from a `Staticfile` file like what is done
+in the Cloud Foundry buildpack, but this doesn't do anything to resolve the
+"slippery slope" concern.
```
dmikusa-pivotal, line 62: People like having the configuration generated for them. It's convenient.

When you get to the point where you're trying to do something that's more advanced than what is exposed through staticfile config, people are reluctant to switch to generating their own config because they have to start from scratch.

Not sure if there is interest, but if there's a way to have the buildpack export its configuration, or possibly we just document where the configuration lives & how someone could copy it out of the container, I think that would help users transition and in turn help to reduce the desire for the more complicated features and the escape hatches (i.e. please include random config file for me).

fg-j: +1 to documenting how someone could look at the generated config living in the buildpack.

I'd also advocate documenting/creating sample apps that show how to achieve the use-cases you named above, Dan (push state for Java client side apps, basic auth, force HTTPS) with nginx and httpd configuration, rather than adding these as features to a no-config build.

dmikusa-pivotal: I really think we should try to push for some level of config generation beyond the standard static file host scenario. 

At the end of the day, it's what separates using buildpacks from `docker run --name some-nginx -v /some/content:/usr/share/nginx/html:ro -d nginx` (standard static file hosting) or `docker run --name my-custom-nginx-container -v /host/path/nginx.conf:/etc/nginx/nginx.conf:ro -d nginx` (providing a total custom nginx config). The same can be done with the httpd image.

IMHO, config generation is the value-add feature that separates buildpacks from using the stock nginx / httpd images for this use case. I agree that we need reasonable limits set on what we will/won't do for config generation and to keep the buildpacks manageable. I just feel like there's a middle ground here, where we can add some more configuration generation options, but keep them locked into specific, common use cases/scenarios.

text/web-servers/0003-zero-config.md
```diff
@@ -0,0 +1,75 @@
...
+location of the files to be served. However, in some cases, the buildpacks
+could choose to allow buildpack users to configure some commonly-specified
+options (e.g. `BP_WEB_SERVER_ROOT=/workspace/my-build-directory`). This RFC
+will not enumerate these options.
+
+## Rationale and Alternatives
+
+### Why not Staticfile?
+This proposal sounds a lot like what the existing Staticfile buildpack is
+already doing, allowing users to indicate that they would like a simple
```
ekcasey, line 35:
```suggestion
This proposal sounds a lot like what the [community Staticfile buildpack](https://github.com/paketo-community/staticfile) does, allowing users to indicate that they would like a simple
```

I assume this is the one we are talking about?

text/web-servers/0003-zero-config.md
```diff
@@ -0,0 +1,75 @@
...
+configuration file is not present in the application workspace. Specifically,
+when the `BP_WEB_SERVER` environment variable is set to either `nginx` or
+`httpd` then the respective buildpack will detect and ensure that a simple
+configuration file is included in the built image such that it can serve very
+basic static file applications.
+
+The configuration files will make assumptions about certain options, like the
+location of the files to be served. However, in some cases, the buildpacks
+could choose to allow buildpack users to configure some commonly-specified
+options (e.g. `BP_WEB_SERVER_ROOT=/workspace/my-build-directory`). This RFC
```
ekcasey, line 28: Do options like `BP_WEB_SERVER_ROOT` take precedence over the contents of a user provided config file, in the case that both are provided?

ekcasey, line 28: I think it might be worth wading into the details here. Are we going to take any opinions about the "default" settings? Which options are we willing to support this way vs which options will we require users to byo config file?

text/web-servers/0003-zero-config.md
```diff
@@ -0,0 +1,75 @@
...
+will not enumerate these options.
+
+## Rationale and Alternatives
+
+### Why not Staticfile?
```
ekcasey, line 33: Is this RFC implicitly asserting that we archive the staticfile buildpack after implementing the suggested enhancements to the web server buildpacks?

text/web-servers/0003-zero-config.md
fg-j: I'd probably be satisfied with some more specificity around what the process will be for deciding which features to initially include / how to add new configuration features. Just so project maintainers have something to cite when, inevitably, feature requests come in that verge into re-implementing nginx or httpd configuration features.

text/web-servers/0003-zero-config.md
ryanmoran: I don't know if we need to go as far as archive that buildpack, but I would certainly want to prevent it from being included into the web servers language family.

text/web-servers/0003-zero-config.md
ryanmoran: I've gone into some detail for how we could support these use cases. Take a look at the latest revision.

text/web-servers/0003-zero-config.md
ryanmoran: I've outlined in this RFC that we should document how a user my "eject" out of this zero-config setup into using the `nginx` or `httpd` buildpack. We would document how users could retrieve the generated configuration file from their built image.

text/web-servers/0003-zero-config.md
ryanmoran: I've detailed those options that it seems reasonable to support. I'd think anything outside of this core group starts to get a bit weedsy and becomes a situation where users really should be using their own configuration file.

fg-j (approved):

thitch97 (approved):

PR #147, opened 2022-02-11
Propose an image retention policy

dmikusa-pivotal:
## Summary

This PR proposes an image retention policy for images published by the Paketo Project.

[Readable](https://github.com/paketo-buildpacks/rfcs/blob/retention-policy/text/0000-image-retention-policy.md)


sophiewigmore:
@dmikusa-pivotal this seems like a good idea to me. It might be nice if we had a mechanism of surfacing the date an image will be deleted, if someone pulls an image that will be deleted in the next 60 days or something, just as an extra cautionary measure for users.

dmikusa-pivotal:
> @dmikusa-pivotal this seems like a good idea to me. It might be nice if we had a mechanism of surfacing the date an image will be deleted, if someone pulls an image that will be deleted in the next 60 days or something, just as an extra cautionary measure for users.
> 
> Edit: Just saw the bit about sending out a message on Slack. That seems reasonable. It would be even better if there was some kind of direct way to alert folks when they are actually using the image, but I can't say I know how that could be done

I'd be open to that, but I'm not sure how we'd go about it. There are some metrics on Docker Hub for our images, but I don't think GCR retains anything. It's also hard to make the buildpack detect this, as we can't change anything in the old, already publish buildpacks.

At the WG last week, someone had mentioned possibly adding functionality to the Buildpacks registry and buildpack tools which seems like an interesting possibility. If we could utilize an existing state like "PULLED" or create a state like "RETIRED" or "ARCHIVED" in the Registry, then it would be possible for tools like pack to provide more customized response messages if someone requests a buildpack to which the image has been deleted. At least if that tool is attempting to pull the image through the registry. If someone is using the image directly then the tool isn't going to be able to help.

We could possibly add something in buildpacks going forward to alert of this as well. The buildpack could retain its published date through metadata and print warnings as it approaches the end of life that Paketo defines through this RFC. Helpful going forward, but not something that would help with the large number of old buildpacks we're still maintaining images for at the moment.

text/0000-image-retention-policy.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Summary
+
+At the moment, the Paketo Project is maintaining images (buildpack, builder, stack, etc..) since the beginning of the project. This RFC proposes that we define an image retention policy that would allow the project to delete old images.
+
+## Motivation
+
+Multiple reasons:
+
+1. It is impractical from a cost and resources perspective for the project to maintain all images throughout the history of time.
```
ekcasey, line 11: I feel like we need some data to justify this assertion. Do we have a rough estimate of the costs to the project?

text/0000-image-retention-policy.md
```diff
@@ -0,0 +1,52 @@
...
+2. It is highly unlikely that images going back to the beginning of the project are still usable. Most buildpacks reference external dependencies and those dependencies may no longer exist.
+3. Older stacks, buildpacks and builder images very likely are full of vulnerable software and should no longer be used.
```
ekcasey, line 13: These images certainly contain vulnerabilities and shouldn't be used, but I don't see that as a strong argument for pulling them. I think having an immutable history of release artifacts is generally in the best interest of users and we should strive to maintain that unless it becomes impossible for cost or logistical reasons.

dmikusa-pivotal: There will still be release artifacts going back to the first release. The primary release artifact is the source code, which is tagged and exists forever on Github (at least as long as Github is around). 

There would also continue to be an archived history in the buildpacks registry. I talked with that team and if we use the "yank" functionality for old releases, that would retain them in the registry but indicate that it's no longer recommended that users consume them. I think this is exactly the statement that we want to make for old and vulnerable images.

In addition, if users need to retain those images for longer periods of time they can easily do so. It's a matter of `crane cp` (or insert your tool of choice) to move the required images onto the user's own storage. The benefit of having a retention policy in place is that users can plan around it. They'll know how the project will maintain images & they can make arrangements to store them elsewhere if that's a requirement they need.

>I think having an immutable history of release artifacts is generally in the best interest of users and we should strive to maintain that unless it becomes impossible for cost or logistical reasons.

I guess I would flip this around and say, why should we keep them? Do we really think anyone is going to use them? How frequently do we think someone is actually going to take advantage of us retaining them?

It's not just the overall cost, but the cost (and cost is not just financial here, I would include environmental cost as well)/benefit. Just because it's possible, should we accept a cost because one day, someone might come around and use one of the images? I think the policy can also help to fairly draw a line in terms of what is an acceptable cost/benefit ratio for the project.

If we pick a cut-off point where 99% (probably even higher, 99.9% or 99.99%) are not using these images, then what is the harm of reclaiming them? To me, the likelihood of someone needing Paketo Buildpacks Java v1.0.0 is very close to zero, unless someone decided to perform a software archaeology or research study project.

dmikusa-pivotal:
@ekcasey - After our discussions, I made a couple of changes.

1. Increase the time from 1 year to 2 years. I think 1 year was being too aggressive.

2. I changed some of the words to reflect a change in the thought process. I think the primary purpose of this RFC isn't to aggressively delete images and minimize costs. I think that would have a higher impact on our users. I think the purpose here is to put an upper bound on costs (financial & environmental). To provide a backstop for the project, but at the same time be something that 99% of our users don't even need to think about.

ryanmoran:
I'd like to see some clarification in this RFC that this retention policy is meant to apply to DockerHub only and that GCR images should already be treated as deprecated. We still have images in both locations, not because we intend to support them in both places, but because we've been slow to move off of GCR. This RFC was birthed out of a discussion of what should be done with the set of GCR images that predate our move to DockerHub and therefore do not exist in both places. I would be disappointed to see an outcome in this RFC that seems to perpetuate our use of GCR when we really want to stop using it as a registry.

dmikusa-pivotal:
> I'd like to see some clarification in this RFC that this retention policy is meant to apply to DockerHub only and that GCR images should already be treated as deprecated. We still have images in both locations, not because we intend to support them in both places, but because we've been slow to move off of GCR. This RFC was birthed out of a discussion of what should be done with the set of GCR images that predate our move to DockerHub and therefore do not exist in both places. I would be disappointed to see an outcome in this RFC that seems to perpetuate our use of GCR when we really want to stop using it as a registry.

Excellent feedback and I totally agree. I pushed some changes. Do those address your concerns?

text/0000-image-retention-policy.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Summary
+
+At the moment, the Paketo Project is maintaining images (buildpack, builder, stack, etc..) since the beginning of the project. This RFC proposes that we define an image retention policy that would allow the project to delete old images.
+
+## Motivation
+
+Multiple reasons:
+
+1. It is impractical from a cost and resources perspective for the project to maintain all images throughout the history of time. Time continues on infinitely, and eventually enough resources will accumulate to where this becomes a burden to the project.
```
ekcasey, line 11: Given this is equally true of dependencies built and provided by the Paketo project I think we should include deps in any retention policy we might adopt (cc @paketo-buildpacks/dependencies-maintainers)

dmikusa-pivotal:
FYI, I did a little investigation for the duration of two years to see how many images that would retain. For the Java buildpacks, the entire history of images is almost exactly two years as I write this. For the composite buildpacks, that is between 70 and 105 images. The couple of component buildpacks I looked at had smaller numbers around 50 for the same time period. This is roughly what I'd expect to see going forward as well, given we're on a weekly release cadence.

The count for run images (and probably build images although I didn't look) is much higher at almost 700 for the same duration. That makes sense as this includes tiny, base, and full.

IMHO, that seems very reasonable for retention. From a Java perspective, that would allow you to get non-LTS Java versions back to 13, which no one should be using anymore. Similarly, it would allow access to very old LTS Java 8 and 11 versions that should not be used anymore.

Any thoughts or objections to the proposed 2-year retention period?

-----

Commands to test:

```
gcrane ls gcr.io/paketo-buildpacks/run --json > data.json  # fetch info, requires gcrane
```

```
jq -r '.manifest | to_entries | .[] | [.key, ((.value.timeUploadedMs | tonumber) / 1000 | strftime("%Y-%m-%d %H:%M:%S")), (.value.imageSizeBytes | tonumber) / 1024 / 1024 ] | @csv' data.json > data.csv  # turn to CSV so you can load into a spreadsheet
```

ekcasey:
> Any thoughts or objections to the proposed 2-year retention period?

I am pretty well convinced at this point, but I would still like to see the deps we host on s3 included here. I propose we clean them up iff they have been orphaned (i.e. all buildpacks that use the deps have been removed).

ekcasey (approved):

ryanmoran (approved):

PR #154, opened 2022-02-23
Apache tomcat launch config

garethjevans:
## Summary
[Readable](https://github.com/sap-contributions/rfcs/blob/apache-tomcat-launch-config/text/java/0000-apache-tomcat-launch-config.md)

## Use Cases
Make it easier to configure lauchtime options in the Tomcat Buildpack.

## Checklist

### DIFF
--- a/text/java/0000-apache-tomcat-launch-config.md
+++ b/text/java/0000-apache-tomcat-launch-config.md
@@ -33,13 +33,20 @@ The resource configuration is static apart from the `url`, `username` and `passw
 
 ## Rationale and Alternatives
 
-TODO
-{{Discuss 2-3 different alternative solutions that were considered. This is required, even if it seems like a stretch. Then explain why this is the best choice out of available ones.}}
+It is possible to use `BPE_*` parameters with the [environment variables buildpack](https://github.com/paketo-buildpacks/environment-variables), these are written into the image at buildtime.
+
+These values could also be configured by updating `JAVA_TOOL_OPTIONS` variable, but this can be quite cumbersome and error prone if there are multiple values to configure, or if this env var already
+contains configuration.
 
 ## Implementation
 
 The implementation will add a helper to the tomcat buildpack that will look for launch variables named `BPL_TOMCAT_ENV_*`, e.g. `BPL_TOMCAT_ENV_POSTGRES_URL`, the helper 
 will take that variable, convert it to the system property `postgres.url` and append it to the environment variable `JAVA_TOOL_OPTIONS`.
+
+Pseudocode:
+```
+	systemProperty = envVar.removePrefix('BPL_TOMCAT_ENV_').toLowerCase().replaceAll('_', '.')
+```
 
 ## Prior Art
 
@@ -47,6 +54,7 @@ At present we've not found anything similar to this.
 
 ## Unresolved Questions and Bikeshedding
 
-* Sensitive variables may be exposed in the application logs.
+* Sensitive variables may be exposed in the application logs. We should ensure that `JAVA_TOOL_OPTIONS` is not logged, or is masked.
+* The (proposed) Apache Tomee buildpack would also benefit from supporting this.
 
 {{REMOVE THIS SECTION BEFORE RATIFICATION!}}
--- a/text/java/0000-apache-tomcat-launch-config.md
+++ b/text/java/0000-apache-tomcat-launch-config.md
@@ -1,0 +1,52 @@
+# Apache Tomcat launch configuration
+
+## Summary
+
+With the current implementation of the Apache Tomcat buildpack, its quite difficult to configure values at launch time that affect values in the `web.xml`, `server.xml` & `context.xml`. We 
+want to provide a way to make some features easily configurable and controlled at launch time rather than build time.
+
+## Motivation
+
+The Apache Tomcat buildpack provides a method of adding a configuration overlay using the build time `BP_TOMCAT_EXT_CONF_*` variables, this takes a pre-built archive
+and extracts this during build time. This method can be cumbersome if there are values that need to be changed on a per-deployment basis.  Apache Tomcat already supports
+the use of [system property replacements](https://tomcat.apache.org/tomcat-9.0-doc/config/systemprops.html#Property_replacements) within its configuration files.  The aim is to make
+the configuring of the system properties at launch time lot easier.
+ 
+## Detailed Explanation
+
+Given an example `context.xml`, which can be included in either an external archive, or inside the applications `META-INF` directory:
+
+```
+<Context>
+    <Resources allowLinking="true"/>
+
+    <Resource name="jdbc/Datasource" auth="Container"
+              type="javax.sql.DataSource" driverClassName="org.postgresql.Driver"
+              url="${postgres.url}" username="${postgres.username}"
+              password="${postgres.password}" maxTotal="25"
+              maxIdle="10" validationQuery="select 1"/>
+
+</Context>
+```
+
+The resource configuration is static apart from the `url`, `username` and `password` parameters, but we do not want these to be written into the image at build time.
+
+## Rationale and Alternatives
+
+TODO
+{{Discuss 2-3 different alternative solutions that were considered. This is required, even if it seems like a stretch. Then explain why this is the best choice out of available ones.}}
+
+## Implementation
+
+The implementation will add a helper to the tomcat buildpack that will look for launch variables named `BPL_TOMCAT_ENV_*`, e.g. `BPL_TOMCAT_ENV_POSTGRES_URL`, the helper 
+will take that variable, convert it to the system property `postgres.url` and append it to the environment variable `JAVA_TOOL_OPTIONS`.
+
+## Prior Art
+
+At present we've not found anything similar to this.
+
+## Unresolved Questions and Bikeshedding
+
+* Sensitive variables may be exposed in the application logs.
+
+{{REMOVE THIS SECTION BEFORE RATIFICATION!}}
--- a/text/java/0000-apache-tomcat-launch-config.md
+++ b/text/java/0000-apache-tomcat-launch-config.md
@@ -43,7 +43,7 @@ contains configuration.
 ### At build time:
 
 * Make a temp directory
-* Copy `$CATALINA_HOME/conf/catalina.properties` to the <temp dir>
+* Copy `$CATALINA_HOME/conf/catalina.properties` to the `<temp dir>`
 * Ensure that `<temp dir>/catalina.properties` is group writable
 * Pass the `<temp dir>/catalina.properties` location through to the launch time via an env variable `BPI_CATALINA_CONFIG`
 
--- a/text/java/0000-apache-tomcat-launch-config.md
+++ b/text/java/0000-apache-tomcat-launch-config.md
@@ -41,12 +41,21 @@ contains configuration.
 ## Implementation
 
 The implementation will add a helper to the tomcat buildpack that will look for launch variables named `BPL_TOMCAT_ENV_*`, e.g. `BPL_TOMCAT_ENV_POSTGRES_URL`, the helper 
-will take that variable, convert it to the system property `postgres.url` and append it to the environment variable `JAVA_TOOL_OPTIONS`.
+will take that variable, convert it to the system property `postgres.url` and append it to tomcat/tomee startup procedure.
 
 Pseudocode:
 ```
 	systemProperty = envVar.removePrefix('BPL_TOMCAT_ENV_').toLowerCase().replaceAll('_', '.')
 ```
+
+In order to avoid any potentially sensitive variables from being logged on startup via `JAVA_TOOL_OPTIONS`, we will write the new properties to end of the file `catalina.properties`.
+
+### Configuration from Bindings
+
+It would also be beneficial to support the setting of the system properties from a binding.  In this approach there are two possible options we would look at:
+
+1) Where the binding value is to be used directly.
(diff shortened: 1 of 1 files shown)


text/java/0000-apache-tomcat-launch-config.md
```diff
@@ -0,0 +1,60 @@
...
+	systemProperty = envVar.removePrefix('BPL_TOMCAT_ENV_').toLowerCase().replaceAll('_', '.')
+```
+
+## Prior Art
+
+At present we've not found anything similar to this.
+
+## Unresolved Questions and Bikeshedding
+
+* Sensitive variables may be exposed in the application logs. We should ensure that `JAVA_TOOL_OPTIONS` is not logged, or is masked.
```
dmikusa-pivotal, line 57: I don't know if you can prevent this, the JVM logs the contents when it starts up. If we turn the values into system properties through a helper, that's going to add them to JAVA_TOOL_OPTIONS so they would show up in the logs.

- It might be possible to suppress that logging from the JVM.
- It might be possible to instead of adding system properties via JAVA_TOOL_OPTIONS, add them another way. Tomcat has `conf/catalina.properties` which is essentially a list of system properties that get set when Tomcat starts up. We might be able to have the helper write them to that file. We should be able to locate `$CATALINA_BASE/conf/catalina.properties` and append to it. The main trick would be that we'd need to make that file group writable at build time. Because the app can be run as a different user, but will always have the same group.

dmikusa-pivotal (changes requested):
I'm definitely in favor of this. There's a gap at the moment because while you can install external config, that gets baked into the image and external config can potentially have secrets in it. As you pointed out, you can use system properties to parameterize the config but you have to pass those through JAVA_TOOL_OPTIONS which is logged :(   

There are a couple of comments included here to address, but in general I'm 👍

text/java/0000-apache-tomcat-launch-config.md
garethjevans: I think that appending the system properties to `$CATALINA_BASE/conf/catalina.properties` is probably the best way to go, we can do that at startup via a helper (as you mentioned need to make sure that is group writable)

text/java/0000-apache-tomcat-launch-config.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Implementation
+
+The implementation will add a helper to the tomcat buildpack that will look for launch variables named `BPL_TOMCAT_ENV_*`, e.g. `BPL_TOMCAT_ENV_POSTGRES_URL`, the helper 
+will take that variable, convert it to the system property `postgres.url` and append it to the environment variable `JAVA_TOOL_OPTIONS`.
+
+Pseudocode:
+```
+	systemProperty = envVar.removePrefix('BPL_TOMCAT_ENV_').toLowerCase().replaceAll('_', '.')
+```
```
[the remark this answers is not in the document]
dmikusa-pivotal: This is the one I was thinking of: https://github.com/paketo-buildpacks/azure-application-insights/blob/main/helper/properties.go

The implementation here would have to be a bit different. Taking the key/values and appending them to `conf/catalina.properties`. The `$CATALINA_BASE` env should be set though, so you can read that to get the path to the config file, open and append to the end of it. I don't think it would be different for a value or for a file path. Unless you're trying to take the contents of the file and put that into catalina.properties, not just the path. That's feasible, just a little more effort.

One thing we'll need to do in the buildpack at build time is changing the permissions on the conf/catalina.properties file to be group writable. It is possible to run buildpack images as a different user, and if that's done the helper won't be able to write to the `/layers` or `/workspace` directory. Alternatively, we could put that file in `/tmp/` and symlink it to the layer but either way, you can write to the file at runtime, and IMHO it seems cleaner to just change the permissions.

text/java/0000-apache-tomcat-launch-config.md
```diff
@@ -41,20 +41,28 @@ contains configuration.
...
 
+In order to avoid any potentially sensitive variables from being logged on startup via `JAVA_TOOL_OPTIONS`, we will write the new properties to end of the file `catalina.properties`.
+
+### Configuration from Bindings
+
+It would also be beneficial to support the setting of the system properties from a binding.  In this approach there are two possible options we would look at:
+
+1) Where the binding value is to be used directly.
+2) Where the binding value specifies a path that should be used.
+
```
dmikusa-pivotal, line 59:
```suggestion

The binding would be set by a user at runtime only.
```

text/java/0000-apache-tomcat-launch-config.md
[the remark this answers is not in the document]
garethjevans: @dmikusa-pivotal, got a bit further on this - the file that exists in the buildpack is in `$CATALINA_HOME/conf/` and this doesn't seem to read by anything.  The logic to load those properties [looks in $CATALINA_BASE] (https://github.com/apache/tomcat/blob/8e2aa5e45ce13388da62386e3cb1dbfa3b242b4b/java/org/apache/catalina/startup/CatalinaProperties.java) but that file doesn't exist.  So rather than making the file group writable, we would just need to copy it into `$CATALINA_BASE/conf` and modify it there?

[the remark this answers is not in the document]
dmikusa-pivotal: That sounds like a good approach to me. Just to make sure I understand, what you're proposing is:

At build time:
1. Make a temp directory
2. Copy `$CATALINA_HOME/conf/catalina.properties` to the temp dir
3. Write `$CATALINA_BASE/conf/catalina.config` which points to the temp dir

At launch time:
1. Read `$CATALINA_BASE/conf/catalina.config` to get the temp dir
2. Read from the binding and append values to `tmp-dir/catalina.properties`.

I'm 👍 with this.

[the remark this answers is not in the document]
garethjevans: Yeah, thats pretty much it, I'm not familiar with the `BPI_` env variables so hadn't considered those - but will look into it.  I'll get the RFC updated to reflect this.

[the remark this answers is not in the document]
dmikusa-pivotal: There's nothing special about `BPI_` just a convention we use on the Java buildpacks for something that's internal. It's typically when you need the build pack at build time to pass something to a helper/exec.d proc at launch time. The `BPI_` env's are not included in the docs or intended to be user-facing config.

text/java/0000-apache-tomcat-launch-config.md
```diff
@@ -0,0 +1,73 @@
...
+These values could also be configured by updating `JAVA_TOOL_OPTIONS` variable, but this can be quite cumbersome and error prone if there are multiple values to configure, or if this env var already
+contains configuration.
+
+## Implementation
+
+### At build time:
+
+* Make a temp directory
+* Copy `$CATALINA_HOME/conf/catalina.properties` to the <temp dir>
```
dmikusa-pivotal, line 46:
```suggestion
* Copy `$CATALINA_HOME/conf/catalina.properties` to the `<temp dir>`
```

dmikusa-pivotal (changes requested):

text/java/0000-apache-tomcat-launch-config.md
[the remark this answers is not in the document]
dmikusa-pivotal: I was looking at the most recent changes and it looks like even using a temp directory, we cannot get around the need to make the file group writable. The buildpack is going to write that file to `/tmp/whatever/catalina.properties` and the owner is going to be set to `cnb` user, but at runtime if the user is different it won't be able to update the file :( At least not without making it group writable.

It makes me think we should just simplify then and do something like:

At build time:

1. Copy `$CATALINA_HOME/conf/catalina.properties` to `$CATALINA_BASE/conf/catalina.properties`
2. Make `$CATALINA_BASE/conf/catalina.properties` group writable.

At launch time:

1. Read `$CATALINA_BASE`
2. Read bindings & append values to `$CATALINA_BASE/conf/catalina.properties`

It's not a huge difference, but we can get rid of the `BPI_` env variable and the extra setting `-Dcatalina.config` and it should function the same and have no difference in security.

What do you think?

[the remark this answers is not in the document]
garethjevans: That makes it much simpler, I've amended the proposal to include this

pivotal-david-osullivan (approved):

text/java/0000-apache-tomcat-launch-config.md
```diff
@@ -0,0 +1,70 @@
...
+
+### At build time:
+
+* Copy the included `resources/catalina.properties` into `$CATALINA_BASE/conf/catalina.properties`
+* Ensure that `$CATALINA_BASE/conf/catalina.properties` is group writable
+
+### At launch time:
+
+* Read the location of the `$CATALINA_BASE/conf/catalina.properties` using the `$CATALINA_BASE` env var
+* Read from all bindings and append values to `$CATALINA_BASE/conf/catalina.properties.`
```
ekcasey, line 51: We shouldn't be copying potentially sensitive data from a binding (usually implemented with a Secret in Kubernetes) to the root filesystem. The container filesystem has different security properties vs a mounted secret.


Instead can we have a property source that reads from the bindings? If we needed to we could contribute a property source similar to [spring cloud bindings](https://github.com/spring-cloud/spring-cloud-bindings) that reads properties from a binding.



But maybe there is an even easier way? Could we have a launch helper configure the location of an additional properties file? (forgive my hand waving here. not sure what all config options are available)

text/java/0000-apache-tomcat-launch-config.md
```diff
@@ -0,0 +1,70 @@
...
+* Read from all environment variables named `BPL_TOMCAT_ENV_*` and append values to `$CATALINA_BASE/conf/catalina.properties.`
```
ekcasey, line 52: Should we encourage the direct use of `org.apache.tomcat.util.digester.EnvironmentPropertySource` instead?

ekcasey, line 52: Should we encourage the direct use of `org.apache.tomcat.util.digester.EnvironmentPropertySource` instead?

ekcasey (changes requested):
I like the idea here but I think we need an implementation that doesn't copy potentially secret values to less secure locations.

text/java/0000-apache-tomcat-launch-config.md
dmikusa-pivotal: This might be the way to go. From the docs, it seems to imply this is what's recommended for K8s environments.

>org.apache.tomcat.util.digester.EnvironmentPropertySource can be used to replace parameters from the process' environment variables, e.g. injected ConfigMaps or Secret objects in container based systems like OpenShift or Kubernetes.

If we go that route, I'm not sure we necessarily need this RFC. It seems like you could reference env variables as place holders and map secrets directly, no buildpack intervention required.

@garethjevans thoughts? Does ^^ satisfy your use case?

garethjevans:
To summarise the three options available to us at the moment:

1. Modify the currently proposed option to allow it to store the updated `catalina.properties` in a temporary directory, the process will then be started with the `-Dcatalina.config=temp-location/catalina.properties`.  The use of a temp directory will ensure that sensitive values are not persisted across container restarts. The creation of the temporary file will be done by a helper at build time.
2. Configuring the digester to use `org.apache.tomcat.util.digester.EnvironmentPropertySource`, this will involve copying the `catalina.properties` from `$CATALINA_HOME/conf` to `$CATALINA_BASE/conf` and appending the config in the copied file. This can be done at build time rather than launch time.  For this to work, environment variables need to be named the same as the property to be replaced. e.g. https://github.com/apache/tomcat/blob/e99da3e6542f2189b2655ad0781d6a8ac816a26d/java/org/apache/tomcat/util/digester/EnvironmentPropertySource.java#L76
3. Creating a custom implementation of `org.apache.tomcat.util.IntrospectionUtils.PropertySource` that can load requested properties directly from a binding.  Also supporting environment variables & system properties would be of use too as not all properties would be sensitive and configured from a binding.

@dmikusa-pivotal / @ekcasey I think this summarises the three options.

text/java/0000-apache-tomcat-launch-config.md
dmikusa-pivotal: Consolidating conversation streams.

text/java/0000-apache-tomcat-launch-config.md
dmikusa-pivotal: The Java docs says you can do this through catalina.properties or just as a system property: https://tomcat.apache.org/tomcat-9.0-doc/api/org/apache/tomcat/util/digester/EnvironmentPropertySource.html

```
   -Dorg.apache.tomcat.util.digester.PROPERTY_SOURCE=org.apache.tomcat.util.digester.EnvironmentPropertySource
```

So we could have the build pack enable this, probably always (thoughts on that appreciated). It seems like a useful thing to have in a container world.

garethjevans:
After quite a bit of internal testing Option 2 does appear to fulfil most of the requirements we have at the moment, having said that, Option 3 would be much nicer approach, especially when it comes to reading from bindings.  I'm not sure that time is something that Tomcat would accept as a contribution as its specific to CNBs, but we could host a jar in the same way as the cloudfoundry logging support? e.g. https://github.com/paketo-buildpacks/apache-tomcat/blob/adeff5081eabfa4c4f534ac6f9738e4dc7a4046d/buildpack.toml#L120

garethjevans:
Yes, that sounds like a reasonable course of action.

dmikusa-pivotal:
OK, closing out.

PR #155, opened 2022-02-24
Moves Python Promotion RFC to Implemented section

ForestEckhardt:
@paketo-buildpacks/python-maintainers Does this feel accurate?


robdimsdale:
Yes, this looks accurate to me. All the existing python buildpacks have been promoted to the `paketo-buildpacks` organization, and new python buildpacks are being created in `paketo-buildpacks` rather than `paketo-community`.

ekcasey (approved):

PR #157, opened 2022-03-02
Adds RFC on build reproducibility

ryanmoran:
## Summary

[Readable](https://github.com/paketo-buildpacks/rfcs/blob/build-reproduction/text/0043-reproducible-builds.md)

## Checklist

### DIFF
--- a/text/0043-reproducible-builds.md
+++ b/text/0043-reproducible-builds.md
@@ -1,0 +1,55 @@
+# Expanding the Criteria for Reproducible Builds
+
+## Summary
+
+Given the same set of inputs and without leveraging caching, Paketo buildpacks
+should reproducibly build images.
+
+## Motivation
+
+The current set of Paketo buildpacks do a reasonably good job of helping
+developers to ship reproducible build artifacts. They reuse layers where
+possible, and don't modify files in ways that aren't reproducible. So, for
+users taking advantage of the caching behaviors of CNBs, the reproducible build
+story is largely solved.
+
+However, we can do better in some other cases. Specifically, given a codebase
+that is not changing, a set of buildpacks that are not changing, and stacks
+that are not changing, we should be able to reproduce the exact same image on
+subsequent runs without leveraging the caching features of CNB.
+
+Unfortunately, there are some cases where we've made choices that prevent this
+type of reproducible build. These cases should be removed or minimized to the
+greatest extent possible.
+
+## Detailed Explanation
+
+Many of the buildpacks contain behavior that results in non-reproducible images
+due to how they treat layer metadata. Specifically, they include a `built_at`
+metadata field that includes a timestamp for the instant that this layer was
+built.
+
+Including this type of field means that two subsequent runs of a build with
+identical inputs will produce different images.
+
+## Alternatives
+
+Alternatively, we could assert that Paketo buildpacks will not be making an
+effort to enable this type of reproducible build. However, this path doesn't
+seem reasonable. Supporting these types of reproducible builds is obviously
+possible given different choices that are not fundamental to the value
+proposition of buildpacks.
+
+## Implementation
+
+We should remove the `built_at` fields from all buildpack layers that include
+them. In our test suites that leverage this field to assert layer reuse, we
+should instead prefer to compare layer SHA values across rebuilds.
+
+We could also include new integration testing of this specific type of
+reproducible build scenario.
+
+## Unresolved Questions and Bikeshedding
+
+* How do we handle versioning of buildpacks that remove this layer metadata
+  field? Is this part of the public API?


robdimsdale:
@ryanmoran I'm in favor of this change. Do you know why the `built_at` field was added originally? If so, can we add a section to the RFC detailing why it was added and why it's ok to remove it now?

text/0043-reproducible-builds.md
```diff
@@ -0,0 +1,55 @@
...
+them. In our test suites that leverage this field to assert layer reuse, we
+should instead prefer to compare layer SHA values across rebuilds.
+
+We could also include new integration testing of this specific type of
+reproducible build scenario.
+
+## Unresolved Questions and Bikeshedding
+
+* How do we handle versioning of buildpacks that remove this layer metadata
+  field? Is this part of the public API?
```
dmikusa-pivotal, line 55: A couple of options come to mind:

- A feature flag to disable `built_at`. That way it remains, but someone can disable it and get reproducibility if they require it. You might be able to disable it by default if combined with some sort of deprecation period warning first.
- A deprecation period, followed by removal. We have some precedent for removing things following a deprecation period and not bumping a major version.
- Retain `build_at`, but always set it to 0. It would still exist, but the date value would always be the same. IMHO, that doesn't break the API cause the field still exists.

Some combination of these.

dmikusa-pivotal (approved):
I did a quick scan for `built_at` and we're not using that convention in the Java buildpacks. I don't think this will impact us.

That said, it sounds reasonable to me, and if there are things we'd need to change I'm up for that. Reproducibility is important.

ryanmoran:
> Do you know why the `built_at` field was added originally?

This field was added in a number of buildpacks to assist in testing layer reuse at the integration level. Outside of that usecase, I don't know of any consumers of this field. Its a field that is exposed in the image metadata, but not one that we've ever really considered or treated as a stable API.

text/0043-reproducible-builds.md
```diff
@@ -0,0 +1,55 @@
...
+effort to enable this type of reproducible build. However, this path doesn't
+seem reasonable. Supporting these types of reproducible builds is obviously
+possible given different choices that are not fundamental to the value
+proposition of buildpacks.
+
+## Implementation
+
+We should remove the `built_at` fields from all buildpack layers that include
+them. In our test suites that leverage this field to assert layer reuse, we
+should instead prefer to compare layer SHA values across rebuilds.
```
ekcasey, line 47: > In our test suites that leverage this field to assert layer reuse, we should instead prefer to compare layer SHA values across rebuilds.

Maybe I am missing something... but if the layer is truly reproducible won't the diffID be the same regardless of whether the buildpack rebuilt or reused the layer? This desirable and reproducible behavior means the the layer SHA is not useful for this specific type of assertion.

Instead we could do one of:
1. test layer reuse decisions only at the unit level
2. toggle build_at metadata with an flag that is off by default, use this for testing

I lean towards option one

text/0043-reproducible-builds.md
ekcasey: Given that this particular type of layer metadata is intended for use by the buildpack itself (on rebuilds) and not by humans or other software, it would be convenient if we could treat this as an implementation detail and not a public API. On the other hand, the world abound with instances of people building dependencies on undocumented APIs especially when they are easily discoverable (like the metadata image label).

IMHO I think we should **not** treat this like a public API and instead document very clearly that the layer metadata is subject to change and should not be depended on. A couple reasons:
1. It seems unlikely that anyone is actually depending on this
2. It will be harder for us in the long term if we make guarantees about the metadata schemas
3. It is in keeping with our historical practice (I can think of examples where we changed layer metadata in incompatible ways in the Java buildpacks without cutting a major)
4. We didn't note this as a type of breaking change in the [semver RFC](https://github.com/paketo-buildpacks/rfcs/blob/main/text/0029-semantic-versioning.md#buildpack-general). Not sure that was deliberate vs we just didn't think about it but we could go back and clearly indicate that this type of change is patch level.

text/0043-reproducible-builds.md
ryanmoran: You are absolutely right that this assertion alone is not enough to tell that a layer was not rebuilt. In fact, that is really the whole purpose of the RFC, to allow a real rebuild to produce a bit-for-bit duplicate of a previous build. This is why we had relied on the `built_at` field to provide a stronger assertion of the behavior of the buildpack from the outside.

We already have test coverage of layer reuse at the unit level. Additionally, after making these proposed changes, we would still have reasonable coverage through an integration test, albeit of a weaker form where we make two different assertions (this SHA-match assertion, and an assertion on log output indicating layer reuse) that allow us to say that this layer was reused rather than rebuilt.

ekcasey (approved):

PR #161, opened 2022-03-08
Adds RFC for disabling SBOM

ryanmoran:
## Summary

[Readable](https://github.com/paketo-buildpacks/rfcs/blob/disable-sbom/text/0043-disable-sbom.md)

## Checklist

### DIFF
--- a/text/0044-disable-sbom.md
+++ b/text/0044-disable-sbom.md
@@ -52,3 +52,12 @@ outputs.
 
 Additionally, when the variable is set, these buildpacks will include the image
 label `io.paketo.sbom.disabled` with a value of `true`.
+
+## Other Notes and Details
+
+In the Buildpacks Spec v0.7/ Platform Spec v0.8, the restorer phase of the
+lifecycle [restores SBOMs from previous
+builds](https://github.com/buildpacks/spec/blob/buildpack/v0.7/platform.md#outputs-2).
+If a layer with an SBOM is restored by the lifecycle and the buildpack chooses
+to reuse that layer, it does not need to explicitly remove that restored SBOM
+from the layer.
--- a/text/0044-disable-sbom.md
+++ b/text/0044-disable-sbom.md
@@ -1,0 +1,46 @@
+# Provide Global Mechanism to Disable SBOM Generation
+
+## Summary
+
+Users should have a mechanism to disable the generation of SBOM documents
+during the build process.
+
+## Motivation
+
+The generation of the SBOM documents for an image can be very time-intensive in
+some cases. We have seen a doubling of build times for Node.js applications.
+Additionally, some users are not interested in taking advantage of the SBOM
+features of the buildpacks.
+
+## Detailed Explanation
+
+If a user wishes to produce an image that does not contain SBOM documents, they
+can set the `BP_DISABLE_SBOM` environment variable to `true`. When unset, the
+value of this environment variable is assumed to be `false`.
+
+When `BP_DISABLE_SBOM` is set to `true`, buildpacks that allow SBOM to be
+omitted from their output should refrain from generating or attaching an SBOM
+in their outputs. This would apply to both new (Syft, CycloneDX, and SPDX
+formats) and old (label) SBOM outputs.
+
+## Rationale and Alternatives
+
+Alternatively, we could not provide a mechanism to disable SBOM generation.
+This does not seem like a great option as the build-time penalty introduced by
+SBOM generation can be substantial.
+
+We could also not specify anything at the organizational level and let each
+buildpack subteam decide how they would like to handle this for their
+ecosystem. However, with regards to this particular feature, it doesn't seem
+like there would be a big need to have differing implementations of this flag
+for each ecosystem. It would mean in some polyglot build scenarios that users
+might need to specify that they'd like to disable SBOM generation in more than
+one way. We and users would likely benefit more from a standardized feature
+flag that worked uniformly across all buildpacks.
+
+## Implementation
+
+Any buildpacks wishing to implement this feature will be required to honor the
+`BP_GENERATE_SBOM` environment variable. When set to `false` these buildpacks
+will neither generate their SBOM documents, nor attach them to any of their
+outputs.
--- a/text/0044-disable-sbom.md
+++ b/text/0044-disable-sbom.md
@@ -23,6 +23,11 @@ omitted from their output should refrain from generating or attaching an SBOM
 in their outputs. This would apply to both new (Syft, CycloneDX, and SPDX
 formats) and old (label) SBOM outputs.
 
+Additionally, when this variable is set to `true` a buildpack should set an
+image label of `io.paketo.sbom.disabled` to `true`. This label interface would
+allow downstream consumers of the image to understand that SBOM generation had
+been explicitly disabled.
+
 ## Rationale and Alternatives
 
 Alternatively, we could not provide a mechanism to disable SBOM generation.
@@ -44,3 +49,6 @@ Any buildpacks wishing to implement this feature will be required to honor the
 `BP_GENERATE_SBOM` environment variable. When set to `false` these buildpacks
 will neither generate their SBOM documents, nor attach them to any of their
 outputs.
+
+Additionally, when the variable is set, these buildpacks will include the image
+label `io.paketo.sbom.disabled` with a value of `true`.
--- a/text/0044-disable-sbom.md
+++ b/text/0044-disable-sbom.md
@@ -46,7 +46,7 @@ flag that worked uniformly across all buildpacks.
 ## Implementation
 
 Any buildpacks wishing to implement this feature will be required to honor the
-`BP_GENERATE_SBOM` environment variable. When set to `false` these buildpacks
+`BP_DISABLE_SBOM` environment variable. When set to `true` these buildpacks
 will neither generate their SBOM documents, nor attach them to any of their
 outputs.
 
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@
 * [0041: Use Direct Processes and exec.d](./text/0041-direct.md)
 * [0042: Adjust Builder Order](./text/0042-adjust-builder-order.md)
 * [0043: Expanding the Criteria for Reproducible Builds](./text/0043-reproducible-builds.md)
+* [0044: Provide Global Mechanism to Disable SBOM Generation](./text/0044-disable-sbom.md)
 
 ## Implemented RFCs
 


robdimsdale:
Is there prior art for the environment variable's name and default value? I have a slight preference for a variable whose default value is `false` as that aligns with the unset value. However, I have a greater preference for consistency with any existing environment variables that any buildpacks may have already chosen.

text/0043-disable-sbom.md
```diff
@@ -0,0 +1,50 @@
...
+like there would be a big need to have differing implementations of this flag
+for each ecosystem. It would mean in some polyglot build scenarios that users
+might need to specify that they'd like to disable SBOM generation in more than
+one way. We and users would likely benefit more from a standardized feature
+flag that worked uniformly across all buildpacks.
+
+## Implementation
+
+All buildpacks that generate an SBOM will be required to honor the
+`BP_GENERATE_SBOM` environment variable. When set to `false` these buildpacks
```
dmikusa-pivotal, line 44:
```suggestion
A buildpack that generates an SBOM and wishes to allow users to skip SBOM generation, should honor the `BP_GENERATE_SBOM` environment variable. When set to `false` these buildpacks
```

Thoughts about making this optional? I can see where it's useful for some buildpacks. If SBOM generation adds a perceptible amount of delay, allowing users to skip would be helpful. 

On the Java side, the scan is very fast. I'm not sure we'd want to implement this unless something changes and the scanning starts to impact build times.

Making this option would ensure the UX is consistent across language families if a language family decides to permit this.

text/0043-disable-sbom.md
```diff
@@ -0,0 +1,50 @@
...
+When `BP_GENERATE_SBOM` is set to `false`, none of the buildpacks executing
+during the build phase should generate or attach an SBOM in their outputs. This
+would apply to both new (Syft, CycloneDX, and SPDX formats) and old (label) SBOM
+outputs.
+
+## Rationale and Alternatives
+
+Alternatively, we could not provide an mechanism to disable SBOM generation.
+This does not seem like a great option as the build-time penalty introduced by
+SBOM generation can be substantial.
```
dmikusa-pivotal, line 30: Does anyone feel strongly enough that everyone should always have an SBOM included? SBOM is one of the marquee features of buildpacks. I assume that there are/will be tools built that consume the SBOM and do interesting stuff with it. Not having an SBOM could impact the success/adoption of those tools.

Do we feel like the performance hit here is related to the tool we are using? Is there a chance this tooling will get better/faster with some time/improvements?

Once we enable a feature like this (disabling SBOM), it would be hard to take it back.

I'm not sure exactly how I feel about the above statements, but I wanted to voice some contrary talking points to see what folks think.

dmikusa-pivotal:
> Is there prior art for the environment variable's name and default value? I have a slight preference for a variable whose default value is `false` as that aligns with the unset value. However, I have a greater preference for consistency with any existing environment variables that any buildpacks may have already chosen.

+1 to what Rob said here. I was just talking with David about some unrelated options we're adding in Java buildpacks and how we should be framing them with a default "false" value. I'm not sure we have this specified in an RFC, although we might (anyone know?), I thought that was the pattern that we are following across buildpacks? I think like 99% of Java config options are structured with default false. My plan was to try and get to 100% eventually.

I hate to throw out the first 🚲  but my preference would be for something like `BP_SBOM_DISABLED` but I'd be OK with `BP_DISABLE_BOM` which seems like the default false version of what's been proposed.

text/0043-disable-sbom.md
```diff
@@ -0,0 +1,50 @@
...
+## Detailed Explanation
+
+If a user wishes to produce an image that does not contain SBOM documents, they
+can set the `BP_GENERATE_SBOM` environment variable to `false`. When unset, the
+value of this environment variable is assumed to be `true`.
+
+When `BP_GENERATE_SBOM` is set to `false`, none of the buildpacks executing
+during the build phase should generate or attach an SBOM in their outputs. This
+would apply to both new (Syft, CycloneDX, and SPDX formats) and old (label) SBOM
+outputs.
```
dmikusa-pivotal, line 24: How do you feel about specifically adding a label when the user opts out that would indicate the user opted out? I think this could be helpful for a couple of reasons:

a.) it would be easy for tooling to check this and react accordingly. The tooling would know why an SBOM isn't there.
b.) similarly, if someone is troubleshooting and SBOM isn't present, then it's easy to differentiate between it being missing due to a technical problem or the user opting out.

text/0043-disable-sbom.md
dmikusa-pivotal: I also wonder about how this might impact adoption for large companies. If you allow disabling SBOM at the developer level, there isn't really a way to ensure that SBOM is there like if that is a company security/compliance requirement. 

If a dev can go in and turn off SBOM, they could bypass that requirement. Then additional work needs to be done to catch and force users to generate SBOM.

genevieve: I think optionality will ensure that you don't discourage casual developers from adopting buildpacks due to much longer build times due to sbom. At the enterprise level, lots of things can be disabled by a developer choosing to avoid security practices set in place by their leadership or security team.

dmikusa-pivotal: >I think optionality will ensure that you don't discourage casual developers from adopting buildpacks due to much longer build times due to sbom.

Right, but it's going to default to being enabled. It'll be slow out of the box. That means a casual developer needs to know to disable this to speed things up. I'm sure some will, but some won't too.

I definitely don't think we should just expect users to be Ok with it being slower, but perhaps there are other options. Improvements to the scanning tools? Different scanning tools? Performing a "lite" scan vs a full scan (possibly we could adjust the options to change the fidelity of information generated to speed up the scan)? Buildpack dev/test/debug mode, where we build the app image in a way that is only suitable for development but in that case, we don't scan (perhaps this would be a deterrent to using this as a way to bypass SBOM generation)?

I'm just concerned about making this optional. SBOM is an important feature of buildpacks. It would be nice if Paketo as a project can declare that all of our buildpacks generate SBOM. If we make it optional, that statement is diluted.

text/0043-disable-sbom.md
sophiewigmore: Personally I think it's more confusing to have some buildpacks in the project (cross-language) that allow it to be disabled and other that do not. If the default is to generate the SBOM, I don't see much downside to allowing users to disable it.

text/0043-disable-sbom.md
```diff
@@ -0,0 +1,50 @@
...
+Additionally, some users are not interested in taking advantage of the SBOM
+features of the buildpacks.
+
+## Detailed Explanation
+
+If a user wishes to produce an image that does not contain SBOM documents, they
+can set the `BP_GENERATE_SBOM` environment variable to `false`. When unset, the
+value of this environment variable is assumed to be `true`.
+
+When `BP_GENERATE_SBOM` is set to `false`, none of the buildpacks executing
```
fg-j, line 21: In the Buildpacks Spec v0.7/ Platform Spec v0.8, the `restorer` phase of the lifecycle [restores SBOMs from previous builds](https://github.com/buildpacks/spec/blob/buildpack/v0.7/platform.md#outputs-2). What's the expected behaviour of turning off the SBOM with this env var if there are preexisting SBOM files from earlier builds?

text/0043-disable-sbom.md
dmikusa-pivotal: @fg-j I think in my mind SBOM is a security-related component. Do we have prior art to allow users to disable security-related features?

fg-j: Do you think source code removal falls into that category, @dmikusa-pivotal? In that case, the ability to switch off source code removal feels like a comparable example.

dmikusa-pivotal: Yes, I could see arguments for that being security-related. 

I was thinking about this from the Java perspective. What I found that I think matches up is the ability to force a JDK to be installed in the runtime container. That would be prior art for something "security-related" (that you can opt-out of and it makes the images less secure). That's widely considered bad practice, cause it's including compiler tools in your runtime image. We allow that option though because there are some use cases that require it (they do on-the-fly compilation).

So it seems like there are a couple of things.

Ultimately what I think would negate this argument, but I'm not sure is possible, would be if a setting like this could be controlled at the builder level or across a platform. In that way, users can set it one way for local development but platform operators/companies can dictate a policy that cannot be side-stepped for production environments. I think there is going to be a desire for control over SBOM generation at the Ops level and that it'll be policy requirements for companies. If we can't deliver that, I worry it could impact adoption.

text/0043-disable-sbom.md
dmikusa-pivotal: I see value in the SBOM being consistently present on all images we generate. It may not matter at dev time, but for production environments, I think adoption in larger companies could be slowed if we allow this to be disabled. Assuming we go ahead with this RFC, making the implementation of it optional would still allow us to still make that guarantee for some language families, where it's feasible.

text/0043-disable-sbom.md
ryanmoran: I think that type of platform-wide override would need to be something that lived at the platform level. For instance, could there not be a `kpack` feature that enforced a specific environment variable was always set to a specific value, no matter what a developer might have provided. The interface for either of these features would still be an environment variable, but its up to the platform to enable this platform-wide override functionality.

text/0043-disable-sbom.md
ryanmoran: If the layer is reused, then we could say that the behavior is that previously generated SBOM outputs may survive into the new image. That is to say, it is not on buildpacks to clear out SBOMs from previous builds. Ultimately, in most cases, the SBOM should be cleared as soon as the layer needs to be rebuilt for a new package or runtime version. Or, the user could clear their cache before the build to ensure a fresh start.

text/0043-disable-sbom.md
dmikusa-pivotal: No disagreements there, a platform implementation makes sense. There is no present implementation though, so if we allow this to be disabled it could be a blocker for adoption with some users.

I'd be more OK with this if the RFC did not require language families to implement it. There is sufficient prior art that settings like this can be adoption at the project level and made optional, like the [RFC 32](https://github.com/paketo-buildpacks/rfcs/blob/main/text/0032-reloadable-process-types.md) and [RFC 37](https://github.com/paketo-buildpacks/rfcs/blob/main/text/0037-remote-debug.md). They define the interface, but don't require it to be implemented.

text/0043-disable-sbom.md
ryanmoran: I've changed the language here to specify that this is opt-in behavior for a buildpack.

text/0043-disable-sbom.md
ryanmoran: Addressed in https://github.com/paketo-buildpacks/rfcs/pull/161/commits/259df7f0550f32004d063b538b4ae9cc9e51652d

text/0043-disable-sbom.md
ryanmoran: Addressed in https://github.com/paketo-buildpacks/rfcs/pull/161/commits/c566e4612de413b399b20e9d4e2882011aab7883

dmikusa-pivotal (approved):

text/0044-disable-sbom.md
```diff
@@ -0,0 +1,63 @@
...
+like there would be a big need to have differing implementations of this flag
+for each ecosystem. It would mean in some polyglot build scenarios that users
+might need to specify that they'd like to disable SBOM generation in more than
+one way. We and users would likely benefit more from a standardized feature
+flag that worked uniformly across all buildpacks.
+
+## Implementation
+
+Any buildpacks wishing to implement this feature will be required to honor the
+`BP_GENERATE_SBOM` environment variable. When set to `false` these buildpacks
```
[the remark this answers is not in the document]
ryanmoran: I missed this. The proposed env var is `BP_DISABLE_SBOM`. The reasoning is that omission of the variable aligns with its meaning when set to `false`.

ekcasey (approved):

PR #169, opened 2022-04-04
Adds stack descriptor RFC

ryanmoran:
## Summary

[Readable](https://github.com/paketo-buildpacks/rfcs/blob/stack-descriptor/text/stacks/0003-stack-descriptor.md)

## Checklist


text/stacks/0003-stack-descriptor.md
```diff
@@ -0,0 +1,83 @@
...
+homepage = "<homepage uri>"      # optional
+maintainer = "<maintainer name>" # optional
+
+include-legacy-sbom = true # optional; default = false; enables https://github.com/paketo-buildpacks/rfcs/blob/main/text/stacks/0001-stack-package-metadata.md
+
+platforms = ["linux/amd64", "linux/arm64", "<other-platform>"] # optional; default = ["linux/amd64"]; choices defined in https://docs.docker.com/desktop/multi-arch/
+
+[distro]
+  name = "<os distro name>"       # optional
+  version = "os distro version>"  # optional
```
sophiewigmore, line 33: Can you explain why the `distro` isn't required?

sophiewigmore:
I can't see a reason not to do this. If anyone has one, I'd love to hear it. Any way we can make the stacks code more configurable and easier to use is a huge win in my eyes. I also feel like a well-defined configuration file demystifies what we'd expect for our stacks.

text/stacks/0003-stack-descriptor.md
ryanmoran: Its my understanding that the distro is not a required part of the stack: https://github.com/buildpacks/spec/blob/1c2ce7beb0d3c2a4950d795d31db49b11ceca0f7/platform.md?plain=1#L159-L198

text/stacks/0003-stack-descriptor.md
```diff
@@ -0,0 +1,83 @@
...
+
+[build]
+  tag = "<image tag>"                 # required
+  dockerfile = "<path to dockerfile>" # required
+  description = "<image description>" # optional
+
+  uid = 1001                   # required
```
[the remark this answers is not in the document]
sophiewigmore: ah ok.

text/stacks/0003-stack-descriptor.md
```diff
@@ -0,0 +1,85 @@
...
+[directions](https://buildpacks.io/docs/operator-guide/create-a-stack/)
+outlined in the buildpacks documentation.
+
+## Implementation
+
+The `create-stack` command will be refactored to support this new configuration
+file format and the features it allows. The codebase will then be moved into
+the `jam` codebase as a command within that tool. This aligns more with how the
+project maintains its tooling and allows us to keep tooling consolidated under
+the tooling subteam.
```
fg-j, line 81: I imagine that @paketo-buildpacks/stacks-maintainers will probably have a stake in this codebase; perhaps would be useful to have both teams as CODEOWNERS for the create-stack subcommand?

text/stacks/0003-stack-descriptor.md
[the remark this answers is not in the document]
robdimsdale: And even if they are the same, both fields are still required, yes?

text/stacks/0003-stack-descriptor.md
```diff
@@ -0,0 +1,85 @@
...
+
+Then the codebase will be merged into the `jam` CLI under a `create-stack`
+command. Once complete, a user should be able to build a stack with a given
+descriptor file, `stack.toml`, using the following command:
+
+`jam create-stack --config stack.toml`
+
+## Rationale and Alternatives
+
+We could abandon the `create-stack` codebase entirely and expect that anyone
```
robdimsdale, line 70: As someone with essentially no exposure to the stack creation, can you briefly describe the downside to the `Dockerfile` approach? I can infer/assume but I'd like to hear your thoughts.

text/stacks/0003-stack-descriptor.md
```diff
@@ -0,0 +1,85 @@
+# Stack Descriptor and Tooling
+
+## Summary
+
+The stack-building codebase,
+[`create-stack`](https://github.com/paketo-buildpacks/stacks/tree/main/create-stack)
+should be moved into the `jam` tooling CLI, and allow users to configure their
```
dmikusa-pivotal, line 7: Possibly a larger question, but it seems like there's a desire to centralize around `jam`, should we maybe consider calling it something else? Like the `paketo` cli, and make a push to have it be the tool for all things Paketo?

text/stacks/0003-stack-descriptor.md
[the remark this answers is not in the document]
ryanmoran: yes.

text/stacks/0003-stack-descriptor.md
ryanmoran: The biggest concern I have between the Dockerfile-only interface and what is proposed here is that we're co-mingling the parts that make up a "base image" (os-specific modifications to an image to make it a reasonable base to build on top of) and a "stack" (the parts of the image that are codified in the Buildpacks Platform Spec or in Paketo RFCs). The effort here is to try to separate those things so its a bit clearer where the line is.

Additionally, we have some features in the existing stacks that aren't part of the Buildpacks Platform Spec that are non-trivial to implement inline in a Dockerfile. Specifically, I am thinking of the code that generates our existing image label-based SBOM for run images. Ideally, I'd like to be able to move the existing stacks onto this new configuration format and tooling since we'll be maintaining them for the next year before they go out of support.

text/stacks/0003-stack-descriptor.md
ryanmoran: I would not be opposed to such a move. I'd be curious to hear what that might mean for the toolchain used by the Java buildpacks.

dmikusa-pivotal: Long term, I'd hope we can deprecate some things. It'll probably take a lot of work/time, but I think from a project standpoint it would be really nice to have one way to do this stuff. A good long-term goal to shoot for.

text/stacks/0003-stack-descriptor.md
sophiewigmore: @brayanhenao ?

brayanhenao: @sophiewigmore Yes, that's a good idea!

text/stacks/0003-stack-descriptor.md
```diff
@@ -0,0 +1,88 @@
...
+```toml
+id = "<stack id>"                # required
+homepage = "<homepage uri>"      # optional
+maintainer = "<maintainer name>" # optional
+
+platforms = ["linux/amd64", "linux/arm64", "<other-platform>"] # optional; default = ["linux/amd64"]; choices defined in https://docs.docker.com/desktop/multi-arch/
+
+[distro]
+  name = "<os distro name>"       # optional
+  version = "os distro version>"  # optional
```
ekcasey, line 31: Should we be using values (`ID` and `VERSION_ID`) in `/etc/os-release` instead of user-provided values?

Or we could make these optional and fail if they do not match `/etc/os-release`.

text/stacks/0003-stack-descriptor.md
```diff
@@ -0,0 +1,88 @@
...
+
+The `create-stack` codebase will be refactored to allow users to build stacks
+using a configuration file with the following format:
+
+```toml
+id = "<stack id>"                # required
+homepage = "<homepage uri>"      # optional
+maintainer = "<maintainer name>" # optional
+
+platforms = ["linux/amd64", "linux/arm64", "<other-platform>"] # optional; default = ["linux/amd64"]; choices defined in https://docs.docker.com/desktop/multi-arch/
```
ekcasey, line 27: Aren't the platform(s) are dictated by the platform(s) base image(s) from which we construct the stack?

Maybe we should:
1. require that the dockerfile accept the base image as a build arg
2. require that provide a reference to the base image in this file
3. If the reference is a single image, inherit the platform
4. If the reference is a multi-arch image, build multiple times times and create a multi-arch image from the result

text/stacks/0003-stack-descriptor.md
ryanmoran: It makes sense to just use what exists in `/etc/os-release`. We can probably omit these labels otherwise since the labels themselves are not required by the spec. I added some details about this in the `Stack Spec Requirements` section.

text/stacks/0003-stack-descriptor.md
ryanmoran: I outlined a bit how I expect the toolchain to handle multi-arch images. The outputs will basically always be formatted as an image index, but may only contain images for a single platform.

dmikusa-pivotal (approved):

sophiewigmore (approved):

brayanhenao (approved):

robdimsdale (approved):

ekcasey (approved):

PR #170, opened 2022-04-06
[PHP] Clarify Composer use, add dependency `composer-packages`

joshuatcasey:
## Summary
Clarify how Composer is used, add dependency `composer-packages` that is provided by `composer-install` and required by `php-start`.


text/php/0001-restructure.md
```diff
@@ -54,15 +54,28 @@ buildpacks[<sup>2</sup>](#note-2):
...
+    and makes it available on the `$PATH`
+
+* **composer-global**:
+  Detect: When `BP_COMPOSER_INSTALL_GLOBAL` is set
+  * provides: none
+  * requires:
+    * `php` at build
+    * `composer` at build
+  Build: Runs `composer global` to ensure that global PHP packages are available on the `$PATH`
+      for composer install scripts
```
sophiewigmore, line 70: What's the main rationale for having this split out from the `composer` buildpack?

joshuatcasey: `composer-global` installs packages that are only meant for composer scripts, not for the PHP application itself. 

Its output should be on the `$PATH` for `composer-install` in the build container but should not be on the `$PATH` in the launch container, so I felt splitting it out would accomplish that.

joshuatcasey:
> To me, this still doesn't address the question of whether the `composer-install` buildpack should know when the packages it's installing will be needed. How will this work in your idea?

It was called out here as an edit to PHP RFC 0001:

```markdown
* **composer-global**:
  Detect: When `BP_COMPOSER_INSTALL_GLOBAL` is set
  * provides: none

text/php/0001-restructure.md
sophiewigmore: Could we add it as a step in `composer-install`?

sophiewigmore: I think other install buildpacks do something similar

sophiewigmore:
@joshuatcasey I do not understand how the composer global buildpack helps `composer install` figure out when the packages installed by running `composer install` should be made available. From what you've laid out, I thought the composer global buildpack makes  other packages available for `composer` to use before running `composer install`. I am wondering about the packages that get installed by `composer install`.

sophiewigmore:
Maybe you could just add a comment to the Composer Install buildpack line in which you mention that it's layer will always be available at launch to support the current language family use cases, which may be subject to change in the future depending on the use cases we support.

text/php/0001-restructure.md
joshuatcasey: Ok, it can be done inside `composer-install`.

thitch97 (approved):

sophiewigmore (approved):

PR #171, opened 2022-04-06
Quarkus Buildpack

matejvasek:
## Summary
Lightweight Quarkus Buildpack.

[readable](https://github.com/matejvasek/rfcs/blob/quarkus-bp/text/java/0012-quarkus.md)

## Use Cases
Can be used to build JVM and native-image Quarkus application.

## Checklist

### DIFF
--- a/text/java/0012-quarkus.md
+++ b/text/java/0012-quarkus.md
@@ -1,0 +1,60 @@
+# Quarkus Buildpack
+
+## Summary
+Lightweight Quarkus Buildpack which would be part of `java` and `java-native-image` group.
+The buildpack would just set environment variables for the `maven`and `native-image` buildpacks.
+
+Proof of concept could be found at: [quarkus-bp][b]
+
+[b]: https://github.com/matejvasek/quarkus-bp
+
+## Motivation
+While building Quarkus application is already possible it's unnecessarily complicated.
+The developer have to set 2 to 4 environment variables which is not obvious.
+It should be as easy as just running: 
+```sh
+pack build my-img -B quay.io/mvasek/builder:base --env BP_NATIVE_IMAGE=1
+```
+instead of:
+```sh
+pack build my-img -B quay.io/mvasek/builder:base \
+  --env BP_NATIVE_IMAGE=1 \
+  --env "BP_MAVEN_BUILD_ARGUMENTS=package -DskipTests=true -Dmaven.javadoc.skip=true -Dquarkus.package.type=native-sources" \
+  --env "BP_MAVEN_BUILT_ARTIFACT=target/native-sources/*" \
+  --env "BP_NATIVE_IMAGE_BUILD_ARGUMENTS_FILE=native-image.args" \
+  --env "BP_NATIVE_IMAGE_BUILT_ARTIFACT=*-runner.jar"
+
+```
+
+## Detailed Explanation
+The Quarkus Buildpack will be moved into the paketo-buildpacks Github organization and added to the Java sub-team. Maintenance for this buildpack will be provided by the Java subteam.
+
+The following changes will be made:
+* [Quarkus Buildpack][b] moved to Paketo Buildpacks org
+* The Buildpack will be published to gcr.io/paketo-buildpacks/quarkus
+* The Buildpack will be added to the `java` and `java-native-image` buildpacks groups just before the `maven` buildpack.
+
+[b]: https://github.com/matejvasek/quarkus-bp
+
+## Rationale and Alternatives
+Alternative approach could be to create much less lightweight buildpacks
+which would do all steps carried out by the `maven`, `executable-jar` and `native-image` buildpacks
+in on single Quarkus Buildpack.
+
+This would allow better control over build:
+* Better enablement for various package types (i.e. other than `fast-jar`).
+* Better control over layering: separating directories according probability of their modification (application code vs. library code).
+
+However, it would be much more complicate to implement, and it would lead to duplicity.
+
+## Implementation
+
+This buildpack will be adopted by the Java Buildpacks team and transferred to the `paketo` GitHub organization.
+
+## Prior Art
+
+N/A
+
+## Unresolved Questions and Bikeshedding
+
+N/A
--- a/text/java/0012-quarkus.md
+++ b/text/java/0012-quarkus.md
@@ -1,7 +1,7 @@
 # Quarkus Buildpack
 
 ## Summary
-Provide a Quarkus Buildpack which will be part of the order groups for `paketo-buildpacks/java` and `paketo-buildpacks/java-native-image`.
+Provide a Quarkus Buildpack which will be part of the order groups for `paketo-buildpacks/java`, `paketo-buildpacks/java-azure`, and `paketo-buildpacks/java-native-image`.
 The primary purpose of this buildpack is to set environment variables for the `paketo-buildpacks/maven` and `paketo-buildpacks/native-image` buildpacks. Building a Quarkus application is presently possible but requires setting multiple environment variables to properly configure the buildpacks. This proposed buildpack will enable users to build Quarkus applications without additional configuration.
 
 Proof of concept could be found at: [quarkus-bp][b]
@@ -27,12 +27,12 @@ pack build my-img -B quay.io/mvasek/builder:base \
 ```
 
 ## Detailed Explanation
-The Quarkus Buildpack will be moved into the paketo-buildpacks Github organization and added to the Java sub-team. Maintenance for this buildpack will be provided by the Java subteam.
+The [Quarkus Buildpack][b] will be moved into the paketo-buildpacks Github organization and added to the Java sub-team. Maintenance for this buildpack will be provided by the Java subteam.
 
 The following changes will be made:
 * [Quarkus Buildpack][b] moved to Paketo Buildpacks org
-* The Buildpack will be published to gcr.io/paketo-buildpacks/quarkus
-* The Buildpack will be added to the `paketo-buildpacks/java` and `paketo-buildpacks/java-native-image` buildpacks groups just before the `paketo-buildpacks/maven` buildpack.
+* The Buildpack will be published to `gcr.io/paketo-buildpacks/quarkus`
+* The Buildpack will be added to the `paketo-buildpacks/java`, `paketo-buildpacks/java-azure`, and `paketo-buildpacks/java-native-image` buildpacks groups just before the `paketo-buildpacks/maven` buildpack.
 
 [b]: https://github.com/matejvasek/quarkus-bp
 
@@ -45,14 +45,14 @@ This would allow better control over build:
 * Better enablement for various package types (i.e. other than `fast-jar`).
 * ~~Better control over layering: separating directories according probability of their modification (application code vs. library code).~~
 
-However, it would be much more complicate to implement, and it would lead to duplicity.
+However, it would be much more complicated to implement, and it would lead to duplicity.
 
 Another approach could be to have two Quarkus buildpacks: one pre maven/gradle build and one post build.
 With these we might be able to handle various build types (fast/legacy/uber jar) ~~and also handle layering~~.
 
 ## Implementation
 
-This buildpack will be adopted by the Java Buildpacks team and transferred to the `paketo` GitHub organization.
+This buildpack will be adopted by the Java Buildpacks team and transferred to the `paketo` GitHub organization as described in the [Detailed Explanation](#detailed-explanation) section above.
 
 ## Prior Art
 
--- a/text/java/0012-quarkus.md
+++ b/text/java/0012-quarkus.md
@@ -2,7 +2,9 @@
 
 ## Summary
 Provide a Quarkus Buildpack which will be part of the order groups for `paketo-buildpacks/java`, `paketo-buildpacks/java-azure`, and `paketo-buildpacks/java-native-image`.
-The primary purpose of this buildpack is to set environment variables for the `paketo-buildpacks/maven` and `paketo-buildpacks/native-image` buildpacks. Building a Quarkus application is presently possible but requires setting multiple environment variables to properly configure the buildpacks. This proposed buildpack will enable users to build Quarkus applications without additional configuration.
+The primary purpose of this buildpack is to set environment variables for the `paketo-buildpacks/maven` and `paketo-buildpacks/native-image` buildpacks. 
+
+Building a Quarkus application is presently possible but requires setting multiple environment variables to properly configure the buildpacks. This proposed buildpack will enable users to build Quarkus applications without additional configuration.
 
 Proof of concept could be found at: [quarkus-bp][b]
 
@@ -11,6 +13,7 @@ Proof of concept could be found at: [quarkus-bp][b]
 ## Motivation
 While building Quarkus application is already possible it's unnecessarily complicated.
 The developer have to set 2 to 4 environment variables which is not obvious.
+
 It should be as easy as just running: 
 ```sh
 pack build my-img -B quay.io/mvasek/builder:base --env BP_NATIVE_IMAGE=1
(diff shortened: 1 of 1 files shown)


text/java/0012-quarkus.md
```diff
@@ -0,0 +1,60 @@
+# Quarkus Buildpack
+
+## Summary
+Lightweight Quarkus Buildpack which would be part of `java` and `java-native-image` group.
+The buildpack would just set environment variables for the `maven`and `native-image` buildpacks.
```
dmikusa-pivotal, line 5:
```suggestion
The primary purpose of this buildpack is to environment variables for the `paketo-buildpacks/maven` and `paketo-buildpacks/native-image` buildpacks. Building a Quarkus application is presently possible but requires setting multiple environment variables to properly configure the buildpacks. This proposed buildpack will enable users to build Quarkus applications without additional configuration.
```

text/java/0012-quarkus.md
```diff
@@ -0,0 +1,60 @@
...
+
+```
+
+## Detailed Explanation
+The Quarkus Buildpack will be moved into the paketo-buildpacks Github organization and added to the Java sub-team. Maintenance for this buildpack will be provided by the Java subteam.
+
+The following changes will be made:
+* [Quarkus Buildpack][b] moved to Paketo Buildpacks org
+* The Buildpack will be published to gcr.io/paketo-buildpacks/quarkus
+* The Buildpack will be added to the `java` and `java-native-image` buildpacks groups just before the `maven` buildpack.
```
dmikusa-pivotal, line 35:
```suggestion
* The Buildpack will be added to the `paketo-buildpacks/java` and `paketo-buildpacks/java-native-image` buildpacks groups just before the `paketo-buildpacks/maven` buildpack.
```

text/java/0012-quarkus.md
```diff
@@ -0,0 +1,60 @@
...
+## Rationale and Alternatives
+Alternative approach could be to create much less lightweight buildpacks
+which would do all steps carried out by the `maven`, `executable-jar` and `native-image` buildpacks
+in on single Quarkus Buildpack.
+
+This would allow better control over build:
+* Better enablement for various package types (i.e. other than `fast-jar`).
+* Better control over layering: separating directories according probability of their modification (application code vs. library code).
+
+However, it would be much more complicate to implement, and it would lead to duplicity.
```
dmikusa-pivotal, line 48: I'd be curious to get your thoughts on a few things here:

1. How important do you see these two items? a.) being able to build different package types and b.) better control over layering. They seem at least moderately important, IMHO.

2. What are the challenges with implementing them with the proposed architecture here? For example, we do layer slicing in the Spring Boot buildpack but it's not controlling the build phase. That just happens after the build.
    
    If there is missing functionality, it could also be possible to extend existing buildpacks to add support for ways for a separate Quarkus buildpack to influence these aspects in the existing buildpacks.

3. Does Quarkus always use Maven for builds? or does it use Gradle too? I'm just trying to understand the amount of functionality that would need to be combined into one buildpack to make all this work in a more monolithic buildpack? It would obviously be less desirable to have a monolithic buildpack, but we do have some of this functionality broken into libraries already, like libbs, and we can technically import code from other buildpacks so you could possibly import existing aspects from a buildpack to combine them without as much code duplication. Anyway, I don't want to 100% rule this out. We should weigh the benefits versus the implementation.

dmikusa-pivotal (commented):
Some initial comments.

text/java/0012-quarkus.md
matejvasek: >Does Quarkus always use Maven for builds? or does it use Gradle too?

It can also use `gradle` so support for is should be added in near future.

matejvasek: >we do layer slicing in the Spring Boot buildpack but it's not controlling the build phase

I think that springboot expose some config files describing layering to be used. AFAIK Quarkus is not producing such a file.

matejvasek: >being able to build different package types 

In order to set envvars for the `maven` buildpack the quarkus buildpack has to run before it. But to detect what build type was used is has to run after it. I would have to have multiple Quarkus buildpacks to handle this: `quarkus-pre-maven` `quarkus-post-maven`.

matejvasek: When spring boot creates layers, does it remove `/workspace/`?

matejvasek: >How does Quarkus normally define its layers? Is that done through the Dockerfiles it uses or some other means?

Scaffold-ed Quarkus project usually contains Dockerfiles with `COPY` in efficient order.

matejvasek: One of the things that bothers me is: the `maven` buildpacks does two things:
* builds the app
* copies artifacts somewhere

In order to copy appropriate artifacts I have to set envvars for the `maven` buildpack therefore I have to run before it.
But in order to detect what build type out of the three (legacy-jar, fast-jar, uber-jar) I have to run after the `maven` buildpack. This make me to enforce only one of the package types.

BarDweller: The maven buildpack _also_ downloads maven.. it's really doing quite a lot.. I'd love to see the maven buildpack split off the provisioning of maven, so then I can provide it, and it can use it. (Admittedly there are some issues here related to maven versions, and supported flags etc). 

I guess if a separate buildpack provisioned maven, you could use the provisioned maven to query the project to ask which deployment type was configured, and then set the appropriate env vars ? then the build-with-maven buildpack could do the actual build with that knowledge already in place. 

(I have a related problem, but it's not so easily solved.. needing to determine which level of java is being asked for by the maven project.. I can hack an approximation by grepping the pom.xml, but if the version is configured in a parent pom that's being read from a maven repo, then it gets harder, and the easiest way is to invoke `mvn` to ask the project for the value of the resolved property.. I'm playing with trying to do this from a natively compiled java app, to avoid the jvm dependency when figuring out which jvm you need).

BarDweller: > How does Quarkus normally define its layers? Is that done through the Dockerfiles it uses or some other means?

I can make this a circular problem ;p one way quarkus can build a container, is by using buildpacks ;p (it also supports various other container-image extensions that each try to do the right thing for layers.. https://quarkus.io/guides/container-image ).. as well as Dockerfiles etc.

matejvasek: @maxandersen I think it won't be necessary I should be able to determine layering based on build type (uber/legacy/fast).

dmikusa-pivotal: >The maven buildpack also downloads maven.. it's really doing quite a lot.. I'd love to see the maven buildpack split off the provisioning of maven, so then I can provide it, and it can use it. (Admittedly there are some issues here related to maven versions, and supported flags etc).

@BarDweller we provide support for maven wrapper which is the suggested way of controlling the specific Maven version that gets used. Or are you saying you want the opposite? Or the buildpack to provide Maven using it's current mechanisms but not do any build?

dmikusa-pivotal: >(I have a related problem, but it's not so easily solved.. needing to determine which level of java is being asked for by the maven project.. I can hack an approximation by grepping the pom.xml, but if the version is configured in a parent pom that's being read from a maven repo, then it gets harder, and the easiest way is to invoke mvn to ask the project for the value of the resolved property.. I'm playing with trying to do this from a natively compiled java app, to avoid the jvm dependency when figuring out which jvm you need).

That's interesting. We could talk about trying to support something like this. There's a thread on Paketo Slack. That would probably be a better place as it'll probably be a more lengthy conversation. Are you on the Paketo Slack?

dmikusa-pivotal: >I think it won't be necessary I should be able to determine layering based on build type (uber/legacy/fast).

Ok, so this circles back to the conversation on Slack about possibly splitting up the build tools buildpacks. It'll be a big project though and something that should have it's own RFC. Bringing things back to this RFC, should we proceed here? or hold this off until we make sufficient changes to the build tools buildpacks?

It'd be nice to get a Quarkus buildpack that's working, but I don't want to release something that we can't migrate into the state required to support slicing and all of the package types if it'll break users.

matejvasek: What is `build tools buildpack` ?

dmikusa-pivotal: >What is build tools buildpack ?

Sorry, I sometimes use that to refer to the full set of all the build tools related buildpacks we support (Gradle, Maven, SBT, etc...). Usually conversations focus on Maven & Gradle, but there are others we have to consider in changes like this too.

matejvasek:
Actually it seems slicing is not an issue. I can forward declare slices even thought the directories/files are yet to be created by some other buildpack.

So the only inconvenience is that the buildpack has to force some specific packaging type (e.g. `fast-jar`) which renders value defined in `application.properties` or in the `QUARKUS_PACKAGE_TYPE` envvar unused.
I think we could live with that.

matejvasek:
>So the only inconvenience is that the buildpack has to force some specific packaging type (e.g. fast-jar) which renders value defined in application.properties or in the QUARKUS_PACKAGE_TYPE envvar unused.
I think we could live with that.

I don't know, maybe the `quarkus.package.type` property could be somehow evaluated even without running `mvn` build.

matejvasek:
@maxandersen  The content of `target/quarkus-app/quarkus:
```
generated-bytecode.jar
quarkus-application.dat
transformed-bytecode.jar
```
seems to be different on each build.
Is there a way to make these reproducible?

maxandersen:
> @maxandersen is there any possible issue with forcing `fast-jar` and disregarding package type declared in `application.properties`?

Probably not - except build and deploy features of quarkus will listen to the property - but if buildpack does the container building and therefore have to force a behavior I don't think it will be a problem.

text/java/0012-quarkus.md
matejvasek: It turned out I can setup layering even with current solution.
The package type is not too important.

text/java/0012-quarkus.md
```diff
@@ -0,0 +1,63 @@
+# Quarkus Buildpack
+
+## Summary
+Provide a Quarkus Buildpack which will be part of the order groups for `paketo-buildpacks/java` and `paketo-buildpacks/java-native-image`.
```
dmikusa-pivotal, line 4:
```suggestion
Provide a Quarkus Buildpack which will be part of the order groups for `paketo-buildpacks/java`, `paketo-buildpacks/java-azure`, and `paketo-buildpacks/java-native-image`.
```

text/java/0012-quarkus.md
```diff
@@ -0,0 +1,63 @@
...
+  --env BP_NATIVE_IMAGE=1 \
+  --env "BP_MAVEN_BUILD_ARGUMENTS=package -DskipTests=true -Dmaven.javadoc.skip=true -Dquarkus.package.type=native-sources" \
+  --env "BP_MAVEN_BUILT_ARTIFACT=target/native-sources/*" \
+  --env "BP_NATIVE_IMAGE_BUILD_ARGUMENTS_FILE=native-image.args" \
+  --env "BP_NATIVE_IMAGE_BUILT_ARTIFACT=*-runner.jar"
+
+```
+
+## Detailed Explanation
+The Quarkus Buildpack will be moved into the paketo-buildpacks Github organization and added to the Java sub-team. Maintenance for this buildpack will be provided by the Java subteam.
```
dmikusa-pivotal, line 30:
```suggestion
The [Quarkus Buildpack][b] will be moved into the paketo-buildpacks Github organization and added to the Java sub-team. Maintenance for this buildpack will be provided by the Java subteam.
```

text/java/0012-quarkus.md
```diff
@@ -0,0 +1,63 @@
...
+
+The following changes will be made:
+* [Quarkus Buildpack][b] moved to Paketo Buildpacks org
+* The Buildpack will be published to gcr.io/paketo-buildpacks/quarkus
```
dmikusa-pivotal, line 34:
```suggestion
* The Buildpack will be published to `gcr.io/paketo-buildpacks/quarkus`
```

text/java/0012-quarkus.md
```diff
@@ -0,0 +1,63 @@
...
+* The Buildpack will be added to the `paketo-buildpacks/java` and `paketo-buildpacks/java-native-image` buildpacks groups just before the `paketo-buildpacks/maven` buildpack.
```
dmikusa-pivotal, line 35:
```suggestion
* The Buildpack will be added to the `paketo-buildpacks/java`, `paketo-buildpacks/java-azure`, and `paketo-buildpacks/java-native-image` buildpacks groups just before the `paketo-buildpacks/maven` buildpack.
```

text/java/0012-quarkus.md
```diff
@@ -0,0 +1,63 @@
...
+[b]: https://github.com/matejvasek/quarkus-bp
+
+## Rationale and Alternatives
+Alternative approach could be to create much less lightweight buildpacks
+which would do all steps carried out by the `maven`, `executable-jar` and `native-image` buildpacks
+in on single Quarkus Buildpack.
+
+This would allow better control over build:
+* Better enablement for various package types (i.e. other than `fast-jar`).
+* ~~Better control over layering: separating directories according probability of their modification (application code vs. library code).~~
```
dmikusa-pivotal, line 46: You have some parts ~~struck out~~, if these are no longer items of concern, I would suggest we just remove them.

text/java/0012-quarkus.md
```diff
@@ -0,0 +1,63 @@
...
+
+However, it would be much more complicate to implement, and it would lead to duplicity.
```
dmikusa-pivotal, line 48:
```suggestion
However, it would be much more complicated to implement, and it would lead to duplicity.
```

text/java/0012-quarkus.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+## Detailed Explanation
+The Quarkus Buildpack will be moved into the paketo-buildpacks Github organization and added to the Java sub-team. Maintenance for this buildpack will be provided by the Java subteam.
+
+The following changes will be made:
+* [Quarkus Buildpack][b] moved to Paketo Buildpacks org
+* The Buildpack will be published to gcr.io/paketo-buildpacks/quarkus
+* The Buildpack will be added to the `paketo-buildpacks/java` and `paketo-buildpacks/java-native-image` buildpacks groups just before the `paketo-buildpacks/maven` buildpack.
+
```
dmikusa-pivotal, line 36: I would like to see what the proposed order group is for `paketo-buildpacks/java` (the others should be basically the same).

text/java/0012-quarkus.md
```diff
@@ -0,0 +1,63 @@
...
+* ~~Better control over layering: separating directories according probability of their modification (application code vs. library code).~~
+
+However, it would be much more complicate to implement, and it would lead to duplicity.
+
+Another approach could be to have two Quarkus buildpacks: one pre maven/gradle build and one post build.
+With these we might be able to handle various build types (fast/legacy/uber jar) ~~and also handle layering~~.
+
+## Implementation
+
+This buildpack will be adopted by the Java Buildpacks team and transferred to the `paketo` GitHub organization.
```
dmikusa-pivotal, line 55:
```suggestion
This buildpack will be adopted by the Java Buildpacks team and transferred to the `paketo` GitHub organization as described in the [Detailed Explanation](#detailed-explanation) section above.
```

dmikusa-pivotal (changes requested):
I think this looks good, just a few minor things to clean up. If these look Ok to you, just merge them and I'll approve. 

I took a look at your buildpack too and that looks good. Thanks for putting this together.

text/java/0012-quarkus.md
```diff
@@ -0,0 +1,63 @@
...
+pack build my-img -B quay.io/mvasek/builder:base \
+  --env BP_NATIVE_IMAGE=1 \
+  --env "BP_MAVEN_BUILD_ARGUMENTS=package -DskipTests=true -Dmaven.javadoc.skip=true -Dquarkus.package.type=native-sources" \
+  --env "BP_MAVEN_BUILT_ARTIFACT=target/native-sources/*" \
+  --env "BP_NATIVE_IMAGE_BUILD_ARGUMENTS_FILE=native-image.args" \
+  --env "BP_NATIVE_IMAGE_BUILT_ARTIFACT=*-runner.jar"
+
+```
+
+## Detailed Explanation
```
ekcasey, line 29: I think it would be good to include a description of the detection criteria here.

After reading the docs from the POC I would like to discuss:
1. could/should we also handle the case where a quarkus fast-jar is provided to the build, rather than building from source?
2. bikeshedding around what to do if some of the env vars are set by the user. If they are set to desired values already maybe this buildpack should still detect?

matejvasek: >a mechanism to disable the buildpack (an escape hatch in the case of false detection)

maybe we could add some envvar e.g. `QUARKU_BP_DISABLED` ?

matejvasek: >could/should we also handle the case where a quarkus fast-jar is provided to the build, rather than building from source?

not sure what you mean

text/java/0012-quarkus.md
```diff
@@ -0,0 +1,63 @@
...
+It should be as easy as just running: 
+```sh
+pack build my-img -B quay.io/mvasek/builder:base --env BP_NATIVE_IMAGE=1
+```
+instead of:
+```sh
+pack build my-img -B quay.io/mvasek/builder:base \
+  --env BP_NATIVE_IMAGE=1 \
+  --env "BP_MAVEN_BUILD_ARGUMENTS=package -DskipTests=true -Dmaven.javadoc.skip=true -Dquarkus.package.type=native-sources" \
+  --env "BP_MAVEN_BUILT_ARTIFACT=target/native-sources/*" \
```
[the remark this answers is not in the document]
matejvasek: So far this BP handles maven only. Definitely would be good to support gradle too. It could be first feature.

text/java/0012-quarkus.md
matejvasek: > > a mechanism to disable the buildpack (an escape hatch in the case of false detection)
> 
> maybe we could add some envvar e.g. `QUARKU_BP_DISABLED` ?

How do other BPs handle this?

text/java/0012-quarkus.md
```diff
@@ -0,0 +1,63 @@
...
+## Rationale and Alternatives
+Alternative approach could be to create much less lightweight buildpacks
+which would do all steps carried out by the `maven`, `executable-jar` and `native-image` buildpacks
+in on single Quarkus Buildpack.
+
+This would allow better control over build:
+* Better enablement for various package types (i.e. other than `fast-jar`).
+* ~~Better control over layering: separating directories according probability of their modification (application code vs. library code).~~
+
+However, it would be much more complicated to implement, and it would lead to duplicity.
```
ekcasey, line 48:
```suggestion
However, it would be much more complicated to implement, and it would lead to duplication.
```

text/java/0012-quarkus.md
ekcasey: > what do you mean by:
> > could/should we also handle the case where a quarkus fast-jar is provided to the build, rather than building from source?

Sorry I could have been more clear here. I'll elaborate.

Right now with a spring native app it is possible to build the JAR locally with maven and then build the image from the JAR using buildpacks:
```
./mvnw -Dmaven.test.skip=true package
pack build samples/native --builder paketobuildpacks/builder:tiny -p ./target/demo-0.0.1-SNAPSHOT.jar --env BP_NATIVE_IMAGE=true
```

Do we want to support a workflow like this for quarkus? Maybe it's impossible because you lose the `native-image.args` file but if this is a workflow users would want we could look into options such as having the buildpack set the native image build arguments directly.

ekcasey: > That's how it is implemented already: https://github.com/matejvasek/quarkus-bp/blob/main/quarkus/detect.go#L27
Edit: sry it's actually other way around; I think we shouldn't "fight" users who wants to set envvar by themselves.

I guess I am just generally asking for these details to be included in this RFC document for future reference and so we can discuss the pros/cons of various changes before we accept the buildpack and include it in the composite.

matejvasek: @ekcasey  I can see it working like:

```
./mvnw -Dmaven.test.skip=true -Dquarkus.package.type=native-source package
pack build samples/native --builder paketobuildpacks/builder:tiny -p ./target/native-sources --env BP_NATIVE_IMAGE=true --env BP_NATIVE_IMAGE_BUILD_ARGUMENTS="$(cat ./target/native-sources/native-image.args)"
```
But it doesn't work now because `-p` has to be jar not a directory containing jars.

ekcasey: I think this buildpack is a good idea but, If you don't mind @matejvasek , I would like to see the answers to some of these questions included in the RFC not just answers in the comments.

I think it's better to include the following in the RFC itself, not just in the linked repo
1. the detection criteria
2. the behavior at build time
3. supported configurations
4. opportunities for future extensions

This will give us a permanent record of the plan and intentions at the time of approval (things may change later), and give anyone reviewing RFC an opportunity to request on any of the above details if they would like to.

matejvasek: OK, I'll update the RFC.

matejvasek:
@ekcasey @dmikusa-pivotal I updated the PR.

ekcasey (approved):

dmikusa-pivotal (approved):

joshuatcasey:
2022-07-26: ready to merge by @dmikusa-pivotal

PR #172, opened 2022-04-07
RFC: Jammy Stacks

ryanmoran:
## Summary

[Readable](https://github.com/paketo-buildpacks/rfcs/blob/jammy-jellyfish/text/stacks/0004-jammy-jellyfish.md)

## Checklist

### DIFF
--- a/text/stacks/0004-jammy-jellyfish.md
+++ b/text/stacks/0004-jammy-jellyfish.md
@@ -1,0 +1,132 @@
+# Stacks based on Ubuntu 2022.04: Jammy Jellyfish
+
+## Summary
+
+A set of stacks based on the Ubuntu 2022.04 LTS (Jammy Jellyfish) release base
+image should be developed, released, and maintained by the Stacks subteam. Like
+the existing Bionic stacks, these new stacks should come in "full", "base", and
+"tiny" variants with similar, if not identical, sets of packages pre-installed.
+
+## Motivation
+
+Ubuntu provides a long-term-support (LTS) release every 2 years in April. These
+releases are supported by Canonical for 5 years. The current Bionic base image,
+first released in April 2018, will therefore be going out of support at the end
+of March 2023. In order to ensure that Paketo continues to provide a supported
+Ubuntu-based stack for our users, we should produce a new set of stack images,
+based on Jammy, alongside the current Bionic stacks.
+
+## Detailed Explanation
+
+### Variants
+
+The Jammy stack will be delivered in 3 variants that align to those already
+offered in the Bionic stack. The variants will be called `base`, `full`, and
+`tiny` just as they are in the Bionic stack. Each should be developed to offer
+roughly the same set of OS-level package support as their Bionic equivalent.
+
+### Stack IDs
+
+Stack IDs will be given to each variant of the stack as follows:
+
+* Base: `io.buildpacks.stacks.jammy`
+* Full: `io.buildpacks.stacks.jammy`
+* Tiny: `io.buildpacks.stacks.jammy.tiny`
+
+### User IDs
+
+These stacks will differ from Bionic with regards to their UID definitions.
+Each variant will ensure that the UID for the `build` phase is different than
+the `run` phase. This change will align more closely with the
+[recommendations](https://github.com/buildpacks/rfcs/blob/main/text/0085-run-uid.md)
+outlined in the Buildpacks Specification.
+
+### "Intermediate" Images
+
+The current Bionic stack is also shipped alongside an "intermediate" or "base"
+image that included all the packages, but is not decorated with the extra
+CNB-specific metadata. The thought going back was that this type of image would
+allow third-parties to more easily build a stack as they could simply extend
+upon the "intermediate" image.
+
+In order to simplify the stack offering going forward, we will no longer ship
+this "intermediate" image. Users wishing to extend the stack can simply build
+on top of the singular stack image as they would for any other type of
+container image workflow.
+
+### Image Naming and Tagging
+
+The stacks will name and tag their release images with the following pattern:
+
+```
+paketobuildpacks/{phase}-jammy-{variant}:{version}
+```
+
+For example we could see the following images for Jammy stacks:
+
+* `paketobuildpacks/jammy-build-base:latest`
+* `paketobuildpacks/jammy-run-tiny:1.2.3`
+
+This choice is different than the choice made with Bionic to better align the
+image repository references with their logical stack definition. This means
+that, with this naming scheme, it will be much more reasonable to understand
+what is stored at the repository reference
+`paketobuildpacks/jammy-run-base:latest`.
+
+Each stack repository should include a README that outlines the stacks that are
+available including links to each other repository allowing users to discover
+the stack variants available from any of the stack repository pages.
+
+#### Implications for Bionic
+
+The benefits of this naming format should also be extended to the Bionic
+stacks. This will help with consistency for stack consumers. However, we will
+need to still produce images that conform to the existing naming format as
+renaming these images is a breaking change that would surprise users.
+
+### SBOM
+
+This stack will **NOT** include an SBOM of the form outlined in Stacks
+[RFC0001](https://github.com/paketo-buildpacks/rfcs/blob/main/text/stacks/0001-stack-package-metadata.md).
+This choice is to ready ourselves to adopt whatever Stack SBOM functionality is
+specified by the Cloud Native Buildpacks project while ensuring we don't
+perpetuate the use of our own format for the entirety of this stack's support
+lifetime.
+
+This stack will adopt whatever official SBOM support is declared upstream by
+the CNB project.
+
+## Rationale and Alternatives
+
+We should at least discuss what it would mean to provide a Debian stack. There
+are [LTS releases](https://wiki.debian.org/LTS) for Debian, but they follow a
+different lifecycle and may not provide enough overlap to allow end-users to
+migrate completely.
+
+## Implementation
+
+After the [final
+release](https://discourse.ubuntu.com/t/jammy-jellyfish-release-schedule/23906)
+date on April 21, 2022, we can release our official stack versions.
+
+### Repositories
+
+The Stacks subteam will create 3 new repos for Jammy:
+
+* `jammy-base-stack`
+* `jammy-full-stack`
+* `jammy-tiny-stack`
+
+Each of these repos will contain the configuration for its variant of the stack
+as well as the releases and their related artifacts.
+
+## Prior Art
+
+* [`stacks`](https://github.com/paketo-buildpacks/stacks)
+* [`base-stack-release`](https://github.com/paketo-buildpacks/base-stack-release)
+* [`full-stack-release`](https://github.com/paketo-buildpacks/full-stack-release)
+* [`tiny-stack-release`](https://github.com/paketo-buildpacks/tiny-stack-release)
+
+## Unresolved Questions and Bikeshedding
+
+* What does this mean for builders?


text/stacks/0004-jammy-jellyfish.md
```diff
@@ -0,0 +1,106 @@
...
+For example we could see the following images for Jammy stacks:
+
+* `paketobuildpacks/jammy-build-base:latest`
+* `paketobuildpacks/jammy-run-tiny:1.2.3`
+
+This choice is different than the choice made with Bionic to better align the
+image repository references with their logical stack definition. This means
+that, with this naming scheme, it will be much more reasonable to understand
+what is stored at the repository reference
+`paketobuildpacks/jammy-run-base:latest`.
```
sophiewigmore, line 59: I much prefer the naming here, but I worry about the confusion that might be caused by having two naming schemas for stacks we will support at the same time for a while.

text/stacks/0004-jammy-jellyfish.md
```diff
@@ -0,0 +1,106 @@
...
+The Jammy stack will be delivered in 3 variants that align to those already
+offered in the Bionic stack. The variants will be called `base`, `full`, and
+`tiny` just as they are in the Bionic stack. Each should be developed to offer
+roughly the same set of OS-level package support as their Bionic equivalent.
+
+### Stack IDs
+
+Each of these stacks will declare their stack ID as `io.paketo.stack.jammy`.
+This choice differs from Bionic in that we will **NOT** differentiate between
+the `tiny` and `base/full` variants using the stack ID.
```
sophiewigmore, line 32: How is this different? I thought that the Stack ID was already 
`"io.buildpacks.stack.id": "io.buildpacks.stacks.bionic"`

sophiewigmore, line 32: How is this different? I think that base and full use `"io.buildpacks.stack.id": "io.buildpacks.stacks.bionic"`, and just tiny has a different ID. Are there any ramifications of changing that?

text/stacks/0004-jammy-jellyfish.md
```diff
@@ -0,0 +1,106 @@
+# Stacks based on Ubuntu 2022.04: Jammy Jellyfish
+
+## Summary
+
+A set of stacks based on the Ubuntu 2022.04 LTS (Jammy Jellyfish) release base
+image should be developed, released, and maintained by the Stacks subteam. Like
+the existing Bionic stacks, these new stacks should come in "full", "base", and
+"tiny" variants with similar, if not identical, sets of packages pre-installed.
```
sophiewigmore, line 8: Will we be shipping two sets of each stack like we do for bionic? One without CNB metadata, and one with it (stacks with the `-cnb` suffix)?

text/stacks/0004-jammy-jellyfish.md
```diff
@@ -0,0 +1,106 @@
...
+release](https://discourse.ubuntu.com/t/jammy-jellyfish-release-schedule/23906)
+date on April 21, 2022, we can release our official stack versions.
+
+### Repositories
+
+The Stacks subteam will create 3 new repos for Jammy:
+
+* `jammy-base-stack`
+* `jammy-full-stack`
+* `jammy-tiny-stack`
```
sophiewigmore, line 92: The repos for the current stacks are `<full/base/tiny>-stack-release`. Could we follow suit and name the repositories `jammy-<full/base/tiny>-stack-release` for consistency?

text/stacks/0004-jammy-jellyfish.md
```diff
@@ -0,0 +1,106 @@
...
+## Prior Art
+
+* [`stacks`](https://github.com/paketo-buildpacks/stacks)
+* [`base-stack-release`](https://github.com/paketo-buildpacks/base-stack-release)
+* [`full-stack-release`](https://github.com/paketo-buildpacks/full-stack-release)
+* [`tiny-stack-release`](https://github.com/paketo-buildpacks/tiny-stack-release)
+
+## Unresolved Questions and Bikeshedding
+
+* What does this mean for builders?
```
sophiewigmore, line 106: I would imagine we'd ship another set of builders for the jammy stacks. @paketo-buildpacks/builders-maintainers thoughts?

brayanhenao:
Is it safe to say that we will maintain support for two different versions of Ubuntu for about a year ([Bionic's EOL](https://wiki.ubuntu.com/Releases)) or will we drop it before then?

text/stacks/0004-jammy-jellyfish.md
ryanmoran: Its different in that the "tiny" variant will have the same stack ID as "base" and "full".

As far as ramifications go, I think the experience for "tiny" now becomes like that for "base" and "full" in that there are some buildpacks that will operate successfully on "full", but not on "base", and that we document that in the builder or documentation for the buildpack itself. Having a separate stack ID just created extra metadata that needed to be added and maintained in all the buildpacks that supported "tiny", with very little technical benefit.

text/stacks/0004-jammy-jellyfish.md
ryanmoran: No. I'd like to see us drop the set of images that isn't a valid CNB stack. We should just ship valid CNB stacks. If users want to build on top of our stack, they can just do that directly. We won't ship them an intermediate image.

Therefore, it makes sense for us to drop the `-cnb` suffix as it doesn't provide much in the way of information.

text/stacks/0004-jammy-jellyfish.md
ryanmoran: Why do these repos have the `-release` suffix on them? I've never really understood that part.

text/stacks/0004-jammy-jellyfish.md
sophiewigmore: I'd be good with that

text/stacks/0004-jammy-jellyfish.md
sophiewigmore: I'm all for that. Do we think we can cut support of the separate stacks in the Bionic stacks as well?

ryanmoran: I'd like to see that happen. Maybe we ship the same image to both the `{variant}` and `{variant}-cnb` tags.

text/stacks/0004-jammy-jellyfish.md
sophiewigmore: Not sure, I didn't make them. Probably to show that that's the repository where releases are cut. Most of the heavy duty logic and important stuff, like the `create-stack`, Dockerfiles, and package lists live in the Stacks repo, so all that goes down in the `-release` repos are just for putting the stack together and releasing it. They don't have much else.

The `-release` suffix doesn't release bother me and as above I'd rather everything just be consistent to make our lives every so slightly  easier

text/stacks/0004-jammy-jellyfish.md
ryanmoran: My hope is that with the addition of the [Stack descriptor RFC](https://github.com/paketo-buildpacks/rfcs/blob/stack-descriptor/text/stacks/0003-stack-descriptor.md) we won't need this type of intermediate image. Folks wanting to recreate our stacks can just use the descriptor and modify it to support their specific use case.

I think we should ship the same image in both locations and then assess the feedback if/when it comes in from users.

text/stacks/0004-jammy-jellyfish.md
ryanmoran: I'd like to see us move the stack metadata into these variant-specific repos. Its weird that we create a stack in one repo and then release it in another, no?

Especially if the Stack descriptor RFC is adopted, then the utility of the `stacks` repo starts to come into question.

text/stacks/0004-jammy-jellyfish.md
```diff
@@ -0,0 +1,113 @@
...
+
+The Jammy stack will be delivered in 3 variants that align to those already
+offered in the Bionic stack. The variants will be called `base`, `full`, and
+`tiny` just as they are in the Bionic stack. Each should be developed to offer
+roughly the same set of OS-level package support as their Bionic equivalent.
+
+### Stack IDs
+
+Each of these stacks will declare their stack ID as `io.paketo.stack.jammy`.
+This choice differs from Bionic in that we will **NOT** differentiate between
```
dmikusa-pivotal, line 31: On the Java buildpacks, there are a couple of places where we have to handle the Tiny stack specifically. I believe it's presently set up to look at the stack id to determine if we are running on Tiny. If we use the same stack id for all three variants how would one go about detecting when you are running on the tiny stack? Is there other metadata which can be used to determine that?

text/stacks/0004-jammy-jellyfish.md
dmikusa-pivotal: Just offering an outsider's point of view on the different repos, they are confusing. It takes some effort to track down where things live and what is located in each part. I think this comes into play if you want to see what's in the various images, or fork and customize. 

It's also a problem if you want to open an issue. Where would one go? It's not exactly clear without investigating more, and not all users will put in that level of effort.

Again, my outsider's point of view would be to simplify and consolidate repos (across stacks & builders). I recognize that may not be technically practical, which is fine, but I think we have to own the documentation and explanation of this in our repos then. [What's here](https://paketo.io/docs/concepts/stacks/#what-is-a-stack) is a start, but it's missing docs around the repos, how to navigate them, and an explanation of how the systems come together to produce the stack.

text/stacks/0004-jammy-jellyfish.md
ryanmoran: Maybe we can detour into why the buildpacks have that code first. Where are the places that the buildpack relies specifically on the tiny stack ID?

dmikusa-pivotal: The one off the top of my head is Tomcat (and now Tomee). Those servers have a similar shell script that runs to launch the process, so process types for full/base use the script. There's no shell on tiny, so we do some similar work to what the script does in the buildpack and generate a process type that can start the script w/out a shell. It's not as full-featured as the script, but it does make it work.

I also know that we're doing some stuff in the Rust buildpack to handle Tiny specifically as well. The tiny stack is missing a library that Rust compiled apps require, so we're doing a static compile with musl instead of glibc to ensure the binary will run on Tiny.

I can check for other instances tomorrow.

text/stacks/0004-jammy-jellyfish.md
```diff
@@ -0,0 +1,113 @@
...
+
+The stacks will name and tag their release images with the following pattern:
+
+```
+paketobuildpacks/jammy-{phase}-{variant}:{version}
+```
+
+For example we could see the following images for Jammy stacks:
+
+* `paketobuildpacks/jammy-build-base:latest`
```
ekcasey, line 52: In my mind one drawback to pushing the differentiation into the repo name vs into the tag name is discoverability. For example, it's easy to go to https://hub.docker.com/_/ubuntu and, by reading the readme,  see what tags are available and how they relate. Yes, you could still go to the `paketobuildpacks` [org page](https://hub.docker.com/u/paketobuildpacks) and looking for stack repos, but there is a lot more noise to wade through and no opportunity to use a readme as a guidepost 

What if we did something like:
`paketobuildpacks/build:jammy-tiny-1.2.3`


Then, in the future we could even do things like repoint `:latest` from one ubuntu LTS to another (provided the CNB tooling eventually figures out how to preform a stack upgrade without generating bad images).

text/stacks/0004-jammy-jellyfish.md
```diff
@@ -0,0 +1,113 @@
...
+The Jammy stack will be delivered in 3 variants that align to those already
+offered in the Bionic stack. The variants will be called `base`, `full`, and
+`tiny` just as they are in the Bionic stack. Each should be developed to offer
+roughly the same set of OS-level package support as their Bionic equivalent.
+
+### Stack IDs
+
+Each of these stacks will declare their stack ID as `io.paketo.stack.jammy`.
+This choice differs from Bionic in that we will **NOT** differentiate between
+the `tiny` and `base/full` variants using the stack ID.
```
ekcasey, line 32: I think there are some existing cases where paketo buildpacks use the stack ID to differentiate between tiny and base in order to behave differently. One example is the Procfile buildpack which behaves differently in the presence and absence of a shell.

I suspect most of these cases are about the presence of bash so maybe we can just remove these as we try to encourage shell-less images? But I think there is at least one example with java native images where the buildpack needs to know whether to statically compile in certain libraries (cc @dmikusa-pivotal )

dmikusa-pivotal: I have this concern as well. Raised above.  https://github.com/paketo-buildpacks/rfcs/pull/172#discussion_r847604706

text/stacks/0004-jammy-jellyfish.md
dmikusa-pivotal: Emily mentioned one other case, https://github.com/paketo-buildpacks/rfcs/pull/172#discussion_r848728085, This is for native-image builds with Java. The root problem is similar to Rust. There is library not in Tiny that causes the default build to fail. We add a flag to the build arguments for Tiny which causes the native image to be built in a way that does not require that library. 

I would have to go back and double-check if this is still an issue. The buildpack is 100% still setting this flag, but I haven't checked to see if the present Tiny stack still requires this and don't have notes on exactly what breaks w/out this flag being set.

So in summary:

1. Tomcat & Tomee & Procfile -> look if it's tiny to take action to deal with their being no shell in Tiny.
2. Rust Cargo -> adjusts build arguments to accommodate the fact that tiny does not contain [libgcc1 package](https://packages.ubuntu.com/bionic-updates/libgcc1). Rust code compiled against glibc requires this library.
3. Native Image -> requires setting an alternative flag on build

text/stacks/0004-jammy-jellyfish.md
ryanmoran: I think the biggest issue we encountered with a consolidated repository structure was in releasing the sub-artifacts (stack or builder) on differing cadences. Tracking things like "what is the next version I should release?" become harder when the repo contains more than one logical artifact. I'm all for documenting this better, but the consolidation problem is probably a bigger issue than I think we should outline a solution for in this RFC.

text/stacks/0004-jammy-jellyfish.md
ryanmoran: What is `:latest` in this case? `jammy-base-latest`? What does this mean if we also have UBI stacks? We will soon be in a state where we have so many stacks and variants that something like a `:latest` tag becomes quite tricky to explain, let alone rely upon in a real usecase.

ryanmoran: As for discovery, I agree that we could be doing better. Barring a discovery of some sort of native solution in Dockerhub, could we consider modifications to the Paketo website to make this type of discovery easier?

text/stacks/0004-jammy-jellyfish.md
```diff
@@ -0,0 +1,113 @@
...
+### Variants
+
+The Jammy stack will be delivered in 3 variants that align to those already
+offered in the Bionic stack. The variants will be called `base`, `full`, and
+`tiny` just as they are in the Bionic stack. Each should be developed to offer
+roughly the same set of OS-level package support as their Bionic equivalent.
+
+### Stack IDs
+
+Each of these stacks will declare their stack ID as `io.paketo.stack.jammy`.
```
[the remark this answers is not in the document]
ryanmoran: Updated.

text/stacks/0004-jammy-jellyfish.md
dmikusa-pivotal: I'm OK with that, taking the documentation approach. This is an area where I've heard feedback, so I wanted to make sure it's on the record. Thanks

text/stacks/0004-jammy-jellyfish.md
fg-j: As it stands, we release 6 builders (tiny, base, full, and buildpackless varieties of these). If we duplicate these efforts for jammy, we'll be on the hook for 12 builders in total. That feels like a lot, but perhaps makes sense until Bionic EOLs next year. 

Since builders aren't needed in perpetuity for rebasing, perhaps once bionic is EOL, the current builder repositories are switched from publish bionic to jammy builders?

There's also the topic of repo and tag naming, which will need to be ironed out. Would we do something like:
`paketobuildpacks/builder:<version>-<buildpackless?>-<OS>-<variant>`
`paketobuildpacks/builder:1.2.3-buildpackless-jammy-full`

text/stacks/0004-jammy-jellyfish.md
sophiewigmore: RE: bionic, is this a note that you could add to the RFC or are you keeping it strictly about Jammy? I'd like our plan documented somewhere.

ryanmoran: I can add something indicating that we'll only be shipping the one image type, and that there will be no "intermediate" image. The language will be specific to Jammy, and we can choose to adopt that in Bionic as well in some other process or RFC.

ryanmoran: I've updated the RFC to explain the decision made here.

text/stacks/0004-jammy-jellyfish.md
ryanmoran: Resolved in https://github.com/paketo-buildpacks/rfcs/pull/172#discussion_r853193113

text/stacks/0004-jammy-jellyfish.md
sophiewigmore: Sounds good to me.

text/stacks/0004-jammy-jellyfish.md
sophiewigmore: In keeping with the resolutions above, I'm OK to do this for jammy stacks, and then we can open a subsequent RFC down the line to move the bionic stacks from the current locations to repositories of naming convention `bionic-<stack>-stack`, as well as propose other changes to the bionic stacks to align with decisions here.

text/stacks/0004-jammy-jellyfish.md
ryanmoran: I've updated the RFC to outline that we should include a README on each stack repository that links out to the other variants to improve discoverability.

As for the underlying naming question, I still think the current RFC specification makes the most sense as a naming structure. It splits the stacks along their logical differences.

For some of the examples that had been proposed, like [Ubuntu](https://hub.docker.com/_/ubuntu), its possible to organize the images in the repository into a single linear ordering of versions. Each tag represents a different version of the same logical artifact, the Ubuntu OS distribution.

For other examples, like [Go](https://hub.docker.com/_/golang), its not clear that the choice to put everything in a single repository is a deliberate one or that it falls out of the constraints of there being no such thing as an organization within the top-level Dockerhub "library".

dmikusa-pivotal (approved):

brayanhenao (approved):

text/stacks/0004-jammy-jellyfish.md
sophiewigmore: Is this a point we should be aligned on for this RFC? If so, is @fg-j suggestion the right path forward here? I can't remember what we discussed around this point at the working group meeting. Would not having jammy builders until next year be a hindrance to users?

ryanmoran: I don't think this is in scope for this RFC. We will need to figure this out, and so I wanted builder maintainers to provide some input, but I don't think we need to actually decide the names for builders for this RFC to pass.

text/stacks/0004-jammy-jellyfish.md
```diff
@@ -0,0 +1,132 @@
...
+this "intermediate" image. Users wishing to extend the stack can simply build
+on top of the singular stack image as they would for any other type of
+container image workflow.
+
+### Image Naming and Tagging
+
+The stacks will name and tag their release images with the following pattern:
+
+```
+paketobuildpacks/jammy-{phase}-{variant}:{version}
```
ekcasey, line 62:
```suggestion
paketobuildpacks/{phase}-jammy-{variant}:{version}
```
Nit: Happy to go with this per-repo approach but what do you think about moving the phase first in the order so that e..g all run image repos begin with`paketobuildpacks/run-`? My approval stand regardless, just a suggestion

ryanmoran: Switched to your proposed naming scheme.

PR #183, opened 2022-04-19
Contributes RFC for a Health Checker buildpack

dmikusa-pivotal:
## Summary

Introduce a Health Checker buildpack. This buildpack will provide access to one or more health checkers, where a health checker is a small binary that will be installed into the container image for the purpose of validating the health of the running image.

[Readable](https://github.com/paketo-buildpacks/rfcs/blob/health-check/text/utilities/0000-health-checker-buildpack.md)

## Checklist

### DIFF
--- a/text/utilities/0000-health-checker-buildpack.md
+++ b/text/utilities/0000-health-checker-buildpack.md
@@ -85,8 +85,8 @@ A user can then utilize the health checker like this:
         Downloading from https://github.com/dmikusa-pivotal/tiny-health-checker/releases/download/v0.5.0/thc-x86_64-unknown-linux-musl
         Verifying checksum
         Copying from /tmp/ce75bb97209981e03bf7e8aa52e2bfab78a50a44c7ed1787f4ace212711d61e5/thc-x86_64-unknown-linux-musl to /layers/paketo-buildpacks_health-checker/thc/bin
-        Writing env.launch/health-check/THC_PATH.default
-        Writing env.launch/health-check/THC_PORT.default
+        Writing env.launch/health/THC_PATH.default
+        Writing env.launch/health/THC_PORT.default
       Process types:
     health-check: thc (direct)
     ```
--- a/text/utilities/0000-health-checker-buildpack.md
+++ b/text/utilities/0000-health-checker-buildpack.md
@@ -91,7 +91,7 @@ A user can then utilize the health checker like this:
     health-check: thc (direct)
     ```
 
-2. You can run this with `docker run --health-cmd '/cnb/process/health-check' --health-interval 5s --health-timeout 2s -it apps/maven`. You could call the health check binary directly, but as you can have different health check dependencies that get installed by the buildpack, using the process type gives you more generic way to call the health check. It also allows for the image to bake in the health check configuration, rather than require the operator to set it. Plus, an operator can always override the configuration by setting the appropriate environment variables at runtime.
+2. You can run this with `docker run --health-cmd 'health' --health-interval 5s --health-timeout 2s -it apps/maven`. You could call the health check binary directly, but as you can have different health check dependencies that get installed by the buildpack, using the process type gives you more generic way to call the health check. It also allows for the image to bake in the health check configuration, rather than require the operator to set it. Plus, an operator can always override the configuration by setting the appropriate environment variables at runtime.
 
 ## Prior Art
 
--- a/text/utilities/0000-health-checker-buildpack.md
+++ b/text/utilities/0000-health-checker-buildpack.md
@@ -1,0 +1,47 @@
+# Health Checker Support
+
+## Summary
+
+Introduce a Health Checker buildpack. This buildpack will provide access to one or more health checkers, where a health checker is a small binary that will be installed into the container image for the purpose of validating the health of the runnning image.
+
+By default, we'll provide a tiny HTTP health checker which can be used to probe a port and path. An HTTP 200 OK response returns exit code 0 and a non-2xx response returns 1.
+
+The Health Checker buildpack would be maintained by the existing [Utilities team](https://github.com/orgs/paketo-buildpacks/teams/utilities). It uses libpak & libcnb.
+
+## Motivation
+
+When running buildpack generated images on Kubernetes, HTTP-based health checking is provided by the platform. When running your buildpack generated images with Docker or Docker Compose, there is no option for HTTP-based health checks. It requires a binary in the image to perform the HTTP operation. This is typically `curl` or `wget`, however, the Paketo tiny and base images do not include these binaries.
+
+Having a buildpack that installs a health checker allows for HTTP-based health checks when running on Docker. In addition, it could allow for other types of health checkers for example if your application does not use HTTP.
+
+## Detailed Explanation
+
+1. A health checker buildpack has been developed. It will be contributed to Paketo.
+   1. By default, no health checker will be installed by the buildpack.
+   2. If a user sets `BP_HEALTH_CHECKER_ENABLED=true` then the buildpack will detect and contribute a health check binary to the `$PATH`.
+   3. A process type of `health-check` will be contributed with the command to run the health checker. This may be invoked by the user rather than needing the full command.
+
+## Rationale and Alternatives
+
+1. A user can create a custom stack and include a tool like `curl` or `wget`. This is a lot of additional work, plus `curl` and `wget` are overkill for a simple HTTP health check. To reduce footprint, they are not recommended.
+
+## Implementation
+
+The new Health Checker buildpack will participate if all the following conditions are met
+
+* A user sets `BP_HEALTH_CHECKER_ENABLED=true`
+
+When conditions are met, the buildpack will do the following:
+
+* Contributes a health checker to a layer marked `launch` with command on `$PATH`
+* Contributes a process type called `health-check` which will execute the installed health checker.
+
+The Health Checker buildpack will optionally be added to the language family buildpacks that wish to include support for it. The primary targets will be the Java and Rust buildpacks.
+
+## Prior Art
+
+None
+
+## Unresolved Questions and Bikeshedding
+
+None
--- a/text/utilities/0000-health-checker-buildpack.md
+++ b/text/utilities/0000-health-checker-buildpack.md
@@ -27,16 +27,52 @@ Having a buildpack that installs a health checker allows for HTTP-based health c
 
 ## Implementation
 
-The new Health Checker buildpack will participate if all the following conditions are met
+The new Health Checker buildpack will participate if either of the following conditions are met
 
 * A user sets `BP_HEALTH_CHECKER_ENABLED=true`
+* An upstream buildpack requests `hc` in the build plan.
 
 When conditions are met, the buildpack will do the following:
 
 * Contributes a health checker to a layer marked `launch` with command on `$PATH`
 * Contributes a process type called `health-check` which will execute the installed health checker.
 
-The Health Checker buildpack will optionally be added to the language family buildpacks that wish to include support for it. The primary targets will be the Java and Rust buildpacks.
+The Health Checker buildpack will optionally be added to the language family buildpacks that wish to include support for it. The primary targets will be the Java and Rust buildpacks. The buildpack should be added reasonably close to the end of an order group for a given language family.
+
+Other buildpacks in the order group may influence the configuration of the the health checker by modifying the `BP_HEALTH_CHECKER_*` environment variables. It is recommended that buildpacks use the `.default` setting when doing this so that any user specified options will have the highest precedence.
+
+The following configuration options will be available to the buildpack.
+
+| Environment Variable               | Description                                                                                                                                                                                      |
+| ---------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `$BP_HEALTH_CHECKER_ENABLED`       | If set to `true` the buildpack will contribute a health checker binary. Defaults to `false`, so no health checker is contributed.                                                                |
+| `$BP_HEALTH_CHECKER_DEPENDENCY`    | The dependency id in `buildpack.toml` of the health checker to install.                                                                                                                          |
+| `$BP_HEALTH_CHECKER_<dep-id>_ARGS` | A set of arguments to pass through to the `<dep-id>` health checker. These only apply for the specific health checker. This allows specific default arguments for each supported health checker. |
+
+To clarify, there will be one `$BP_HEALTH_CHECKER_<dep-id>_ARGS` env variable for each health checker that is supported by the buildpack. This allows the buildpack to set different defaults per health checker. It also allows other buildpacks to override arguments only for a specific health checker. Like if Spring Boot CNB wants to set one path for the HTTP-based Actuator endpoints and a different configuration for RSocket-based health checks.
+
+#### Default Health Checker
+
+The default health checker that we'll support initially is [tiny-health-checker](https://github.com/dmikusa-pivotal/tiny-health-checker). This has support for minimal HTTP GET health checks. It only allows health checks to `localhost`, but does support a configurable port and path. This is intentional to prohibit it from being used to exfiltrate data in a comprmised container.
+
+Full usage:
+
+```
+USAGE:
+	thc [port] [path]
+
+ARGS:
+
+	port is the port to which a connection will be made, default: 8080
+	path is the path to which a connection will be made, default: /
+
+ENV:
+
+	CONN_TIMEOUT sets the connection timeout, default: 10
+	REQ_TIMEOUT sets the request timeout, defaults: 15
+
+	**NOTE** Host is not configurable and will always be localhost
+```
 
 ### Example
 
--- a/text/utilities/0000-health-checker-buildpack.md
+++ b/text/utilities/0000-health-checker-buildpack.md
@@ -38,6 +38,29 @@ When conditions are met, the buildpack will do the following:
 
 The Health Checker buildpack will optionally be added to the language family buildpacks that wish to include support for it. The primary targets will be the Java and Rust buildpacks.
 
+### Example
+
+A user can then utilize the health checker like this:
+
+1. For a Spring Boot app, run `pack build apps/maven -p target/demo-0.0.1-SNAPSHOT.jar -b paketo-buildpacks/java -b docker.io/paketo-buildpacks/health-checker -e BP_HEALTH_CHECKER_ENABLED=true -e BP_HEALTH_CHECKER_THC_ARGS='8080 /actuator/health'` (it won't be necessary to include the `-b` arguments once the buildpack has been integrated with language family buildpacks). You'll see output like this at the end. It's installing the health checker and setting a process type.
+
+    ```
+    Paketo Health Checker Buildpack vDEVELOPMENT
+      https://github.com/paketo-buildpacks/health-checker
+      Build Configuration:
+        $BP_HEALTH_CHECKER_DEPENDENCY  thc                    which health checker to contribute
+        $BP_HEALTH_CHECKER_ENABLED     true                   contributes a health checker if enabled
+        $BP_HEALTH_CHECKER_THC_ARGS    8080 /actuator/health  arguments passed to tiny-health-checker, if used
+      Tiny Health Checker 0.4.0: Contributing to layer
+        Downloading from https://github.com/dmikusa-pivotal/tiny-health-checker/releases/download/v0.4.0/thc-x86_64-unknown-linux-musl
+        Verifying checksum
+        Copying from /tmp/ce75bb97209981e03bf7e8aa52e2bfab78a50a44c7ed1787f4ace212711d61e5/thc-x86_64-unknown-linux-musl to /layers/paketo-buildpacks_health-checker/thc/bin
+      Process types:
+    health-check: thc 8080 /actuator/health (direct)
+    ```
+
+2. You can run this with `docker run --health-cmd '/cnb/process/health-check' --health-interval 5s --health-timeout 2s -it apps/maven`. You could call the health check binary directly, but as you can have different health check dependencies that get installed by the buildpack, using the process type gives you more generic way to call the health check. It also allows for the image to bake in the health check arguments, rather than require the operator to set them. Plus, an operator could always call the binary directly if they wanted.
+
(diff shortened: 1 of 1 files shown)


garethjevans:
Will the healthcheck thats created by this buildpack be compatible with https://docs.docker.com/engine/reference/builder/#healthcheck?

dmikusa-pivotal:
> Will the healthcheck thats created by this buildpack be compatible with https://docs.docker.com/engine/reference/builder/#healthcheck?

The [default implementation I built](https://github.com/dmikusa-pivotal/tiny-health-checker) does this:

>>0: success - the container is healthy and ready for use
>>1: unhealthy - the container is not working correctly

The goal for the buildpack is that it's not tied to a particular health checker though. If mine stinks and we can add another one and users can pick. In addition, my implementation only checks HTTP. Maybe your app speaks another protocol like rsocket, so there could be a different health checker for it. 

Ultimately, whatever health checkers we include should follow the Docker specification since that's the primary target for this. Kubernetes follows the same basic pattern, although any non-zero return code is a failure with K8s.

text/utilities/0000-health-checker-buildpack.md
```diff
@@ -0,0 +1,47 @@
...
+1. A health checker buildpack has been developed. It will be contributed to Paketo.
+   1. By default, no health checker will be installed by the buildpack.
+   2. If a user sets `BP_HEALTH_CHECKER_ENABLED=true` then the buildpack will detect and contribute a health check binary to the `$PATH`.
+   3. A process type of `health-check` will be contributed with the command to run the health checker. This may be invoked by the user rather than needing the full command.
+
+## Rationale and Alternatives
+
+1. A user can create a custom stack and include a tool like `curl` or `wget`. This is a lot of additional work, plus `curl` and `wget` are overkill for a simple HTTP health check. To reduce footprint, they are not recommended.
+
+## Implementation
```
ekcasey, line 28: I think it would be good to include an example of how this could be used with `docker run` (maybe `docker compose` for extra credit but it's not as essential).

text/utilities/0000-health-checker-buildpack.md
```diff
@@ -0,0 +1,47 @@
...
+
+The new Health Checker buildpack will participate if all the following conditions are met
+
+* A user sets `BP_HEALTH_CHECKER_ENABLED=true`
+
+When conditions are met, the buildpack will do the following:
+
+* Contributes a health checker to a layer marked `launch` with command on `$PATH`
+* Contributes a process type called `health-check` which will execute the installed health checker.
```
ekcasey, line 37: I am assuming we will want runtime configuration options for setting the health url (or host/port/path separately)? If so
1. What should the defaults be?
2. Should other buildpacks in the composite buildpack be able to configure these values (.e.g the sping boot buidpacks sets the URL to the actuator endpoint)?

dmikusa-pivotal: > 1.What should the defaults be?

At the moment, my plan is to have the following config options:

- `BP_HEALTH_CHECKER_ENABLED` -> default false, true means we contribute a health checker
- `BP_HEALTH_CHECKER_DEPENDENCY` -> defaults to `thc` (tiny-health-checker, which is a small http-based health checker I wrote), but can be any dependency id from buildpack.toml's metadata. The idea is that there can be more than one health checker since not every app might provide an HTTP endpoint (could be some other protocol).
- Generically, `BP_HEALTH_CHECKER_<dep-id>_ARGS` would be options you can pass to a given health checker. I have this namespace by the dep id because I think you would want different defaults per health checker. I could possibly see a case for just having `BP_HEALTH_CHECKER_ARGS` that apply to any health checker though, so interested in thoughts on that.

>2. Should other buildpacks in the composite buildpack be able to configure these values (.e.g the sping boot buidpacks sets the URL to the actuator endpoint)?

Right now, I have it set so that an upstream buildpack can request `hc` in the build plan and it'll activate and add a health checker. I did not expose the ability to pick a particular type of health checker or pass args, but I think a buildpack could set that through env variables too, right? If there's a need to configure those through the buildplan, I could add that too.

Do you think this needs to be documented in the RFC? I was thinking the specific options were not super important since this is only intended to be a single implementation, not a standard that multiple buildpacks would implement. I can add it if you want, just let me know.

text/utilities/0000-health-checker-buildpack.md
dmikusa-pivotal: Example added. Docker compose should be similar, just set the health check args in yaml.

text/utilities/0000-health-checker-buildpack.md
ekcasey: > I did not expose the ability to pick a particular type of health checker or pass args, but I think a buildpack could set that through env variables too, right? If there's a need to configure those through the buildplan, I could add that too.

Conventionally we don't have buildpacks to set `BP_*` environment variables, reserving those for users, instead preferring a `BPI_* (buildpack internal) for buildpack-to-buildpack communication. Although, I don't actually see a practical problem so long as the buildpack uses `.default` to ensure the user-provided options take precedence.

Is the health-checker intended to go at the end or the beginning of the order? If it is at the end environment variables are the correct option for passing config forward. If it is at the beginning we need a build plan mechanism for passing config upstream (relative to execution order). My instincts say having it at the end is better but I don't have a strong argument for why and I could be persuaded to go the other way.

>  Do you think this needs to be documented in the RFC? I was thinking the specific options were not super important since this is only intended to be a single implementation, not a standard that multiple buildpacks would implement.

I would like to see details for the buildpack interface and the `thc` interface included in the RFC. If other buildpacks want to include an http healthcheck wouldn't they be using this interface? If I am understanding correclty other implementations would require additions to this interface?

> Generically, BP_HEALTH_CHECKER_<dep-id>_ARGS would be options you can pass to a given health checker. I have this namespace by the dep id because I think you would want different defaults per health checker. I could possibly see a case for just having BP_HEALTH_CHECKER_ARGS that apply to any health checker though, so interested in thoughts on that.

That makes sense given multiple health check implementations. Can you talk through the tradeoff of multiple implementations as the extension point vs. adding configuration options for a single implementation (e.g. http vs. tcp). This doesn't seem to provide a mechanism for injecting an arbitrary implementation and if we are controlling it a single implementation might be easier? Curious to hear your thoughts.

text/utilities/0000-health-checker-buildpack.md
```diff
@@ -0,0 +1,70 @@
...
+
+1. For a Spring Boot app, run `pack build apps/maven -p target/demo-0.0.1-SNAPSHOT.jar -b paketo-buildpacks/java -b docker.io/paketo-buildpacks/health-checker -e BP_HEALTH_CHECKER_ENABLED=true -e BP_HEALTH_CHECKER_THC_ARGS='8080 /actuator/health'` (it won't be necessary to include the `-b` arguments once the buildpack has been integrated with language family buildpacks). You'll see output like this at the end. It's installing the health checker and setting a process type.
+
+    ```
+    Paketo Health Checker Buildpack vDEVELOPMENT
+      https://github.com/paketo-buildpacks/health-checker
+      Build Configuration:
+        $BP_HEALTH_CHECKER_DEPENDENCY  thc                    which health checker to contribute
+        $BP_HEALTH_CHECKER_ENABLED     true                   contributes a health checker if enabled
+        $BP_HEALTH_CHECKER_THC_ARGS    8080 /actuator/health  arguments passed to tiny-health-checker, if used
```
ekcasey, line 53: I would like to see a discussion of the implementation. I.E what does the buildpack do with `BP_HEALTH_CHECKER_THC_ARGS`? Are they added as `args` in the process definition? That would make them hard to override at runtime without the changes proposed in currently in-flight [process refactor RFC](https://github.com/buildpacks/rfcs/blob/main/text/0093-remove-shell-processes.md) but might work very nicely once it lands.

The other option would be to configure the `thc` via runtime environment variables (with defaults set the buildpack).

text/utilities/0000-health-checker-buildpack.md
dmikusa-pivotal: >Is the health-checker intended to go at the end or the beginning of the order?

End. Adding to the RFC.

>My instincts say having it at the end is better but I don't have a strong argument for why and I could be persuaded to go the other way.

Same. I'm not totally sure why, but I tend to add things to the end unless there's a specific reason not to.

>BPI_* (buildpack internal) for buildpack-to-buildpack communication. Although, I don't actually see a practical problem so long as the buildpack uses .default`

I've typically used BPI for moving internal settings from build to runtime (like exec.d proc needs XYZ), things a user shouldn't see/care about. Using `.default` would make sense since the user should always have the final say.

I hadn't really considered the case of another buildpack adding the health check when I was writing this because it requires knowledge about the runtime deployment of the image being generated to really know if you should configure a health check. If you're targeting Docker or something without a health check then add it, otherwise it's noise/wasted space. Only the user really knows that. 

I do think there could be a benefit though, like you said, in having Spring Boot CNB look to see if a health check has been requested (if the env is set to true) and then configure the health checker arguments to look at the actuator endpoint, although that location can be reconfigured so I'd have to look into how we could reliably get that info.

Anyway, the intent is to have it at the end of the order groups so env variables make sense. I'll add a note about buildpack-to-buildpack comms in the RFC and recommend using `.default`.

>I would like to see details for the buildpack interface and the thc interface included in the RFC. If other buildpacks want to include an http healthcheck wouldn't they be using this interface? 

Done.

>If I am understanding correclty other implementations would require additions to this interface?

It depends on how you look at it :) I'm hoping it is sufficient to document the interface as having one `$BP_HEALTH_CHECKER_<dep-id>_ARGS` per health checker implementation that is supported. So the interface is defined as dependent on the name of the health checker dependency.

I guess if we wanted to be more explicit about it, I could put the specific dep id that we plan to support out of the box and we could submit adjustment RFCs to add new health checker implementations. If you feel like we should go that route, it's fine by me.

>Can you talk through the tradeoff of multiple implementations as the extension point vs. adding configuration options for a single implementation (e.g. http vs. tcp). This doesn't seem to provide a mechanism for injecting an arbitrary implementation and if we are controlling it a single implementation might be easier?

My thought for having multiple separate implementations is that it makes the health checkers smaller and simpler. The present one is a static binary just a little under 1M. I haven't done it yet, but a quick test says`upx` would get that down to about 350k. Hopefully there is a day where we can safely ship upx compressed binaries in buildpacks again :) 

Anyway, people seem to really value keeping the size of dependencies small so that was the primary factor in going this route. If we look at adding additional features like Rsocket or gRPC, that would pull in dependencies not everyone needs and making the binary larger.

Aside from that, I was hoping that over time there might be some OSS "standard" or community agreed-upon health checker that pops up and we could use that instead, so we don't have to maintain an actual health checker.

text/utilities/0000-health-checker-buildpack.md
dmikusa-pivotal: This is a fair point. 

The current stance is that we use `BP_HEALTH_CHECKER_THC_ARGS` (`BP_*` specifically) because the arguments get set at build time and are not intended to be changed at runtime. They are added directly to the process type arguments list.

The thought process is that the application developer will know what health check they have exposed and what is a meaningful health check, thus they should define it and make it part of the image. That makes things easier for the operator at runtime. They don't have to think about what health check to use, they just run `/cnb/process/health-check` and it should work.

I'm not necessarily opposed to adding runtime configuration, if others think it will be of value. I'm not sure that using `BPL_*` makes sense though. There would be some complications to make `BPL_*` work. The health checker isn't going to natively understand them, so you'd need an exec.d or you'd need to tightly couple the health checker to the buildpack (ie. the health checker would need to look at `BPL_...`). I'm also to sure how you'd get around the limitations exec.d where it cannot modify the process types. Essentially, you'd need a health checker that is fully configurable via env variables. We could do that to `thc`, but it would make using other off-the-shelf health checkers harder. In addition, if we require the health checker to be configurable via env variables then we could probably just not use `BPL_*` to begin with and use the direct health checker config.

Besides that, there is a backdoor built-in here that would allow the user to override the health check and that is to just invoke the health checker directly, instead of through the `/cnb/process/health-check` launcher. The main benefit of using the launcher is so the operator doesn't need to know how to run the health checker, but if the operator wants to override the settings at runtime then it seems reasonable to expect them to look up the arguments on how to run the health checker. The only inconvenience would be that they'd need the full path to the health checker or to invoke the launcher so that `$PATH` is set up correctly.

dmikusa-pivotal: >The other option would be to configure the thc via runtime environment variables (with defaults set the buildpack).

What would be the path to make this work in a similar way to the use of args. Right now, we can bake the args into the image, so it's just a single command to run, `/cnb/process/health-check`. 

You can't set env variables on a process type, so while we could add a process type, it would have less utility. The user would need to configure the env variables still. 

The only way I can think to bake in the env variables so that there is a default "it just works" scenario is by pairing this with the environment-variables buildpack. Is that what you were thinking too? or am I overlooking a better way to do that?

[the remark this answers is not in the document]
dmikusa-pivotal: That's a fair criticism. I think the rationale for positional args here is this:

1. I wanted the smallest static binary as possible, so not pulling in a library to parse arguments helps keep the size as low as possible. Right now, it's super simple to pull the positional arguments.

2. If we take the route of having support for multiple health checkers in the buildpack, that keeps things flexible as well. We can have `thc` essentially be feature complete. It is just doing HTTP health checks, it does that, so it's done (maybe the method should be configurable? support for HEAD?). We can then add another different checker for TCP, a different one for gRPC, a different one for RSocket, or whatever else we need to support.

If all else fails, we can expose additional functionality through env variable settings, rather than continuing to add arguments. It might even make sense to do this with PORT & PATH, not that we strictly need it at the moment.

text/utilities/0000-health-checker-buildpack.md
```diff
@@ -0,0 +1,106 @@
...
+## Rationale and Alternatives
+
+1. A user can create a custom stack and include a tool like `curl` or `wget`. This is a lot of additional work, plus `curl` and `wget` are overkill for a simple HTTP health check. To reduce footprint, they are not recommended.
+
+## Implementation
+
+The new Health Checker buildpack will participate if either of the following conditions are met
+
+* A user sets `BP_HEALTH_CHECKER_ENABLED=true`
+* An upstream buildpack requests `hc` in the build plan.
```
ryanmoran, line 33: How would this work if this buildpack appears at the end of an order group?

dmikusa-pivotal: I don't think it would. The primary plan for Paketo buildpacks is to use the env variable. This build plan part is mostly just to make the buildpack flexible for potential other use cases. If that's additional noise in the RFC, we can delete this line.

text/utilities/0000-health-checker-buildpack.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Summary
+
+Introduce a Health Checker buildpack. This buildpack will provide access to one or more health checkers, where a health checker is a small binary that will be installed into the container image for the purpose of validating the health of the runnning image.
+
+By default, we'll provide a tiny HTTP health checker which can be used to probe a port and path. An HTTP 200 OK response returns exit code 0 and a non-2xx response returns 1.
+
+The Health Checker buildpack would be maintained by the existing [Utilities team](https://github.com/orgs/paketo-buildpacks/teams/utilities). It uses libpak & libcnb.
+
+## Motivation
```
ForestEckhardt, line 11: I am not sure that I am understanding a motive or use for this beyond something like integration testing an image with no app on it. I would love another use case for this if you have one at hand.

dmikusa-pivotal: Health check is a feature of your platform, like K8s or Cloud Foundry. Support from platform to platform varies. For example, K8s has HTTP, TCP and process health checks (it's getting/got gRPC health checks). What you get with other platforms will vary.

The primary focus for this is when running with Docker or Docker Compose. Docker only provides process health checks. Anything other health checks are done through a binary that must be run and return exit 0 (OK), 1 (FAIL). That is typically `curl` or `wget` for HTTP health checks, but those binaries are not in our Paketo base/tiny images nor do we want them to be, it adds a lot of surface area for CVEs.

The motivation here is to have a buildpack that provides custom-purpose health check binaries that can be used in Docker or on other platforms where health check support is not as extensive.

text/utilities/0000-health-checker-buildpack.md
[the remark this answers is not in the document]
ekcasey: > I wanted the smallest static binary as possible, so not pulling in a library to parse arguments helps keep the size as low as possible. Right now, it's super simple to pull the positional arguments.

We could use the [flag package](https://pkg.go.dev/flag) from the go standard library which is pretty lean, if not as flexible/pretty as something like cobra

ForestEckhardt (approved):

text/utilities/0000-health-checker-buildpack.md
dmikusa-pivotal: Excellent. Let me make some adjustments with this in mind.

ryanmoran (approved):

dmikusa-pivotal:
@ekcasey I have cleaned up & resolved the conversations here. I've incorporated the feedback and your comments into the RFC. The health checker configuration now all happens via env variable instead of arguments, which simplifies a few things, so that was a nice win.

I think this is ready for your re-review.

text/utilities/0000-health-checker-buildpack.md
```diff
@@ -36,66 +36,62 @@ When conditions are met, the buildpack will do the following:
...
+  * If any known environment variables for [supported health checkers](#default-health-checker) are set at build time, they will be set as process specific environment variables on the `health-check` process type at the `.default` level.
 
 The Health Checker buildpack will optionally be added to the language family buildpacks that wish to include support for it. The primary targets will be the Java and Rust buildpacks. The buildpack should be added reasonably close to the end of an order group for a given language family.
 
-Other buildpacks in the order group may influence the configuration of the the health checker by modifying the `BP_HEALTH_CHECKER_*` environment variables. It is recommended that buildpacks use the `.default` setting when doing this so that any user specified options will have the highest precedence.
+Other buildpacks in the order group may influence the configuration of the the health checker by modifying the `BP_HEALTH_CHECKER_DEPENDENCY` or health checker specific environment variables. It is recommended that buildpacks use the `.default` setting when doing this so that any user specified options will have the highest precedence.
```
ekcasey, line 43: fwiw at build time (so for `BP_HEALTH_CHECK_DEPENDENCY`) user provided env vars always win. But I still like the strategy of recommending `.default` aesthetically

At runtime `.default` vs `.override` will matter.

text/utilities/0000-health-checker-buildpack.md
```diff
@@ -36,66 +36,62 @@ When conditions are met, the buildpack will do the following:
...
+    health-check: thc (direct)
     ```
 
-2. You can run this with `docker run --health-cmd '/cnb/process/health-check' --health-interval 5s --health-timeout 2s -it apps/maven`. You could call the health check binary directly, but as you can have different health check dependencies that get installed by the buildpack, using the process type gives you more generic way to call the health check. It also allows for the image to bake in the health check arguments, rather than require the operator to set them. Plus, an operator could always call the binary directly if they wanted.
+2. You can run this with `docker run --health-cmd '/cnb/process/health-check' --health-interval 5s --health-timeout 2s -it apps/maven`. You could call the health check binary directly, but as you can have different health check dependencies that get installed by the buildpack, using the process type gives you more generic way to call the health check. It also allows for the image to bake in the health check configuration, rather than require the operator to set it. Plus, an operator can always override the configuration by setting the appropriate environment variables at runtime.
```
ekcasey, line 94:
```suggestion
2. You can run this with `docker run --health-cmd 'health' --health-interval 5s --health-timeout 2s -it apps/maven`. You could call the health check binary directly, but as you can have different health check dependencies that get installed by the buildpack, using the process type gives you more generic way to call the health check. It also allows for the image to bake in the health check configuration, rather than require the operator to set it. Plus, an operator can always override the configuration by setting the appropriate environment variables at runtime.
```

More extreme bikeshedding (feel free to ignore): We do put the processes on the front of the path and I wonder if we should be documenting the easier usage (w/o the absolute path) in more place

ekcasey (approved):

thitch97 (approved):

text/utilities/0000-health-checker-buildpack.md
```diff
@@ -36,66 +36,62 @@ When conditions are met, the buildpack will do the following:
...
         $BP_HEALTH_CHECKER_ENABLED     true                   contributes a health checker if enabled
-        $BP_HEALTH_CHECKER_THC_ARGS    8080 /actuator/health  arguments passed to tiny-health-checker, if used
-      Tiny Health Checker 0.4.0: Contributing to layer
-        Downloading from https://github.com/dmikusa-pivotal/tiny-health-checker/releases/download/v0.4.0/thc-x86_64-unknown-linux-musl
+      Tiny Health Checker 0.5.0: Contributing to layer
+        Downloading from https://github.com/dmikusa-pivotal/tiny-health-checker/releases/download/v0.5.0/thc-x86_64-unknown-linux-musl
         Verifying checksum
         Copying from /tmp/ce75bb97209981e03bf7e8aa52e2bfab78a50a44c7ed1787f4ace212711d61e5/thc-x86_64-unknown-linux-musl to /layers/paketo-buildpacks_health-checker/thc/bin
+        Writing env.launch/health-check/THC_PATH.default
+        Writing env.launch/health-check/THC_PORT.default
```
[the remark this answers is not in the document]
dmikusa-pivotal: I had that thought as well.

text/utilities/0000-health-checker-buildpack.md
dmikusa-pivotal: ✅  Ok, also good to know. Thanks

fg-j:
From WG 7/19/22 :
Ping @utilities-maintainers last call for votes!

joshuatcasey:
2022-07-26 Working Group notes: ready to merge by @dmikusa-pivotal

joshuatcasey:
@dmikusa-pivotal I noticed that this was merged with number `0000`, should that be `0007`?

dmikusa-pivotal:
@joshuatcasey - See https://github.com/paketo-buildpacks/rfcs/pull/226.

PR #185, opened 2022-04-21
RFC: Extend .Net SDK

ForestEckhardt:
[Readable](https://github.com/paketo-buildpacks/rfcs/blob/extend-dotnet-sdk/text/dotnet-core/rfcs/0004-extend-dotnet-sdk.md)

### DIFF
--- a/text/dotnet-core/rfcs/0004-extend-dotnet-sdk.md
+++ b/text/dotnet-core/rfcs/0004-extend-dotnet-sdk.md
@@ -110,14 +110,15 @@ itself.
 ## Unresolved Use Cases
 During the course of the authoring this RFC a use case was brought to our
 attention where users can have a project that requires 2 different .NET Runtime
-versions at build time to compile correctly. The current version of the .NET
-Core buildpack does not support this use case, since only one version of the
-.NET Runtime is installed during a given build. This proposed change to the
-.NET SDK buildpack **does not** fully support this use case. Multi-runtime apps
-may work for online builds, since the .NET SDK can download needed compile-time
-dependencies. However, offline builds will fail because the SDK cannot download
-the missing .NET Runtime dependencies. This is considered out of scope for this
-RFC, but could be supported in the future.
+versions at build time to compile correctly, an example of this type of app can 
+be seen [here](https://github.com/macsux/multi-version-dotnet-project). 
+The current version of the .NET Core buildpack does not support this use case, 
+since only one version of the .NET Runtime is installed during a given build. 
+This proposed change to the.NET SDK buildpack **does not** fully support this 
+use case. Multi-runtime apps may work for online builds, since the .NET SDK can 
+download needed compile-timedependencies. However, offline builds will fail 
+because the SDK cannot download the missing .NET Runtime dependencies. This 
+is considered out of scope for this RFC, but could be supported in the future.
 
 ## Source Material
 
--- a/text/dotnet-core/rfcs/0004-extend-dotnet-sdk.md
+++ b/text/dotnet-core/rfcs/0004-extend-dotnet-sdk.md
@@ -2,9 +2,8 @@
 
 ## Proposal
 
-Reconfigure the .Net SDK dependency to be identical to the one that can be
-downloaded from Microsoft which includes the runtimes that the SDK is dependent
-on to run.
+Replace the current .Net SDK dependency with the one that can be downloaded
+from Microsoft which includes the runtimes that the SDK is dependent on to run.
 
 ## Motivation
 As of [.Net SDK version ~6.0.100 publish no longer properly functions for
@@ -69,13 +68,17 @@ would obviously use the cached resource.
 
 ## Implementation
 
-When constructing the .Net SDK dependency the pipeline should no longer strip
-any files out of the given dependency. It should instead take all of the files
-from the Microsoft release and compress them as we are doing for our current
-.Net SDK dependency. This will allow us to install a fully realized .Net SDK
-into a build only layer meaning that none of the extra SDK libraries or the
+The .Net SDK dependency referenced in the buildpack should be identical to the
+one provided by Microsoft. This will allow us to install a fully realized .Net
+SDK into a build only layer meaning that none of the extra SDK libraries or the
 additional runtimes will be present during build. The runtime buildpacks will
 still construct a runtime .Net root for themselves.
+
+This change means that the .Net Publish buildpack will no longer need to
+require the .Net Runtime and .Net ASPNetcore during build as both of those
+runtimes will now be included in the .Net SDK artifact. As a result the .Net
+Runtime and .Net ASPNetcore will no longer need to make their locations and
+other environment variables available during build.
 
 ## Alternatives
 There are two alternatives that have been discussed among the .Net Maintainers
@@ -97,9 +100,21 @@ We could pass information of the location of all of the dependencies to the
 Dotnet Publish buildpack that it could use to construct a Hive directory for
 itself.
 
-####
+#### Cons
 - Lots of unnecessary file operations.
 - Feels like the most opaque option.
+
+## Unresolved Use Cases
+During the course of the authoring this RFC a use case was brought to our
+attention where users can have a project that requires 2 different runtime
+versions to compile correctly. This means that our current solution will not
+resolve this issue for offline buildpack builds but our current buildpack does
+not cover this use case because we download only 1 corresonding SDK and if it
+cannot reach out to the internet it cannot download nugget packages that is
+requires to be backwards compatible. I still beleive that this is the simplest
+solution to a problem that is breaking out buildpack currently. I would,
+however, like to raise that this is a workflow that should be addressed in the
+future.
 
 ## Source Material
 
--- a/text/dotnet-core/rfcs/0004-extend-dotnet-sdk.md
+++ b/text/dotnet-core/rfcs/0004-extend-dotnet-sdk.md
@@ -54,7 +54,7 @@ and the size of the overall offline bundle right now versus the proposed state.
 
 
 As you can see the overall size of the offline buildpack has decreased by
-~90MB, however the overall size of each dependency has increased around
+~90MB (~30%), however the overall size of each dependency has increased around
 ~30-50%. Based on data from
 [Speedtest](https://www.speedtest.net/global-index), with the world average at
 the time of writing being 62.52 Mbps, the average increase in download times
--- a/text/dotnet-core/rfcs/0004-extend-dotnet-sdk.md
+++ b/text/dotnet-core/rfcs/0004-extend-dotnet-sdk.md
@@ -1,45 +1,46 @@
-# Extend .Net SDK Dependency
+# Extend .NET SDK Dependency
 
 ## Proposal
 
-Replace the current .Net SDK dependency with the one that can be downloaded
+Replace the current .NET SDK dependency with the one that can be downloaded
 from Microsoft which includes the runtimes that the SDK is dependent on to run.
 
 ## Motivation
-As of [.Net SDK version ~6.0.100 publish no longer properly functions for
-certain types of .Net
-apps](https://github.com/paketo-buildpacks/dotnet-core/issues/670). Upon
-testing we were able to get the buildpack to function again if we used an .Net
-SDK dependency that still had the runtimes libraries packaged with it. After
-further investigation we discovered that the removal of those local runtimes
-was what was causing the failure, leading us to believe that the .Net SDK no
-longer follows the runtime search path that it used to.
+As of .NET SDK version ~6.0.100, publish no longer properly functions for
+certain types of .NET apps. See
+https://github.com/paketo-buildpacks/dotnet-core/issues/670. Upon testing, we
+were able to get the buildpack to function again if we used a .NET SDK
+dependency that still had the runtime's libraries packaged with it. After
+further investigation, we discovered that the removal of those local runtimes
+caused the failure, leading us to believe that the .NET SDK no longer follows
+the runtime search path that it used to.
 
 It is becoming ever more apparent to me that Microsoft is dead set on not
-having flexibility of file location during `dotnet publish` therefore I think
-that we should capitulate to Microsoft and create an expected .Net Hive
-installation, which is a layer where the file structure matches the layout that
-is present in the official .Net SDK release from Microsoft (includes the SDK,
-runtime, and ASP.NET). We can accomplish this installation structure by using
-the .Net SDK dependency unaltered from Microsoft, as it stands currently we
-remove the included runtime and ASP.NET files in an effort to not have
-duplicated files. I think that this will make the buildpack less brittle to
-change in the future and will construct a build environment that is more
-similar to what .Net developers actually have installed on their machines and
-what Microsoft has in their Docker images.
+having flexibility of file location during `dotnet publish`. Therefore I think
+that we should capitulate to Microsoft and create an expected .NET Hive
+installation. The .NET Hive is the expected structure and contents of the
+official .NET SDK release provided by Microsoft. It includes the .NET SDK, .NET
+Runtime, and ASP.NET. We can accomplish this installation structure by using
+the .NET SDK dependency unaltered from Microsoft. As it stands currently, we
+remove the .NET runtime and ASP.NET files included in the SDK distribution in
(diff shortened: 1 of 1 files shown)


text/dotnet-core/rfcs/0004-extend-dotnet-sdk.md
```diff
@@ -0,0 +1,50 @@
+# Extend .Net SDK Dependency
+
+## Proposal
+
+Reconfigure the .Net SDK dependency to be identical to the one that can be downloaded from Microsoft which includes the runtimes that the SDK is dependent on to run.
+
+## Motivation
+As of .Net SDK version ~6.0.100 publish now longer properly functions for certain types of .Net apps. Upon testing we were able to get the buildpack to function again if we used an .Net SDK dependency that still had the runtimes libraries packaged with it. After further investigation we discovered that the removal of those local runtimes was what was causing the failure, leading us to believe that the .Net SDK no longer follows the runtime search path that it used to.
```
sophiewigmore, line 8: nit: can you link out to the issue that has a lot of details about the failure? https://github.com/paketo-buildpacks/dotnet-core/issues/670

text/dotnet-core/rfcs/0004-extend-dotnet-sdk.md
```diff
@@ -0,0 +1,50 @@
...
+
+It is becoming ever more apparent to me that Microsoft is dead set on not having flexibility of file location during `dotnet publish` therefore I think that we should capitulate to Microsoft and create an expected .Net Hive installation. I think that this will make the buildpack less brittle to change in the future and will construct a build environment that is more similar to what .Net developers actually have installed on their machines and what Microsoft has in their Docker images.
```
sophiewigmore, line 10: nit: can you link to .Net Hive docs you looked at?  It's a concept that is new to the buildpacks

sophiewigmore: It would also be helpful if you described what a .Net Hive is, it's not clear that it includes runtime and ASP.Net

text/dotnet-core/rfcs/0004-extend-dotnet-sdk.md
```diff
@@ -0,0 +1,50 @@
...
+
+There are some additional advantages to using the .Net SDK dependency with its included runtimes chief among them being that newer versions of the .Net SDK can compile any version of an app for any given version of a runtime that is in the same `major.minor` family. This would mean that we would only have to support one version of the .Net SDK for every minor version as opposed to the two we currently support. On top of that we would no longer need to align our launch time runtime versions one-to-one with a single .Net SDK meaning that our version resolution logic would be vastly simplified. Also because we would only need to support one .Net SDK, users that are building offline capable buildpacks would be able to build a significantly smaller offline buildpack. Below are tables of each of the sizes of each dependency and the size of the overall offline bundle right now versus the proposed state.
```
sophiewigmore, line 12: `
```suggestion
There are some additional advantages to using the .Net SDK dependency with its included runtimes chief among them being that newer versions of the .Net SDK can compile any version of an app for any given version of a runtime that is in the same `major.minor` family. This would mean that we would only have to support one version of the .Net SDK for every major version as opposed to the two we currently support. On top of that we would no longer need to align our launch time runtime versions one-to-one with a single .Net SDK meaning that our version resolution logic would be vastly simplified. Also because we would only need to support one .Net SDK, users that are building offline capable buildpacks would be able to build a significantly smaller offline buildpack. Below are tables of each of the sizes of each dependency and the size of the overall offline bundle right now versus the proposed state.
```

right?

text/dotnet-core/rfcs/0004-extend-dotnet-sdk.md
ForestEckhardt: That is fair, I don't know that it is an official term but I can put in an explanation.

ForestEckhardt: Could you take another look and see if I addressed your question. I gave a small expaination.

text/dotnet-core/rfcs/0004-extend-dotnet-sdk.md
```diff
@@ -0,0 +1,102 @@
...
+### .Net SDK + Runtimes XZ Compressed
+| Version | Size  |
+|---------|-------|
+| 3.1.418 | 78MB  |
+| 6.0.202 | 114MB |
+| Total   | 192MB |
+
+
+As you can see the overall size of the offline buildpack has decreased by
+~90MB, however the overall size of each dependency has increased around
```
fg-j, line 57:
```suggestion
~90MB (~30%), however the overall size of each dependency has increased around
```
nit: added the percentage size decrease

fg-j:
@ForestEckhardt Do you have a running proof-of-concept of this sitting somewhere? If possible, it'd be great to expose that so folks can play around with it.

fg-j:
All in all I support this proposal because I think it aligns with Paketo's buildpacks [philosophy](https://blog.paketo.io/posts/buildpack-philosophy-part-1/) around "Meet[ing] language-ecosystem expectations" rather than exploiting undocumented quirks (which it seems like the .NET buildpacks are currently doing).

Moreover, Docker's official [documentation](https://docs.docker.com/samples/dotnetcore/) on how to containerize a .NET app recommends a multistage build that pulls the SDK image for build and aspnet image for run. However, It seems that the official ASP.NET Core and .NET Runtime images' layers are subsets of the .NET SDK image. This means there's no "double download" during a docker multistage build.

I wonder if there's a way that we can set up the dependencies so that we avoid double-downloading the .NET Runtime and ASP.NET parts of the SDK dependency.

ForestEckhardt:
> @ForestEckhardt Do you have a running proof-of-concept of this sitting somewhere? If possible, it'd be great to expose that so folks can play around with it.

`dotnet-core-runtime`, `dotnet-core-aspnet`, `dotnet-core-sdk`, and ` dotnet-publish` all have a `publish-spike` branch. If you build all of those buildpacks from that branch you can get an idea of what this new procedure would look like.

text/dotnet-core/rfcs/0004-extend-dotnet-sdk.md
```diff
@@ -0,0 +1,102 @@
...
+SDK dependency that still had the runtimes libraries packaged with it. After
+further investigation we discovered that the removal of those local runtimes
+was what was causing the failure, leading us to believe that the .Net SDK no
+longer follows the runtime search path that it used to.
+
+It is becoming ever more apparent to me that Microsoft is dead set on not
+having flexibility of file location during `dotnet publish` therefore I think
+that we should capitulate to Microsoft and create an expected .Net Hive
+installation, which is a layer where the file structure matches the layout that
+is present in the official .Net SDK release from Microsoft. I think that this
```
sophiewigmore, line 21:
```suggestion
that we should capitulate to Microsoft and create an expected .Net Hive
installation, which is a layer where the file structure matches the layout that
is present in the official .Net SDK release from Microsoft (includes the SDK, runtime, and ASP.NET). I think that this
```

sophiewigmore:
> I wonder if there's a way that we can set up the dependencies so that we avoid double-downloading the .NET Runtime and ASP.NET parts of the SDK dependency.

@fg-j not a bad idea- is this something that has been explored at all?

ForestEckhardt:
> * This involves "surgery" on the .NET SDK release artifact -- we lose the supportability benefit of shipping official distributions of the runtime and ASP.NET

I think that this is my biggest concern out of this request. The whole reason that I am purposing this RFC is to avoid having to do surgery in order to hopefully make the buildpack more predictable and in line with what Microsoft supports.

fg-j:
The concern around double-downloading would be mitigated if CNB supported some sort of asset caching across builds. Today, if two apps are built on the same node and they both require the same dependency (e.g. the .NET runtime), each build will download the artifact separately, completely unaware that there are now two copies of an identical asset on the node.

If, instead, CNB platforms had some way of caching build assets across builds, only one app build would reach out to the internet for the dependency source. The other would simply use the cached version of the asset already present on the node. [This draft RFC](https://github.com/buildpacks/rfcs/blob/652520c8f4d9192f08cf230e0da545bbefec424e/text/0000-writable-asset-cache.md) in the upstream CNB project outlines the problem and a proposed solution in more depth.

Perhaps the double-download issue is better thought of as a limitation of CNB platforms rather than the .NET buildpack. After all, Docker builds rely on cross-build layer caching to solve this performance issue.

text/dotnet-core/rfcs/0004-extend-dotnet-sdk.md
```diff
@@ -0,0 +1,102 @@
+# Extend .Net SDK Dependency
+
+## Proposal
+
+Reconfigure the .Net SDK dependency to be identical to the one that can be downloaded from Microsoft which includes the runtimes that the SDK is dependent on to run.
```
ekcasey, line 5: What are the barriers to directly using the .Net SDK distributed by Microsoft?

text/dotnet-core/rfcs/0004-extend-dotnet-sdk.md
```diff
@@ -0,0 +1,102 @@
...
+testing we were able to get the buildpack to function again if we used an .Net
+SDK dependency that still had the runtimes libraries packaged with it. After
+further investigation we discovered that the removal of those local runtimes
+was what was causing the failure, leading us to believe that the .Net SDK no
+longer follows the runtime search path that it used to.
+
+It is becoming ever more apparent to me that Microsoft is dead set on not
+having flexibility of file location during `dotnet publish` therefore I think
+that we should capitulate to Microsoft and create an expected .Net Hive
+installation, which is a layer where the file structure matches the layout that
```
ekcasey, line 20: Some more details here would be appreciated, either the proposed file structure of the layer (compared to the current state), or an external link with similar information.

macsux:
Just adding my 2c on this.

.NET SDK can compile ANY backwards compatible version of code as long as reference assemblies are included. Reference assemblies are "stubs" of real APIs that are being targeted by the code, but the are void of implementation. It's basically an interface.

These reference assemblies are distributed as nuget packages, and are placed in nuget cache (by default `~/.nuget` folder but can be overridden via env var). If the SDK encounters code that requires a reference package (cuz it's older then current installed runtimes), it will look for it in nuget cache and try to restore it from internet if not present.

At runtime, those reference assemblies are substituted by real assemblies distributed as part of `runtime` package.

Each .NET SDK comes bundled with the same version runtime. This is needed because a lot of SDK tooling used during compilation is written in .NET and depends on the presence of the said runtime in order to execute. 

We can confirm that this works by creating a simple .NET 3.1 app and using official .NET 6 SDK docker image to containerize:

```
docker run -it --rm -v C:\temp\netapp31\netapp31:/app -v nugetcache:/root/.nuget mcr.microsoft.com/dotnet/sdk:6.0
cd /app
dotnet build
```

this populates the nuget cache with the following packages
```
root@591bd7eb8337:~/.nuget/packages# tree -d
.
|-- microsoft.aspnetcore.app.ref
|   `-- 3.1.10
|       |-- data
|       `-- ref
|           `-- netcoreapp3.1
|-- microsoft.netcore.app.host.linux-x64
|   `-- 3.1.24
|       `-- runtimes
|           `-- linux-x64
|               `-- native
`-- microsoft.netcore.app.ref
    `-- 3.1.0
        |-- data
        `-- ref
            `-- netcoreapp3.1
```

We can confirm that the SDK can now compile this even with no internet access by remounting the cached nuget volume
```
docker run -it --rm -v C:\temp\netapp31\netapp31:/app -v nugetcache:/root/.nuget --network none mcr.microsoft.com/dotnet/sdk:6.0
```

We can confirm that the .NET 6 SDK is still able to build this app (but it will fail if you delete the `~/.nuget/packages/` folder and it can't redownload them).

Therefore it is my opinion and recommendation that the best approach is to include major versions of SDK in the buildpack, and pre-populate nuget cache with reference assemblies targeting old versions of runtimes, even those that are out of support. The reference assemblies for each .net runtime version in unpacked format in `.nuget` folder is ~18MB. However, this includes the original .nupkg file PLUS all the extracted content from it.  If space is of concern, only the nuget packages can be included in the buildpack and used to repopulate the nuget cache folder by extracting them in the appropriate structure (they are just zip files with additional metadata, plus an [additional metadata file that has sha256 of the original nuget it was extracted from](https://github.com/NuGet/Home/wiki/Nupkg-Metadata-File). This would reduce size to about ~6mb for a given runtime.

ForestEckhardt:
@macsux I appreciate this deep dive into the inner workings of the SDK dependency. I think that my biggest hang up is that in this RFC we are trying to away from customizing the artifact and more towards consuming the artifacts that Microsoft uses directly. My hope is that by using the artifacts as is from Microsoft we will have make our buildpack more consistent as we will align with the intended Microsoft experience. We have struggled in the passed trying to make our buildpacks modular in ways the Microsoft does not expect/intended and it has made some of the behavior in our buildpacks flakey. This is a step towards trying to provide an expected .Net experience for developers with as little magic from the buildpacks as possible.

text/dotnet-core/rfcs/0004-extend-dotnet-sdk.md
ekcasey: thanks for clarifying that we are trying to use the version distributed by Microsoft

text/dotnet-core/rfcs/0004-extend-dotnet-sdk.md
```diff
@@ -97,10 +100,22 @@ We could pass information of the location of all of the dependencies to the
...
 - Lots of unnecessary file operations.
 - Feels like the most opaque option.
 
+## Unresolved Use Cases
+During the course of the authoring this RFC a use case was brought to our
+attention where users can have a project that requires 2 different runtime
+versions to compile correctly. This means that our current solution will not
+resolve this issue for offline buildpack builds but our current buildpack does
+not cover this use case because we download only 1 corresonding SDK and if it
+cannot reach out to the internet it cannot download nugget packages that is
```
sophiewigmore, line 113:
```suggestion
cannot reach out to the internet it cannot download nuget packages that is
```

text/dotnet-core/rfcs/0004-extend-dotnet-sdk.md
```diff
@@ -97,10 +100,22 @@ We could pass information of the location of all of the dependencies to the
...
+requires to be backwards compatible. I still beleive that this is the simplest
```
sophiewigmore, line 114:
```suggestion
required to be backwards compatible
```

text/dotnet-core/rfcs/0004-extend-dotnet-sdk.md
```diff
@@ -97,10 +100,22 @@ We could pass information of the location of all of the dependencies to the
...
+solution to a problem that is breaking out buildpack currently. I would,
```
sophiewigmore, line 115:
```suggestion
solution to a problem that is breaking the buildpack currently. I would,
```

text/dotnet-core/rfcs/0004-extend-dotnet-sdk.md
```diff
@@ -69,14 +68,18 @@ would obviously use the cached resource.
 
 ## Implementation
 
-When constructing the .Net SDK dependency the pipeline should no longer strip
-any files out of the given dependency. It should instead take all of the files
-from the Microsoft release and compress them as we are doing for our current
-.Net SDK dependency. This will allow us to install a fully realized .Net SDK
-into a build only layer meaning that none of the extra SDK libraries or the
+The .Net SDK dependency referenced in the buildpack should be identical to the
+one provided by Microsoft. This will allow us to install a fully realized .Net
```
sophiewigmore, line 72: For clarity with the dependency work happening: this means that the URI (and source URI) of the SDK dependency will be the Microsoft upstream URI, rather than the URI to a dependency that we host in the dep-server?

ForestEckhardt: I was trying to leave that implementation ambiguous for the time being cause I imagine it will start out as a reference to something the is hosted on dep-server and then move to the URI from Microsoft. I can state that it will still be on the dep-server.

fg-j: I think it's ok to leave this ambiguous. This RFC is about what the dependency _is_, not where it's hosted.

fg-j (approved):

sophiewigmore (approved):

PR #194, opened 2022-05-03
Support for Yarn Berry

thitch97:
[Readable](https://github.com/paketo-buildpacks/rfcs/blob/yarn2/text/nodejs/0014-yarn-berry.md)

## Checklist

### DIFF
--- a/text/nodejs/0014-yarn-berry.md
+++ b/text/nodejs/0014-yarn-berry.md
@@ -50,6 +50,17 @@ packages in the build plan across the board. Since Berry explicitly excludes
 `node_modules`, however, it seems reasonable to instead use a generic term for
 node packages such as `node_pkgs`.
 
+The buildpacks affected by this change to the `node_modules` build plan dependency are:
+
+- `npm-install`
+- `yarn-install`
+- `npm-start`
+- `yarn-start`
+- `node-start`
+- `node-run-script`
+- `node-module-bom`
+- `rails-assets`
+
 Though the Berry CLI is a complete departure from Classic, the latter may still
 be used as a global orchestrator for per-project Berry installations. The
 [recommended
--- a/text/nodejs/0014-yarn-berry.md
+++ b/text/nodejs/0014-yarn-berry.md
@@ -39,7 +39,7 @@ support it has steadily increased.
 
 `yarn-berry-install`
 
-Provides: `yarn_packages` OR `node_modules` Requires: `node`, during `build`
+Provides: `yarn_packages` OR `node_modules` Requires: `node`, during `build` and `launch`
 
 The buildpack does not require `yarn` since the Berry release itself is meant
 to be checked into version control under `.yarn/releases`. Node v16.10 and
@@ -169,13 +169,16 @@ The buildpack should run `yarn install` in the following cases:
 - If an application uses `node_modules`, which may be determined by the
   presence of a `node_modules` directory or through `.yarnrc.yml` configuration.
 
-- If there is no `.yarn/cache` in the working directory.
+- If there is no local cache (`.yarn/cache` by default, but can be configured) in
+  the working directory, whether `.pnp.cjs` exists or not.
+
+- If there is no `.pnp.cjs` in the working directory, even if a local cache exists.
 
 - If the `yarn.lock` has changed.
 
 The buildpack should NOT run `yarn install` when:
 
-- Both a `.pnp.cjs` file and `.yarn/cache` are present in the working directory.
+- Both a `.pnp.cjs` file and a local cache are present in the working directory.
 
 
 The buildpack will store the Yarn cache in a layer and configure Yarn to
@@ -186,7 +189,7 @@ existing Yarn Install buildpack, with accommodations for the CLI changes.
 ### Layer Reuse 
 
 The Yarn Install buildpack currently uses a checksum of the output from `yarn
-config list` and the contents of `yarn.lock` to decide whether to reuse cached layers.
+config list` and the contents of `yarn.lock` to decide whether to reuse package layers.
 `yarn config list` has been removed in Berry, but the new `yarn info` command, which
 provides information about the project's direct and transitive dependencies,
 should work well as a replacement.
@@ -224,7 +227,7 @@ start command in the Yarn Start buildpack may need to be modified.
 ## Rationale and Alternatives
 
 - Include logic for Yarn Berry workflows in the existing `yarn-install`
-  buildpack. 
+  buildpack.
 
   Berry includes breaking changes to the `yarn` CLI used by the buildpack to
   install dependencies. The changes to CLI options are such that at the very
@@ -239,9 +242,32 @@ start command in the Yarn Start buildpack may need to be modified.
 
 ## Unresolved Questions and Bikeshedding
 
+
+- Should this buildpack require `yarn`?
+
+Classic Yarn can be used as a global orchestrator of per-project Yarn
+installations. Corepack is shipped with Node v16.10 or greater and is the
+recommended tool for managing these installations. It would be ideal to conform
+to the recommended workflow through Corepack, but doing so adds a bit of
+complexity for offline builds. Corepack  attempts to download Yarn releases
+directly from the registry, which it cannot do in an offline environment. There
+are [workarounds](https://nodejs.org/api/corepack.html#offline-workflow) for
+this, but they rely on a global [corepack
+cache](https://github.com/nodejs/corepack#corepack-prepare--nameversion) which
+would have to be replicated within the build/run container.
+
+- What should happen if the user selects a version of Node < 16.10?
+
+If the ultimate decision is not to require `yarn` by default (the subject of
+the previous item), the buildpack could optionally require `yarn` in this case.
+This would result in a global install of Classic Yarn which can be used as an
+orchestrator.
+
+- Could slices be utilized to optimize layer reuse for .yarn/releases, local cache, etc.?
+
 - ~~Should the buildpack support building Yarn Berry applications which use
   `node_modules` as well as Plug 'n' Play?~~ Yes. It seems that some users have
-  foregone Plug 'n' Play and are using Yarn Berry with the node_modules linker.
+  foregone Plug 'n' Play and are using Yarn Berry with the `node_modules` linker.
 
 - ~~Yarn Berry supports setting a global cache. Should we hold off on
   supporting this workflow until it is requested?~~ Yes. This feature is
--- a/text/nodejs/0014-yarn-berry.md
+++ b/text/nodejs/0014-yarn-berry.md
@@ -237,12 +237,13 @@ start command in the Yarn Start buildpack may need to be modified.
   supported. For this reason, it is even more prudent to keep the logic
   separate to facilitate the eventual transition.
 
-- Do nothing.
-
 ## Unresolved Questions and Bikeshedding
 
-- Should the buildpack support building Yarn Berry applications which use
-  `node_modules` as well as Plug 'n' Play?
+- ~~Should the buildpack support building Yarn Berry applications which use
+  `node_modules` as well as Plug 'n' Play?~~ Yes. It seems that some users have
+  foregone Plug 'n' Play and are using Yarn Berry with the node_modules linker.
 
-- Yarn Berry supports setting a global cache. Should we hold off on supporting
-  this workflow until it is requested?
+- ~~Yarn Berry supports setting a global cache. Should we hold off on
+  supporting this workflow until it is requested?~~ Yes. This feature is
+  valuable for the monorepo use-case but is not critical for the initial
+  implementation and can be added once there is demand.
--- a/text/nodejs/0014-yarn-berry.md
+++ b/text/nodejs/0014-yarn-berry.md
@@ -1,0 +1,248 @@
+# Support for Yarn "Berry" 
+
+## Summary
+
+The Paketo Node.js Buildpack supports building apps which use
+[Yarn](https://yarnpkg.com/) as a package/project management tool. Over the
+last few years, the developers of Yarn have released new major versions (v2 &
+v3), collectively called Yarn "Berry", which represent a completely new
+direction for managing dependencies in Node.js projects. Chief among the
+changes is the introduction of [Plug 'n
+Play](https://yarnpkg.com/features/pnp), a system which removes the reliance on
+`node_modules` and can eliminate the need to run `yarn install` at all. The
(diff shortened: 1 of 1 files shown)


text/nodejs/0014-yarn-berry.md
```diff
@@ -0,0 +1,249 @@
...
+- If an application uses `node_modules`, which may be determined by the
+  presence of a `node_modules` directory or through `.yarnrc.yml` configuration.
+
+- If there is no `.yarn/cache` in the working directory.
+
+- If the `yarn.lock` has changed.
+
+The buildpack should NOT run `yarn install` when:
+
+- Both a `.pnp.cjs` file and `.yarn/cache` are present in the working directory.
```
robdimsdale, line 178: Is it possible to have a `.pnp.cjs` file but no `.yarn/cache`? What about vice-versa - `.yarn/cache` but no `.pnp.cjs`? What should the buildpack do in those cases?

text/nodejs/0014-yarn-berry.md
```diff
@@ -0,0 +1,249 @@
...
+group since it detects on the presence of `.yarnrc.yml` and both app types
+will contain `yarn.lock` files.
+
+### Detection
+
+- `.yarnrc.yml` was created with the intention of making it easy to detect
+  whether a project uses Yarn Classic or Berry (Berry ignores `.yarnrc`
+  completely), and should be used for the detection of `yarn-berry-install`.
+
+### Build
```
robdimsdale, line 159: Will `Build` create a filesystem-based (i.e. syft-based) SBOM like `yarn-install` does today?

text/nodejs/0014-yarn-berry.md
```diff
@@ -0,0 +1,249 @@
...
+
+
+### API
+
+`yarn-berry-install`
+
+Provides: `yarn_packages` OR `node_modules` Requires: `node`, during `build`
+
+The buildpack does not require `yarn` since the Berry release itself is meant
+to be checked into version control under `.yarn/releases`. Node v16.10 and
```
robdimsdale, line 45: Will we have to validate that `yarn` is vendored under `.yarn/releases` - either during `Detect` or during `Build`? Some of the workflows below invoke `yarn` so if users do not provide it under `.yarn/releases` and this buildpack does not require the `yarn` buildpack then those workflows will fail.

Perhaps more concretely, is it possible to pass `Detect` (by validating that `.yarnrc.yml` exists) and then fail at run time because `.yarn/releases` does not exist?

text/nodejs/0014-yarn-berry.md
thitch97: In both cases, `yarn install`  would need to be run. The first case is implied above but could admittedly be made more explicit. I’ll update to include the second case as well.

text/nodejs/0014-yarn-berry.md
thitch97: Good question. Yes, syft [supports yarn berry](https://github.com/anchore/syft/pull/868). That should be called out here explicitly.

text/nodejs/0014-yarn-berry.md
thitch97: The build would fail since `corepack` (which manages per-project installations of yarn & comes bundled with Node) won’t be able to locate the release as specified by `.yarnrc.yml`.  This raises additional questions and there will be challenges supporting offline builds with this approach. I’ll outline those issues in `Unresolved Questions`.

text/nodejs/0014-yarn-berry.md
```diff
@@ -0,0 +1,277 @@
...
+
+
+## Implementation
+
+
+### API
+
+`yarn-berry-install`
+
+Provides: `yarn_packages` OR `node_modules` Requires: `node`, during `build` and `launch`
```
ryanmoran, line 42: Why does this require anything at launch? At launch, this buildpack isn't really doing anything other than making sure that there are some JavaScript files available on the filesystem. I'd expect that if someone wanted to run these files, they'd need to require `node` some other way, no?

text/nodejs/0014-yarn-berry.md
```diff
@@ -0,0 +1,277 @@
...
+
+- If an application uses `node_modules`, which may be determined by the
+  presence of a `node_modules` directory or through `.yarnrc.yml` configuration.
+
+- If there is no local cache (`.yarn/cache` by default, but can be configured) in
+  the working directory, whether `.pnp.cjs` exists or not.
+
+- If there is no `.pnp.cjs` in the working directory, even if a local cache exists.
+
+- If the `yarn.lock` has changed.
```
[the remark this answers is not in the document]
ryanmoran: Can you clarify what happens if there is a `.pnp.cjs` file and a local cache and the `yarn.lock` file has changed?

ryanmoran (changes requested):

text/nodejs/0014-yarn-berry.md
thitch97: You are correct, I've updated accordingly.

text/nodejs/0014-yarn-berry.md
```diff
@@ -0,0 +1,280 @@
...
+should work well as a replacement.
+
+### `--frozen-lockfile`
+
+The existing Yarn Install buildpack passes the `--frozen-lockfile` argument to
+every `yarn install` call, which will not update the `yarn.lock` file, and will
+fail if the `yarn.lock` file needs to be updated. As such, developers must run
+`yarn install` locally in order to update their app's `yarn.lock` file.
+`--frozen-lockfile` has been deprecated in favour of `--immutable`, which this
+buildpack will utilize.
```
aurelienbottazini, line 209: I am bit concerned by this.
I agree `--immutable` is a good idea in theory.
But I had strange cases where for an identical node and yarn version, `yarn install --immutable` would fail on some env and succeed on others. Very frustrating and hard to debug.

Should we instead not specify the option by default and let users either
- add `enableImmutableInstalls: false` to `.yarnrc.yml`
- or set with yarn env vars mechanism: `YARN_ENABLE_IMMUTABLE_INSTALLS`

aurelienbottazini, line 209: I am bit concerned by this.
I agree `--immutable` is a good idea in theory.
But I had strange cases where for an identical node and yarn version, `yarn install --immutable` would fail on some env and succeed on others. Very frustrating and hard to debug.

Should we instead not specify the option by default and let users either
- add `enableImmutableInstalls: false` to `.yarnrc.yml`
- or set with yarn env vars mechanism: `YARN_ENABLE_IMMUTABLE_INSTALLS`

thitch97: This is helpful feedback, thank you. I'm okay with making this configurable.

text/nodejs/0014-yarn-berry.md
```diff
@@ -0,0 +1,280 @@
...
+  Berry includes breaking changes to the `yarn` CLI used by the buildpack to
+  install dependencies. The changes to CLI options are such that at the very
+  least, the buildpack would need to decide which ones to use based on the type
+  of application being built. Trying to incorporate both workflows will
+  increase the scope of responsibility of the buildpack, likely increasing its
+  complexity at the expense of maintainability. Classic Yarn is now in
+  maintenance mode and, though it is still heavily used now (due to the
+  relatively slow adoption of Berry), its unclear how long it will be
+  supported. For this reason, it is even more prudent to keep the logic
+  separate to facilitate the eventual transition.
```
ryanmoran, line 246: I think it might be reasonable to still get Berry support in the existing buildpack if we just take some effort to keep the "Classic" and Berry parts separate. If we did something like move all the existing code in the buildpack into a `classic` package and then created a new `berry` package we can keep them logically separated. We could then just have a small amount of code at the top-level that decides with codepath to invoke.

thitch97: I hadn't considered separate packages, that's an interesting suggestion. It would remove a lot of overhead in the event that we drop support for Classic altogether.

text/nodejs/0014-yarn-berry.md
```diff
@@ -0,0 +1,280 @@
...
+Berry has [changed the way pre- & post- scripts are
+invoked](https://yarnpkg.com/getting-started/migration#explicitly-call-the-pre-and-post-scripts).
+
+The build logic of `yarn-start` will need to be modified to accommodate this
+change.
+
+### `yarn node` vs `node` 
+
+All `node` commands run via the shell should be run with `yarn node` instead. The default
+start command in the Yarn Start buildpack may need to be modified.
```
ekcasey, line 230: Can we elaborate on any changes needed to the yarn-start or node-start (included in the berry group above) buildpacks? Seems like there could be some gotchas here with detection criteria and cache vs modules configurations.

text/nodejs/0014-yarn-berry.md
```diff
@@ -0,0 +1,280 @@
...
+
+  [[order.group]]
+    id = "paketo-buildpacks/watchexec"
+    optional = true
+
+  [[order.group]]
+    id = "paketo-buildpacks/node-engine"
+
+  [[order.group]]
+    id = "paketo-buildpacks/yarn-berry-install"
```
ekcasey, line 71: Should this be optional and opt out if users are going with the [zero-installs](https://yarnpkg.com/features/zero-installs) method? This might mess up detection for the group without some changes in (yarn start), but it could be confusing to see yarn install opt-in when you are explicitly striving for zero-installs?

text/nodejs/0014-yarn-berry.md
```diff
@@ -0,0 +1,280 @@
...
+
+  [[order.group]]
+    id = "paketo-buildpacks/node-module-bom"
```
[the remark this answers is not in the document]
thitch97: Good question. I think the inclusion of `node-module-bom` here is a mistake on my part. The buildpack shouldn't need to leverage that buildpack at all for generating the SBoM and would instead include logic for SBoM generation akin to what [exists](https://github.com/paketo-buildpacks/yarn-install/blob/b28f4b300c67bf5bb3c7a55e06cc782f24ed0dee/build.go#L240) in `yarn-install` currently.

text/nodejs/0014-yarn-berry.md
thitch97: This order group may now be obsolete as I was able to prove out a sane way to keep the build logic for both Yarn classic and Berry in a single buildpack (per @ryanmoran's suggestion). As for the UX, I figured we could indicate through the logs whenever the buildpack has chosen to forgo running `yarn install` based on adherence to the zero-installs setup.

divoxx:
Is this RFC still actively being worked on? We have interest in helping push this forward as well, if help is needed.

ekcasey:
@thitch97 IIRC you mentioned during WG you were making some largish changes to the RFC. Do you mind posting a comment describing where it is at and what your intentions are? If it needs more work before it is ready for further discussion/reviews maybe we should put it back into draft temporarily?

thitch97:
@divoxx @ekcasey My intention at the moment is to shift to the model initially enumerated in the "Rationale and Alternatives" section. It turns out that hosting both Classic and Berry codepaths in one buildpack isn't as heavy an engineering/maintenance burden as it seemed. As such, many of the technical decisions here, especially the order groupings, are now obsolete.

fg-j:
From WG 7/19/22:

- In latest update of this RFC, the Build Plan dependency `node_modules` has been renamed `node_packages`. This simplifies plan resolution.
- In latest update, Yarn Berry builds still require `yarn` at build, instead of using `corepack`. This presents some issues that are hashed out in the Alternatives and Unresolved Questions sections.
- Concern: Yarn Berry doesn't seem to play well with offline builds. 
  - Comment: "offline" builds are rare; builds that only have access to a private network are more common
  - Comment: there still is a fully offline workflow for yarn 1
  - Q: Could we enable users to volume-mount a package cache for offline builds? The maven buildpack currently supports this?
    -  This sounds tricky but possible. In yarn context, your global cache on your workstation might contain dependencies that don't have anything to do with the app in question. Also, the location of the package cache isn't configurable for yarn.

- **Requested change:** Add a list of buildpacks that provide or require the `node_modules` plan entry so it's easy to tell which buildpacks would need to change 

Seeking @paketo-buildpacks/nodejs-maintainers  feedback on these changes.

text/nodejs/0014-yarn-berry.md
```diff
@@ -0,0 +1,246 @@
...
+The buildpacks affected by this change to the `node_modules` build plan dependency are:
+
+- `npm-install`
+- `yarn-install`
+- `npm-start`
+- `yarn-start`
+- `node-start`
+- `node-run-script`
+- `node-module-bom`
+- `rails-assets`
```
ryanmoran, line 62: Can you provide some more of the rational behind this choice? I'm not seeing how we don't end up in a state where its either confusing to Berry users or confusing to non-Berry users about what is being provided here.

fg-j: To me, it'd be more intuitive if the yarn classic and yarn berry paths resulted in two slightly different build plans, since they do involve dealing with different sets of artifacts. 

I'd be in favor of the yarn-install buildpack providing `node_modules` if it plans on running `yarn install` and `yarn_cache` if it doesn't. Then the start command buildpacks could require one `Or` the other.

thitch97: @ryanmoran does @fg-j's comment more closely align with what you'd like to see? I'm fine with additional logic for the sake of clarity.

joshuatcasey:
2022-07-26 Working Group notes:

text/nodejs/0014-yarn-berry.md
```diff
@@ -0,0 +1,230 @@
...
+but this feature is experimental and arguably unnecessary in the context of
+buildpack builds where there is only one installation to manage. There are also
+challenges with supporting offline use cases using the Corepack workflow (see
+Rationale & Alternatives). For now, the buildpack should require `yarn` to
+preserve existing functionality & reduce initial engineering cost.
+
+### Detection
+
+The buildpack's detection criteria will remain unchanged, except for the
+replacement of the `node_modules` keyword with `node_pkgs`.
```
ryanmoran, line 62: I don't think this matches the described API below anymore.

thitch97: @ryanmoran good catch. I've updated that now.

ryanmoran (approved):

dmikusa-pivotal:
2022-08-02 WG Notes:

Approved by Node maintainers. Final call for any comments. Scheduled to be merged soon.

PR #203, opened 2022-06-01
[Utilities] Add Live Reload Library

joshuatcasey:
[Utilities] Add Live Reload Library

This has the impact of standardizing and reducing boilerplate for buildpack authors to implement live reload, typically using Watchexec. For further context, see paketo-buildpacks/packit#343.

[readable](https://github.com/joshuatcasey/rfcs/blob/jtc/utilities-add-live-reload-library/text/utilities/0006-add-live-reload-library.md)

### DIFF
--- a/text/utilities/0006-add-live-reload-library.md
+++ b/text/utilities/0006-add-live-reload-library.md
@@ -36,7 +36,7 @@ in this case having a generic interface ("live reload") implies that eventually 
 
 ## Implementation
 
-Create a new repository to host the library: `paketo-buildpacks/libreload`.
+Create a new repository to host the library: `paketo-buildpacks/libreload-packit`.
 The root package `reload` will contain an interface that looks something like this:
 
 ```go


text/utilities/0000-add-live-reload-library.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Implementation
+
+Create a new repository to host the library: `paketo-buildpacks/reload`. The root package `reload` will contain an interface
+that looks something like this:
+
+```go
+package reload
+
+import packit "github.com/paketo-buildpacks/packit/v2"
```
dmikusa-pivotal, line 44: A comment/challenge, could we make this work in a library-agnostic way? Without sacrificing too much on the API.

Alternatively, could we hide the library-specific bits in their own packages so we could support both packit and libcnb without needing to include both projects in a consuming buildpack?

dmikusa-pivotal: I was hoping that conversion might be an option, but the structs in packit and libcnb are slightly different and Go doesn't like that, so it won't convert them.

I haven't played with generics yet as it's new in 1.18 and we haven't updated to that yet on the Java buildpacks side. I would not be opposed to it though, we'll be updating to 1.18 soon enough.

Anyway, my concern is that if we add a library for only one of the two libraries being used by Paketo buildpacks, we're making things harder on buildpack authors. It's already tough enough given we have the library split, if we can make it work with both that would prevent some friction. Hopefully, in the not too distant future, we'll have a packit that's based on libcnb and we'll be sharing the same core structs.

joshuatcasey: I've implemented a beta over at https://github.com/joshuatcasey/libreload-beta but it's still very coupled to `packit v2`. I'll see if I can work out a way to de-couple the `packit` dependency.

dmikusa-pivotal: >How close are we to having a libcnb v2?

I can't say exactly and I think it depends somewhat on how much effort there is to implement what still needs to be done for a v2. 

What I can say is that we've switched the main branch of libcnb to v2, which I think is a good step/sign that we're getting close. I will ask around more and see if I can get a feel for how much work is left.

joshuatcasey: > I don't think generics will help, because you'll still have to import both libraries if you want concrete instantiations of each library's type.

Yeah, this was the point I was trying to make above. I think for now we should just pick `packit v2` and try that - there are about eight packit-based buildpacks that could benefit from this utility, so we get a chance to standardize many different implementations of `watchexec`.

There will also be some lag between `libcnb v2` release and being able to incorporate that into the existing buildpacks that we have. I think we could have a `libreload v1` based on `packit v2` and a `libreload v2` based on `libcnb v2`.

dmikusa-pivotal: Why not put this into packit then? It'll be tied to packit. If that's not an option, I know it was discussed, I think we should perhaps call the proposed library `packit-reload` or something to indicate it's based on packit.

I would like to see us targeting libcnb v2, but I guess it depends on the speed at which we need a solution here. We discussed in the BAT team meeting today and I'd say we're in the milestone stage for a v2 release. We'd like people to start using it and provide some feedback, at the same time, things could still change before we release it.

>I think we could have a libreload v1 based on packit v2 and a libreload v2 based on libcnb v2.

I worry this would be confusing to users, I think that's why a name like `packit-reload` would be more suitable. It's clear what you get.

I worry changing versions could also create a maintenance burden. If there's a v1 release based on packit now, and in three months we cut a v2, how long do we continue to support v1?

dmikusa-pivotal: I'm 👍 for that name. 

I'm not sure that having a separate repo addresses the concerns about maintenance burden, but I'll defer to @ryanmoran  on that since this is now a packit specific library.

ryanmoran: I think creating a new repo called `libreload-packit` is fine. I think my objection to including this code in `packit` is that it is highly coupled to a specific set of buildpacks while the rest of `packit` is generic to any type of buildpack a developer might want to make.

text/utilities/0000-add-live-reload-library.md
```diff
@@ -0,0 +1,71 @@
...
+type ReloadableProcessSpec struct {
+	WatchPaths []string
+	IgnorePaths []string
+	Shell string
+	VerbosityLevel int
+}
+
+type Reloader interface {
+	ShouldEnableLiveReload() (bool, error)
+	GetBuildPlanRequirement() packit.BuildPlanRequirement
```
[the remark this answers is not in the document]
ForestEckhardt: Hmmm. It feels a little unnecessary and like it abstracts away the build plan with I am really not a fan of. The build plan has a ton of control over what happens during the build and I would like an author to be more intentional with their modification of the build plan.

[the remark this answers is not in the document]
joshuatcasey: I think the idea here is that this library enables buildpack authors to add "reload capability". Once they've chosen their preferred implementation (the only implementation now available is `watchexec`) they don't have to know any details about the implementation, such as the name of the provision.

That being said, buildpack authors do at least have to choose the implementation mechanism so it would make sense that the buildpack would know what to request when it creates the buildplan. @ForestEckhardt if you're comfortable with the remainder of the RFC I can drop `GetBuildPlanRequirement()` from the API.

ForestEckhardt (approved):

joshuatcasey:
Thanks for the approval @ForestEckhardt . It looks like I have the ability to merge this, but I'll rename the file first. 

Once it's merged, could you make the new repo `paketo-buildpacks/libreload-packit`?

ForestEckhardt:
@joshuatcasey I would like to see a couple more approvals from people that have given reviews

ryanmoran (approved):

text/utilities/0006-add-live-reload-library.md
```diff
@@ -0,0 +1,75 @@
...
+1. Include the library code in packit. See [PR 343](https://github.com/paketo-buildpacks/packit/pull/343).
+While we did consider the possibility of including the interface ("live reload") in Packit, we knew that the implementation
+("watchexec") was too tightly coupled with `watchexec` to live in Packit.
+2. Include the library code in [the `watchexec` buildpack](https://github.com/paketo-buildpacks/watchexec).
+We felt that we should not include both library code and buildpack code in the buildpack repository, especially since
+in this case having a generic interface ("live reload") implies that eventually another implementation might be possible.
+
+## Implementation
+
+Create a new repository to host the library: `paketo-buildpacks/libreload`.
```
ryanmoran, line 39:
```suggestion
Create a new repository to host the library: `paketo-buildpacks/libreload-packit`.
```

PR #205, opened 2022-06-01
[Dependency Management Phase 1]: Adopting Federated Model RFC

sophiewigmore:
## Summary

[Readable](https://github.com/paketo-buildpacks/rfcs/blob/dependency-management-step-one/text/dependencies/rfcs/0000-dependency-management-phase-one.md)

Outlines the shift to a federated model.
- Related to https://github.com/paketo-community/explorations/issues/8

## Checklist


text/dependencies/rfcs/0000-dependency-management-phase-one.md
```diff
@@ -0,0 +1,287 @@
...
+In the case that the dependencies need to be compiled or processed, the
+metadata generation code should omit the `URI` and the `SHA256` from the
+metadata. This will be used in automation (described in detail in a subsequent
+RFC) to let the dependency management system to trigger compilation of the
+dependency. When the dependency is compiled and uploaded to a bucket, the
+bucket URI will be the URI in the metadata, and the compiled dependency SHA256
+will be the SHA256 in the metadata.
+
+### Smoke Test
+In the dep-server, a smoke test is run against every dependency before the
```
robdimsdale, line 216: I haven't fully visualized the order of operations, so this might be in the wrong place, but I think it's valuable to run some sort of test before bumping the dependency in the source of truth in all cases - not just compiled dependencies.

We should definitely test compiled/processed dependencies before uploading them (and hence making them the source of truth), but I'm also thinking about the situation where a dependency that we take directly from upstream results in broken behavior.

I think it's good enough for now to have a PR model (which runs the acceptance tests) for a potential version bump. That's basically what we have now - i just don't want to lose that feature.

text/dependencies/rfcs/0000-dependency-management-phase-one.md
```diff
@@ -0,0 +1,287 @@
...
+
+### Transfer Version Retrieval Code
+
+Version discovery code comes from
+[dep-server/pkg/dependency](https://github.com/paketo-buildpacks/dep-server/tree/main/pkg/dependency).
+Each dependency has a slightly different way new versions are made
+discoverable. This code should be ported from the dep-server repository to the
+buildpack under a directory named `dependency/retrieval`. It should be
+well-documented and useable on a local environment. The location must be
+standardized for use in automation.
```
robdimsdale, line 186: Are we planning on making a library for common abstracted features?

I'm ok with deferring this decision until we start rolling out the changes as it will be easier to reason about the boundaries of common abstractions once we start to separate the logic from the single existing location.

robdimsdale: I see the comment below about a dep-server library for licenses/pURL info. If we're planning on starting there, that would be a good place to add any abstracted retrieval behavior.

text/dependencies/rfcs/0000-dependency-management-phase-one.md
sophiewigmore: I would recommend not reading these until they're undrafted, theyre undergoing changes.

text/dependencies/rfcs/0000-dependency-management-phase-one.md
sophiewigmore: I think eventually we can, and when we actually get to implementing it may become obvious that we need some common library functions. For the first go-around it's probably find to leave off.

sophiewigmore: I'll add a note about adding other potential abstracted logic to the same dep-server library though.

joshuatcasey (commented):
<removing>

text/dependencies/rfcs/0000-dependency-management-phase-one.md
sophiewigmore: @joshuatcasey made the same comment to me, that we should make an effort to do this in the future so our implementations of other parts of the process are uniform/we can save time in each buildpack. Agreed!

text/dependencies/rfcs/0000-dependency-management-phase-one.md
```diff
@@ -0,0 +1,295 @@
...
+### Bucket Setup
+
+The new dependency management automation (described in a subsequent RFC) will
+rely on the `buildpack.toml` and dependency-specific code as the source of
+truth for the latest metadata and known versions. This will lead to greater
+transparency and discoverability in how we update dependencies.
+
+Since metadata and known versions won't be stored data anymore, buckets will
+only be needed to store dependencies that have been compiled or processed in
+some way. For these cases, a dependency-specific bucket set up in Paketo
```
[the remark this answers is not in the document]
ryanmoran: How would this work? The buildpack installing this dependency would clone its own repo to some SHA and then copy the tarball into the layer?

text/dependencies/rfcs/0000-dependency-management-phase-one.md
sophiewigmore: Sure thing. It might actually be confusing to have all of this stuff live in the dep-server.

text/dependencies/rfcs/0000-dependency-management-phase-one.md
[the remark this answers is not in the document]
joshuatcasey: I was thinking the `buildpack.toml` would have a link to github.com, probably something like this:

https://github.com/joshuatcasey/lfs-demo has a "large file" named `test.lfs`, with a deep link of https://github.com/joshuatcasey/lfs-demo/raw/main/test.lfs

fg-j:
From a buildpack author perspective, I'd like a clearer sense of the proposed API between the contents of the `dependency` directory and the Github Actions that will automate dependency management. 

For example, does `dependency/compilation` need to contain a `compile.sh`? A Dockerfile? What are the expected inputs and outputs of the `dependency/compilation` tool? Likewise, what are the expected inputs and outputs of `dependency/retrieval` and `dependency/metadata`?

text/dependencies/rfcs/0000-dependency-management-phase-one.md
```diff
@@ -0,0 +1,300 @@
...
+Dependencies project within GCP to store data. The directory name will be the
+name of the dependency and it will contain dependency archives (if compiled).
+Push credentials to this bucket will be reserved for the buildpack and
+maintainers, but the contents will be publicly available.
+
+### Remove Known Versions
```
ForestEckhardt, line 174: So my understanding of having known versions is so that not only can the buildpack pick from a set of versions but so the retrieval logic knows what buildpacks it has already seen before. Is the main assumption going forward going to be that if the version number is greater than the one of the lines we are checking for it is new? Just a thought and would like to hear what the general expected strategy is.

ForestEckhardt: Ok I am reading into phase two now and that does seem to answer this question. A quick shout out here would be appreciated.

sophiewigmore: you got it. Done in https://github.com/paketo-buildpacks/rfcs/pull/205/commits/cb117e4f45d2bc4e8d7551667100b5c8b9d7e89b

sophiewigmore:
> From a buildpack author perspective, I'd like a clearer sense of the proposed API between the contents of the dependency directory and the Github Actions that will automate dependency management.

@fg-j this is something that @joshuatcasey and I are looking into with the proof of concept we're working on now, since we need to see how everything can be put into generic workflows. Once we have a clearer idea of the plan I'll update the RFC with the details you're looking for, but there will be a certain level of standardization, without explicitly requiring Golang

text/dependencies/rfcs/0000-dependency-management-phase-one.md
```diff
@@ -0,0 +1,302 @@
...
+dependencies, and stacks, business logic that pertains to specific dependencies
+should be the responsibility of the relevant buildpack.
+
+## Detailed Explanation
+
+The outlined proposal contains a number of changes to the current system which
+should be standardized in order to fit together with generalized Github Actions
+workflows down the line.
+
+### Background
```
ekcasey, line 36: I think it's worth detailing the current workflow for the Java buildpacks as well as the other language families either in background or as an alternative. What tradeoffs would the Java buildpacks have to make to move to this model? Have we considered the pros/cons of adopting more aspects of the Java buildpack workflows in new proposal?

Can we even standardize across all buildpacks without first standardizing the metadata schema?

cc @dmikusa-pivotal

text/dependencies/rfcs/0000-dependency-management-phase-one.md
```diff
@@ -85,7 +85,139 @@ dependency compilation, and code usability.
...
+  * Inputs:
+    * `tarballPath` - the location of the tarball to test
+    * `version` - the dependency version
+
+### Transfer/Combine Version Retrieval and Metadata Generation
+
+The new  version retrieval code will merge together the job of discovering
+versions and getting metadata for each version. It will pick up newer versions
+than what the `buildpack.toml` contains, and outputs JSON-formatted metadata
+for each dependency. This code should be ported from the dep-server repository
```
joshuatcasey, line 169: I might recommend something like "Buildpack authors should place this code under a directory named `dependency/retrieval` and are free to re-use or re-write code from dep-server as necessary", just to make sure to emphasize that buildpack authors can construct this code however they wish (using `bash`, `golang`, `ruby`, etc) and aren't specifically required to port in `dep-server` code. Of course, buildpack authors would be wise to be informed by the decisions encoded in `dep-server`.

sophiewigmore: great idea!

text/dependencies/rfcs/0000-dependency-management-phase-one.md
```diff
@@ -0,0 +1,302 @@
...
+(except the compilation code which will still need to be brought into the
+Paketo organization). Then, language maintainers will be codeowners of certain
+codepaths within the repository. This has the significant drawback that the
+codebase and related automation will still have to significantly change in
+order to support more stacks/architectures. The scope of the dep-server
+repository is already large, as it accounts for every dependency in the
+project, potentially making it hard to contribute and maintain.
+
+## Unresolved Questions and Bikeshedding (Optional)
+- Who is going to pay for the GCP buckets?
```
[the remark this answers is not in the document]
sophiewigmore: Why AWS over GCP? I'm not married to either specifically, but just wondering if there's a reason?

text/dependencies/rfcs/0000-dependency-management-phase-one.md
```diff
@@ -0,0 +1,416 @@
...
+    * `tarballPath` - the location of the tarball to test
+    * `version` - the dependency version
+
+### Transfer/Combine Version Retrieval and Metadata Generation
+
+The new  version retrieval code will merge together the job of discovering
+versions and getting metadata for each version. It will pick up newer versions
+than what the `buildpack.toml` contains, and outputs JSON-formatted metadata
+for each dependency. This code should be ported from the dep-server repository
+to the buildpack under a directory named `dependency/retrieval`, taking in the
```
fg-j, line 171: nit: "must be ported from dep server" contradicts what's said later, "feel free to reuse or rewrite the dep server..."

text/dependencies/rfcs/0000-dependency-management-phase-one.md
```diff
@@ -0,0 +1,416 @@
...
+
+#### 1. Version Retrieval (`make retrieve`)
+  * Description: Discovers dependency versions newer than `buildpack.toml`
+    versions (within constraints) and returns each with a set of generated
+    metadata in the form of a `metadata.json` file.
+  * Inputs:
+    * `buildpackTomlPath`  - the path to the  `buildpack.toml` file
+    * `output` - the path to a file into which an output metadata JSON will be
+      written
+  * Outputs:
```
fg-j, line 128: Once it's solidified, a sample JSON output that represents the output schema would be helpful here.

text/dependencies/rfcs/0000-dependency-management-phase-one.md
```diff
@@ -0,0 +1,416 @@
...
+
+The images listed in the `target.json` file will directly be used as the image
+used during the compilation step during Github Action runs. When compiling a
+dependency locally, users should be able to pull the image themselves and run
+the compilation code on the container.
+
+### Image Preparation
+In the event that a dependency is compiled as outlined in the section above,
+the compilation will take place on a Docker image of the buildpack maintainers'
+choosing. Hopefully, the image that will be used will contain everything needed
```
fg-j, line 326: Is there a way for maintainers to define a Dockerfile for the desired compilation environment – without needing to store that image in a registry somewhere?

text/dependencies/rfcs/0000-dependency-management-phase-one.md
sophiewigmore: added

text/dependencies/rfcs/0000-dependency-management-phase-one.md
sophiewigmore: fixed

text/dependencies/rfcs/0000-dependency-management-phase-one.md
```diff
@@ -0,0 +1,416 @@
...
+    The following files should be made available in the `outputDir`:
+    * Dependency artifact - There should be exactly one tarball, with a name of
+      `${dependency}-<description>.tgz`, where <description> can be any
+      combination of OS, architecture, target platform, version, etc that the
+      buildpack author desires. The dependency name will be highly dependent on
+      if the dependency needs to be compiled separately for different
+      OS/architecture combinations.
+    * Artifact SHASUM -  A file with the same name as the dependency artifact,
+      with a `sha256` suffix, containing the SHA256 of the compiled dependency.
+
```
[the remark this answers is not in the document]
sophiewigmore: Yes. However I am rethinking this.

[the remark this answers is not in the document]
sophiewigmore: If folks re-use the existent dependency smoke tests, some of them test the SHA256 of the dependency is right, so this could be helpful for that.

text/dependencies/rfcs/0000-dependency-management-phase-one.md
sophiewigmore: We looked into this by way of having compilation be written as an action, with different Dockerfiles for different compilation environment possibilities. GHA environments don't allow for dynamic actions like that so we couldn't do it like that really. This is where the preparation script came from, it'll essentially take an existent image and stuff on top of it if needed.

We figured in most cases, the environment can be an existent stack image, in which case adding things won't be necessary anyway.

text/dependencies/rfcs/0000-dependency-management-phase-one.md
joshuatcasey: There are a small handful of JSON files used to pass information around through the workflow in a structured way.

I suspect we could add golang struct representations of these in `libdependencies` that could be the ongoing "canonical" representation of the JSON. Buildpack authors can choose any language to produce or ingest the JSON but should realize that the workflow will probably rely on the representation from `libdependencies`.

text/dependencies/rfcs/0000-dependency-management-phase-one.md
joshuatcasey: Maintainers could also maintain a docker image completely outside the scope of dependencies and then reference that docker image for the dependencies workflow. I know it's a few extra steps but we couldn't find a way for the workflow to allow for buildpack authors to choose either an image or a Dockerfile, so we chose to use images.

joshuatcasey: The workflow will allow buildpack authors to specify some additional "preparation steps" in case an off-the-shelf image doesn't quite have everything. Those steps will be in the form of a shell script that will only run for the target. Something like so:

```shell
If target=="macos";
  brew update && brew install coreutils
else if target=="ubuntu";
  apt-get update && apt-get install -y wget
else if target=="alpine";
  apk update && apk add wget
fi
```

joshuatcasey: And.... just realized @sophiewigmore 's already described this over under '# Image Preparation'. :)

text/dependencies/rfcs/0000-dependency-management-phase-one.md
```diff
@@ -0,0 +1,302 @@
...
+0000](https://github.com/paketo-buildpacks/rfcs/blob/dependency-management-top-level/text/dependencies/rfcs/0000-dependency-management-overview.md),
+the current dependency management system employed
+in the project leads to a number of painpoints, predominantly around
+maintainability for Dependencies maintainers, and lack of extensibility by
+compiling dependencies for Ubuntu 18.04. In order to ensure the Paketo
+Buildpacks project can continue to grow and introduce new buildpacks,
+dependencies, and stacks, business logic that pertains to specific dependencies
+should be the responsibility of the relevant buildpack.
+
+## Detailed Explanation
```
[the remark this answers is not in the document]
sophiewigmore: Actually, I've updated Phase 2 with some details. Let me know if you'd like more info still.

text/dependencies/rfcs/0000-dependency-management-phase-one.md
sophiewigmore: Update from team share out - I am going to look into an alternative to this proposal involving Dockerfiles. Please hold!

text/dependencies/rfcs/0000-dependency-management-phase-one.md
```diff
@@ -0,0 +1,437 @@
...
+account of in the new compilation code.
+
+The code will eventually be used in a Github Actions workflow, so the location
+needs to be standardized across buildpacks. The workflows and Github Actions to
+use this code will be enumerated in a separate RFC.
+
+### Bucket Setup
+
+For dependency metadata and the dependencies that must be compiled, there
+should be a dependency-specific bucket set up in Paketo Dependencies project
```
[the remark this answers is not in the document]
ryanmoran: Can we at least make this less specific? We won't be making a new Dependencies project, and it might ultimately no end up on GCP at all. I feel like we don't need to be this specific here.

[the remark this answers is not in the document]
sophiewigmore: Sure.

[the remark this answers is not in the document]
sophiewigmore: https://github.com/paketo-buildpacks/rfcs/pull/205/commits/9136ab0ec53c631579b8fc66953e978dcda13820

text/dependencies/rfcs/0000-dependency-management-phase-one.md
sophiewigmore: Depending on Java-side buy in, we may loosen this schema a bit. TBD.

text/dependencies/rfcs/0000-dependency-management-phase-one.md
sophiewigmore: Alright, per @fg-j  recommendation I tried out an approach of rolling compilation into an action so that the image compiled on is configurable via Dockerfile, so there can be different Dockerfiles for each target. This works **beautifully** locally, allowing users to run compilation locally with a few Docker commands.

Unfortunately, any way you slice it, there's absolutely no way I have discovered to dynamically swap out the Dockerfile used in an action, so combining this strategy with a matrix (i.e. we run compilation for each target) isn't feasible. The only way around it would be if the workflow was hardcoded to run compilation for each target.

sophiewigmore: Check out https://github.com/joshuatcasey/bundler/commit/5009e1635ce5bb27cd77ed64c35ed5375fc28c33 to see what this would look like in practice

text/dependencies/rfcs/0000-dependency-management-phase-one.md
sophiewigmore: Added one for the sake of an example though anyways.

text/dependencies/rfcs/0000-dependency-management-phase-one.md
sophiewigmore: Alright, the RFC has been updated to reflect this idea as if https://github.com/paketo-buildpacks/rfcs/pull/205/commits/ca21bfc023a52cd51f6e768e65605a64e3d62dd1

text/dependencies/rfcs/0000-dependency-management-phase-one.md
```diff
@@ -0,0 +1,446 @@
...
+
+```
+[
+  {
+    "target": <generic name of the group, to be used during compilation>,
+    "stacks": [ <list of compatible stacks>]
+  },
+  ...
+]
+```
```
fg-j, line 357: Now that there will be 1 Dockerfile per target inside `dependency/actions/compile` (if I'm understanding that correctly), do you think there's a clever way we could avoid needing a `targets.json` alltogether? Part of me thinks the fewer descriptive files we have to check in, the better. What do others think about this?

sophiewigmore: Yea, I definitely see your point and agree it would be nice to get rid of this, it's unfortunate to include a file just for the sake of making the automation run more smoothly... The `targets.json`  provides a nice way to iterate over the variants of a dependency, and have that laid out clearly. The workflow outlined in Phase 2 heavily depends on iterating over a set of variants that come from each target. 

I think we could get away with removing the `targets.json` file, and relying on the `retrieval` step to be hardcoded to output a batch of metadata for each target variant. Automation could then iteration over each metadata set to perform compilation /  testing. We'd probably want to add a `target` output to the metadata set.

In my mind, the `targets.json` strategy may be clearer because maintainers have to explicitly lay out the separate variants and their compatible stacks. They'd still have to do this in the other proposal I just outlined, but it would be more hidden away as an implementation detail.

sophiewigmore: Open to removing the `targets.json` field, interested to hear what others think about this too.

sophiewigmore: I've been mulling this over and I've actually come around to removing the `targets.json` file in favour of spitting out different metadata entries for each target. The latest commit I added has this outlined. I will update Phase 2 now.

text/dependencies/rfcs/0000-dependency-management-phase-one.md
fg-j: Is this schema more solidified now? Are there optional vs. required fields?

sophiewigmore: @dmikusa-pivotal  [cited concerns](https://github.com/paketo-buildpacks/rfcs/pull/204#discussion_r915930540) around differences in metadata approaches, but I think that this schema should be solidified in order for all of the workflows to leverage the same "assembly" step to put stuff into the `buildpack.toml`. I'll add clarity to this in the RFC.

fg-j (approved):

text/dependencies/rfcs/0000-dependency-management-phase-one.md
robdimsdale: I am slightly confused as to whether there will be one `metadata.json` file for each version (or each constraint?) or whether there will be a single `metadata.json` file with an array of objects described below. Either could work, but I'd appreciate some more clarity.

robdimsdale (approved):

text/dependencies/rfcs/0000-dependency-management-phase-one.md
sophiewigmore: It's a single `metadata.json` file! I added a little bit of language clarity to some of the sections

ryanmoran (approved):

text/dependencies/rfcs/0000-dependency-management-phase-one.md
```diff
@@ -0,0 +1,440 @@
...
+Code to generate/gather all of the metadata for a dependency should be moved and combined
+into the retrieval code.  The code should
+do essentially the same thing that the existent code does, and support the same
+metadata fields.
+The supported metadata fields are: `source URI`, `source SHA256`, `version`,
+`URI`, `SHA256`, `ReleaseDate`, `DeprecationDate`, `stacks`, `purl`, `licenses`
+and `CPE`.
+
+#### Separating OS Variants and Supporting Multiple Stacks
+An additional metaadata field named `target` will be added, which will be the
```
dmikusa-pivotal, line 199: A couple of concerns about this:

1. At the CNB level, there has been talk for a while about removing the concept of a "stack". See [RFC 96](https://github.com/buildpacks/rfcs/blob/main/text/0096-remove-stacks-mixins.md). There are a number of references to stack in this proposal, in particular this section. I think we should try to get ahead of that change so we're not coupling this process to a term that's being phased out. I think we should try to use [this terminology instead](https://github.com/buildpacks/rfcs/blob/main/text/0096-remove-stacks-mixins.md#base-image-metadata).

2. Somewhat related, but I think we need to include support for multiple architectures in this proposal. ARM64 support is hopefully not that far out and dependency support for multiple architectures is a requirement for that. I'm not sure if `target` was intended to cover architecture as well. I could see this being `target` being a format like `ARCH-VENDOR-SYS-ABI` ex: `aarch64-apple-darwin` or `x86_64-unknown-linux-gnu` or `x86_64-unknown-linux-musl`. I could also see having individual properties for each of those value, like `arch=x86_64`, `vendor=ubuntu-bionic`, `sys=linux`, `abi=musl` (not sure we need all of these fields, names could change as well, just an example).

sophiewigmore: @dmikusa-pivotal I have been aware that stacks will go away in the future- beyond the specific language that calls out "stacks" in the RFC, I actually think that this proposal should still work more or less for the new world with the new fields, without any major architectural changes. I think it should be fine to move forward with the proposal as is and then when we actually get to a place in the project where we remove stacks - we can make amendments as needed.

dmikusa-pivotal: Discussed in WG meeting.

There's no reason to believe that deferring this work and removing stack terminology to a later date will be an overly difficult task (we can add/deprecate fields in metadata and adjust implementations). There is also concern that trying to do it now might create challenges as the CNB project hasn't implemented this yet, so there's still a chance for things to change/deviate from the RFC. Thus the consensus on the call was to proceed now with the stack terminology.

sophiewigmore:
@paketo-buildpacks/steering-committee @paketo-buildpacks/dependencies anything needed here? Looking for comments or approvals!

joshuatcasey (approved):

ForestEckhardt (approved):

dmikusa-pivotal (approved):

brayanhenao (approved):

PR #206, opened 2022-06-01
[Dependency Management Phase 2]: Github Actions/Workflows RFC

sophiewigmore:
## Summary

[Readable](https://github.com/paketo-buildpacks/rfcs/blob/dependency-management-step-two/text/dependencies/rfcs/0000-dependency-management-phase-two.md)

Outlines the new process for discovering and updating dependencies.
- Related to https://github.com/paketo-community/explorations/issues/8

## Checklist


text/dependencies/rfcs/0000-dependency-management-phase-two.md
```diff
@@ -0,0 +1,201 @@
...
+- Picking up new versions with buildpack code will be a workflow step
+  It should run twice per day.
+- Figuring out which versions will be added to the `buildpack.toml` will be
+  rolled into an action that outputs versions.
+- A Github Actions [`matrix
+  strategy`](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#using-a-matrix-strategy)
+  will be used to iterate over each version to be added, to generate metadata
+  using the code from within the buildpacks.
+- In the same matrix loop, dependency compilation will be kicked off with the
+  code from the buildpacks on the basis of whether the metadata contains the
```
robdimsdale, line 124: Is the logic to detect if we need to compile or not going to be in github actions configuration or in an executable in the buildpack?

Phrased another way - if you were running the executable locally would you expect the user to know whether to compile (i.e. the logic lives in the action config) or the executable will decide for them (i.e. the logic is in the executable in the buildpack)

text/dependencies/rfcs/0000-dependency-management-phase-two.md
```diff
@@ -0,0 +1,201 @@
...
+  `SHA256` and `URI` of the dependency.
+- The dependency will be uploaded via an action
```
robdimsdale, line 126: This is only in the compiled/processed case, yes? Can we clarify this point in the RFC?

text/dependencies/rfcs/0000-dependency-management-phase-two.md
```diff
@@ -0,0 +1,201 @@
...
+- The metadata is updated with the SHA256 and URI in a step
+- Updating the `buildpack.toml` with the new versions and metadata will also be
+  delegated to a new action
+- A pull request will be opened with the changes to the `buildpack.toml`
```
robdimsdale, line 130: And this PR will run the usual battery of checks - integration tests etc?

text/dependencies/rfcs/0000-dependency-management-phase-two.md
```diff
@@ -0,0 +1,201 @@
...
+
+All of these steps will run as a workflow, but can be run easily manually.
+Additionally, we could eventually create a script inside of the
+`<buildpack>/scripts` directory to perform the main steps of the
+`buildpack.toml` update process locally.
+
+- TODO: how to deal w multiple variants of the same dependency
```
robdimsdale, line 137: does "multiple variants of the same dependency" refer to different architectures (i.e. stacks)? Or something else?

text/dependencies/rfcs/0000-dependency-management-phase-two.md
```diff
@@ -0,0 +1,201 @@
...
+- TODO: how to isolate failures? Should we split it into jobs incase failures occur
+- TODO: manually running? can you do a manual dispatch
```
robdimsdale, line 139: Yes, this should be simple and we should document it. I think that this will need to be run manually if we're in a rush - e.g. trying to ship a high-priority fix.

text/dependencies/rfcs/0000-dependency-management-phase-two.md
sophiewigmore: It'll be in the buildpack code, outlined in Phase 1 RFC. A user does not need to know this information, I would expect the buildpack code to contain this logic

text/dependencies/rfcs/0000-dependency-management-phase-two.md
sophiewigmore: Good idea, I'll add.

text/dependencies/rfcs/0000-dependency-management-phase-two.md
```diff
@@ -0,0 +1,195 @@
...
+
+This phase will address concerns around the opaque complexity of the dep-server
+process. Additionally, it addresses the painpoint of maintainability.
+Dependencies maintainers will own the new workflows and actions, ensuring every
+relevant buildpack follows the same overall system architecture, while giving
+buildpacks maintainers freedom to make dependency-specific choices separately
+and debug issues with full authority.
+
+The new actions will also be simple to run locally in the event of
+Github Actions degradations, making the dependency-management system more
```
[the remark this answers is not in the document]
sophiewigmore: Great question. I think this is something we need to do more exploration on.

text/dependencies/rfcs/0000-dependency-management-phase-two.md
```diff
@@ -0,0 +1,195 @@
...
+
+Dependencies are updated in the buildpacks in a workflow that runs
+in the buildpack's Github Actions:
+- **Update Dependencies**: Runs every hour and runs [`jam
+  update-dependencies`](https://github.com/paketo-buildpacks/github-config/blob/f85a65d14fd90f6d63f1af2d408ee38f17ce5c0b/actions/dependency/update/action.yml#L45-L46)
+  against the `buildpack.toml`. If there are new versions available in the
+  dep-server, this workflow will discover them and open a PR to update the
+  `buildpack.toml`.
+
+### New Plan Overview
```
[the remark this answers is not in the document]
sophiewigmore: I like this idea. @joshuatcasey and I are investigating what this looks like now and will update the RFC if we find it viable

text/dependencies/rfcs/0000-dependency-management-phase-two.md
```diff
@@ -0,0 +1,195 @@
...
+`actions/dependency` and `implementation/.github/workflows/<all
+dependency-related workflows>`.
+
+
+Each of the step s outlined in the overview will be converted into workflow
+steps.
+
+Workflow 1:
+1. Picking up new versions with buildpack code will be a workflow step, which
+   runs on a timer twice per day.
```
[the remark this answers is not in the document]
sophiewigmore: Done in latest update.

text/dependencies/rfcs/0000-dependency-management-phase-two.md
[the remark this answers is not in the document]
sophiewigmore: We may need to also pass in the OS/arch as arguments to the `make` command, because we've encountered a bug in which compilation code for bundler works differently for bionic vs jammy.

text/dependencies/rfcs/0000-dependency-management-phase-two.md
```diff
@@ -0,0 +1,195 @@
...
+Dependencies maintainers will own the new workflows and actions, ensuring every
+relevant buildpack follows the same overall system architecture, while giving
+buildpacks maintainers freedom to make dependency-specific choices separately
+and debug issues with full authority.
+
+The new actions will also be simple to run locally in the event of
+Github Actions degradations, making the dependency-management system more
+flexible in use and transparent in process.
+
+## Detailed Explanation
```
joshuatcasey, line 33: Buildpack authors will need to specify the desired operating systems somehow. This could be specified in a file such as `<buildpack>/dependency/operating_systems.json` that looks like this:

```json
["ubuntu-18.04", "ubuntu-22.04", "macos-latest"]
```

Note that it's currently written on a single line (see https://github.com/joshuatcasey/bundler/runs/6802820650?check_suite_focus=true) but that should be fixable.

joshuatcasey: Do we need to give some thought to those OS/arch combinations that are not available natively via github actions virtual environments (https://github.com/actions/virtual-environments/)?

UBI is built on RHEL, are we sure that binaries built on Ubuntu will work for them? https://www.redhat.com/en/blog/introducing-red-hat-universal-base-image

text/dependencies/rfcs/0000-dependency-management-phase-two.md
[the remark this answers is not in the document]
joshuatcasey: We may also consider a `prepare` step that could do such things as install additional tools. It could be invoked as something like `make prepare os=<whatever>` so that it can handle the different operating systems.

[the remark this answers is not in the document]
joshuatcasey: We may want to add a file such as `dependency/NAME` that contains the ID of the dependency.

Also, are we aware of any buildpacks that compile more than one ID at a time?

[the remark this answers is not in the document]
joshuatcasey: We may also need a `prepare-pull-request` step that combines 6, 7, and 8

text/dependencies/rfcs/0000-dependency-management-phase-two.md
```diff
@@ -0,0 +1,195 @@
...
+The github-config repository `CODEOWNERS` file will be updated to set the
+`@paketo-buildpacks/dependencies-maintainers` to the owners of the
+`actions/dependency` and `implementation/.github/workflows/<all
+dependency-related workflows>`.
+
+
+Each of the step s outlined in the overview will be converted into workflow
+steps.
+
+Workflow 1:
```
fg-j, line 115: I'd find it easier to understand the explanation of workflows if they were pseudo-coded as actual workflow yaml, maybe with explanatory comments sprinkled throughout.

text/dependencies/rfcs/0000-dependency-management-phase-two.md
```diff
@@ -0,0 +1,195 @@
...
+
+### Updating Dependencies Manually
+All of these steps will run as a workflow with the option for a manual
+dispatch, but can be run easily manually on a local system.  Additionally, we
+could eventually create a script inside of the `<buildpack>/scripts` directory
+to perform the main steps of the `buildpack.toml` update process locally. This
+will replace the need for the `jam update-dependencies` command that the
+automation uses currently. Since dependency-related logic will live alongside
+the buildpack, it makes more sense for manual dependency-update scripts to live
+there as well in order to leverage code.
```
ForestEckhardt, line 167: I am a little curious why we would want to write a new script instead of just modifying the functionality of `jam update-dependencies` to do whatever that script would do, or even potentially adding some new command if you want to keep the previous generations functionality.

text/dependencies/rfcs/0000-dependency-management-phase-two.md
sophiewigmore: Yea, this is something I had looked into, I think we might be able to do something with the base image used in the Github Action Dockerfile

text/dependencies/rfcs/0000-dependency-management-phase-two.md
sophiewigmore: I think that explaining the workflows this way is easier for someone less familiar 
with Github Actions details to get the point across of what the automation is responsible for, which is all I wanted for this section. This avoids shoehorning the details of the workflows, which may change when actually implemented. I think I'd prefer to leave the explanation at this level for the RFC, especially since the workflows will be implemented once and rolled out to buildpacks, rather than buildpacks implementing their own.

text/dependencies/rfcs/0000-dependency-management-phase-two.md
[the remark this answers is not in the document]
fg-j: Running compilation in containers solves part of this problem. For containers that are compatible with the operating system you're running locally, you can simply spin up the same container used in CI, even if container is based on a different OS. For example. to compile  for alpine amd64  in an alpine-based container , simply pull the compilation image onto your mac with amd64 architecture and run it. 

Things get a bit more complicated compiling for an architecture that's different from your local machine. Is it possible at all to compile for ARM64 architectures on an amd64 box? What about compiling for Windows on a non-Windows box? Devs may need to use a remote runner for these types of dependency builds, even when they're building outside the context of the CI pipeline.

text/dependencies/rfcs/0000-dependency-management-phase-two.md
```diff
@@ -0,0 +1,195 @@
...
+1. Pick up all versions of the dependency, using code from the Phase 1 RFC in
+   `<buildpack>/dependency/retrieval`
+2. Compare the list of all versions with the version constraints and supported
+  versions in the `buildpack.toml` to determine what versions can be updated.
+3. Generate metadata for each of the versions to be added using code from Phase
+   1 in `<buildpack>/dependency/metadata`.
+4. If the `URI` and `SHA256` are missing from the metadata, the code will be
+  compiled with the code from `<buildpack>/dependency/compilation`.
+5. Test the dependency (whether compiled or not) using the test from
+   `<buildpack>/dependency/test`
```
ekcasey, line 87: If we are using dependencies directly from upstream do we need to test them? In most cases, where we aren't compiling it seems like the workflow should be as simple as generating metadata and PRing `buildpack.toml`. If we need to test anything we could run those tests as a gate on the PR.

text/dependencies/rfcs/0000-dependency-management-phase-two.md
```diff
@@ -0,0 +1,195 @@
...
+   `SHA256` and `URI` of the dependency.
+
+6. A smoke test will be run against the dependency in both the
+   compiled/non-compiled cases via a step that simply runs the test from
+   `<buildpack>/dependency/test`.
+
+7. (If compiled) The dependency will be uploaded via an action
+
+8. (If compiled) The metadata is updated with the SHA256 and GCP bucket URI in
+   a workflow step
```
ekcasey, line 149: Have we considered making this a separate workflow that can be used in the rare cases where it is needed? Then the update/metadata workflow could be the same in all cases, the only difference being it looks for our compiled dependency instead of an upstream artifact.

text/dependencies/rfcs/0000-dependency-management-phase-two.md
ForestEckhardt: I guess I just think it would be better to keep as many buildpack development tool in `jam` as possible to hopefully help spread this way of doing dependencies but I am not tied to the idea.

text/dependencies/rfcs/0000-dependency-management-phase-two.md
sophiewigmore: addressed in the latest version of the rfc

text/dependencies/rfcs/0000-dependency-management-phase-two.md
sophiewigmore: A lot of the functionality of `jam update-dependencies` is spread out into smaller, composable pieces so that they can be plugged in with other functionality in order to support a bunch of different use cases.

I'd say that the "Assemble" step (see step 9 in [the overview](https://github.com/paketo-buildpacks/rfcs/blob/dependency-management-step-two/text/dependencies/rfcs/0000-dependency-management-phase-two.md#new-plan-overview)) might be the best candidate for pushing into `jam update-dependencies` instead of a Github action. It would be a command that just knows how to update the `buildpack.toml` given metadata. I'm not particularly married to it being an action or a `jam` command really

text/dependencies/rfcs/0000-dependency-management-phase-two.md
sophiewigmore: I'm not sure I understand why this should be a separate workflow step. I think it can pretty easily just be coupled with the compilation job in the workflow, such that if compilation/artifact upload step runs, then the SHA256 and URI are just updated in the metadata. It seems pretty coupled, I can't wrap my head around why we'd need a separate mechanism to do this.

The other steps such as original metadata generation, and the `buildpack.toml` update will be the same in every case

text/dependencies/rfcs/0000-dependency-management-phase-two.md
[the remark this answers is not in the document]
sophiewigmore: I think we've addressed this somewhat in the RFC. Running ARM64 compilation will require a bit more work but I think is achievable with the same structure as outlined more or less.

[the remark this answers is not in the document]
ryanmoran: Multi-arch builds are achievable on a single machine. This is what `jam create-stack` is leveraging. Check out the docs [here](https://docs.docker.com/buildx/working-with-buildx/#build-multi-platform-images) and [here](https://docs.docker.com/desktop/multi-arch/). Basically, Docker can leverage this thing called `binfmt_misc` to handle different architectures than the host. You sacrifice a bit of performance, but it should generally work.

text/dependencies/rfcs/0000-dependency-management-phase-two.md
sophiewigmore: Let's leave this up to maintainers to decide.

text/dependencies/rfcs/0000-dependency-management-phase-two.md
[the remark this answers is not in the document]
fg-j: Maybe we can take advantage of multi-arch support via that `--platform` flag to compile dependencies for ARM64 in containers without needing to self-host ARM64 Github runners.

text/dependencies/rfcs/0000-dependency-management-phase-two.md
fg-j: I’d say, better to be an action if that step should run in a container, better to put in `jam` if it’ll “just work” on any CLI users local workstation, given the appropriate inputs.

sophiewigmore: alright seeing as two tooling maintainers are both on board with the metadata update step as a part of `jam`, I have updated the RFC to reflect that. It does not need to run inside of a container specifically, so it probably doesn't need to be an action.

sophiewigmore:
@dmikusa-pivotal [the latest commit](https://github.com/paketo-buildpacks/rfcs/pull/206/commits/bfadcf3c06349fedbe863fabf6294512fa01fac5) adds some generalization around where dependency data is stored so that we aren't locked into the `buildpack.toml`. Let me know if these changes seem reasonable / if this RFC is compatible with your other proposal.

ryanmoran (approved):

sophiewigmore:
@paketo-buildpacks/steering-committee @paketo-buildpacks/dependencies @robdimsdale @ForestEckhardt @fg-j anything needed here? Looking for comments or approvals!

joshuatcasey (approved):

ForestEckhardt (approved):

fg-j (approved):

dmikusa-pivotal (approved):

brayanhenao (approved):

PR #212, opened 2022-06-23
RFC: Remove `dep` and `dep-ensure` Buildpacks

ForestEckhardt:
[Readable](https://github.com/paketo-buildpacks/rfcs/blob/deprecate-dep/text/go/0001-deprecate-dep.md)

### DIFF
--- a/text/go/0001-remove-dep.md
+++ b/text/go/0001-remove-dep.md
@@ -1,8 +1,8 @@
-# Deprecate the `dep` and `dep-ensure` Buildpacks
+# Remove the `dep` and `dep-ensure` Buildpacks
 
 ## Proposal
 
-The `dep` and `dep-ensure` Buildpacks should be deprecated.
+The `dep` and `dep-ensure` Buildpacks should be removed.
 
 ## Motivation
 
@@ -12,7 +12,7 @@ deprecated and archived by the team working on it (See the [Github
 Repository](https://github.com/golang/dep) and the [`dep`
 documentation](https://golang.github.io/dep/docs/introduction.html)). Because
 this tool has been deprecated and `go mod` has been widely adopted by the Go
-community, we should deprecate our buildpacks.
+community, we should remove our buildpacks.
 
 ## Implementation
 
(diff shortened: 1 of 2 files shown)


joshuatcasey:
Should there be a warning period if we think anyone is using these buildpacks? Could look like any of the following:Put something on Slack and wait 30 days?

ForestEckhardt:
@joshuatcasey I am open to all of those things as potential options and any more if people have opinions. I would say that `dep` has been deprecated for around 2 years already so we are not doing this like 1 month after they called it quits.

text/go/0001-remove-dep.md
```diff
@@ -0,0 +1,26 @@
...
+
+`dep` was a package management tool created for Go before the introduction of
+`go mod`. Since the introduction of `go mod` the `dep` tool has since been
+deprecated and archived by the team working on it (See the [Github
+Repository](https://github.com/golang/dep) and the [`dep`
+documentation](https://golang.github.io/dep/docs/introduction.html)). Because
+this tool has been deprecated and `go mod` has been widely adopted by the Go
+community, we should remove our buildpacks.
+
+## Implementation
```
ekcasey, line 17: I would like to see the buildpack help users migrate. For example, what if we shipped a version of the buildpack that still worked but printed loud deprecation warnings and a link to migration steps if it detected/ran? We could include that version in the language family buildpack for a period of time before removing it.

If we just remove it without warning that it feels a bit like we have failed to provide a deprecation window, which isn't the best experience for users, even if the tool itself is deprecated.

joshuatcasey:
As a nice side benefit, we will no longer need to maintain the `dep` dependency (https://github.com/paketo-buildpacks/dep/blob/main/buildpack.toml).

Should we add a step in here to remove automation that generates this dependency?

text/go/0001-remove-dep.md
fg-j: I think a deprecation period a good idea. A month or so seems reasonable, given the underlying tech is already out of support.

ForestEckhardt: Ok I have specified that a 30 day deprecation window should be implemented.

ryanmoran (approved):

ekcasey (approved):

PR #214, opened 2022-06-29
Python Dependencies

joshuatcasey:
Remove the Paketo-hosted dependencies for `poetry` and `pipenv`.

[readable](https://github.com/joshuatcasey/rfcs/blob/jtc/python-dependencies/text/python/0000-dependencies-update.md)

### DIFF
--- a/text/python/0000-dependencies-update.md
+++ b/text/python/0000-dependencies-update.md
@@ -1,0 +1,93 @@
+# Decide which Python dependencies will be Paketo-hosted
+
+## Proposal
+
+Currently four Python buildpacks have Paketo-hosted dependencies.
+Only two Paketo-hosted dependencies should be kept, and two Paketo-hosted dependencies 
+should be removed in favor of consuming the upstream-provided dependency.
+
+| Buildpack                                                                        | Action |
+|----------------------------------------------------------------------------------|--------|
+| [cpython](https://github.com/paketo-buildpacks/cpython/blob/main/buildpack.toml) | Keep   |
+| [poetry](https://github.com/paketo-buildpacks/poetry/blob/main/buildpack.toml)   | Remove |
+| [pip](https://github.com/paketo-buildpacks/pip/blob/main/buildpack.toml)         | Keep   |
+| [pipenv](https://github.com/paketo-buildpacks/pipenv/blob/main/buildpack.toml)   | Remove |
+
+### Offline not supported
+
+Note that the Python language family does not support a fully-offline case, since several of its components require internet connectivity to work.
+These include at least the following: 
+
+* pipenv: [#limitations](https://github.com/paketo-buildpacks/pipenv#limitations)
+* poetry: [#known-issues-and-limitations](https://github.com/paketo-buildpacks/poetry#known-issues-and-limitations)
+
+## Rationale
+
+### CPython
+
+Keep this as a Paketo-hosted dependency.
+
+Currently this uses `dep-server`, `binary-builder`, and `buildpacks-ci` to build, but language family maintainers
+will transition this to the new Github Action workflow once that has been approved. Due to the complexity of the current
+build process, this RFC does not propose removing this as a Paketo-hosted dependency.
+
+### Poetry
+
+Remove the Paketo-hosted dependency.
+
+This dependency seems to be the result of a `pip download` command as seen in [buildpacks-ci](https://github.com/cloudfoundry/buildpacks-ci/blob/6e8873b5a1535c2faf56f810fde7063864db7585/tasks/build-binary-new/builder.rb#L98-L122).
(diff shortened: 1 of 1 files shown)


text/python/0000-dependencies-update.md
```diff
@@ -0,0 +1,97 @@
...
+* poetry: [#known-issues-and-limitations](https://github.com/paketo-buildpacks/poetry#known-issues-and-limitations)
+
+## Rationale
+
+### CPython
+
+Keep this as a Paketo-hosted dependency.
+
+Currently this uses `dep-server`, `binary-builder`, and `buildpacks-ci` to build, but language family maintainers
+will transition this to the new Github Action workflow once that has been approved.
```
sophiewigmore, line 30: The rationale behind why we need to continue hosting this ourselves is not clear. What's the problem with using the version directly from upstream?

sophiewigmore: What's going on in the compilation code that's needed?

text/python/0000-dependencies-update.md
```diff
@@ -0,0 +1,97 @@
...
+As of 2022-06-28, this buildpack is actually not using the dependency from its `buildpack.toml`.
+Instead, the buildpack is running `pip download poetry==$VERSION`, see its [source code](https://github.com/paketo-buildpacks/poetry/blob/ca9e73b34d7018ce00f7d75c76382a08df554a41/poetry_install_process.go#L35).
+
+This RFC does not require Python maintainers to incorporate all logic from `buildpacks-ci` into the `poetry` buildpack.
+
+Also, see [Offline not supported](#offline-not-supported).
+
+#### Poetry layer caching and reproducibility
+
+Remove support for `poetry` layer caching and image reproducibility.
```
sophiewigmore, line 48: This seems like a pretty hallmark feature of buildpacks. Is there no way to consume Poetry another way? Or to prevent the `pip download` to be skipped on rebuilds?

text/python/0000-dependencies-update.md
robdimsdale: There are currently [quite a lot of steps](https://github.com/cloudfoundry/buildpacks-ci/blob/ff0027d31d6749d67670b83a7b2da500d992c435/tasks/build-binary-new/builder.rb#L248) to build cpython. We could explore moving all these steps to the buildpack but I don't think it's a trivial change.

text/python/0000-dependencies-update.md
```diff
@@ -0,0 +1,97 @@
...
+
+Since we rely on `pip` to install `poetry`, the buildpack may not know the version of `poetry` being installed until after the installation completes,
+meaning it is not generally possible to determine if a pre-existing layer can be re-used.
+
+Also, since `poetry` depends on many other packages which can update at any time,
```
robdimsdale, line 53: I think it is ok to not update transitive dependencies when we have a cached layer. We don't have any reason to believe that this is an issue that consumers face, and the resolution is fairly simple - clear the image cache (e.g. `docker system prune`) and re-run the build.

fg-j:
From Working Group 7/19/22:

- Next steps: RFC will be updated to incorporate a way to implement layer caching for dependencies that are sourced directly from upstream.

joshuatcasey:
TLDR on the changes:

- emphasize that `cpython` will be kept (for now) as a Paketo-hosted dependency due to its complexity
- have deterministic installs of `poetry` and `pipenv` by using `metadata.default-versions`
- ensure that the current layer reuse strategy is preserved with these changes

text/python/0000-dependencies-update.md
joshuatcasey: The compilation process involves installing system packages, so this RFC proposes keeping `CPython` as a Paketo-hosted dependency built outside of the buildpack.

text/python/0000-dependencies-update.md
```diff
@@ -0,0 +1,97 @@
+# Decide which Python dependencies will be Paketo-hosted
+
+## Proposal
+
+Currently four Python buildpacks have Paketo-hosted dependencies.
+Only two Paketo-hosted dependencies should be kept, and two should be removed.
```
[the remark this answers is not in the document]
joshuatcasey: I've updated the language

text/python/0000-dependencies-update.md
joshuatcasey: I've updated the language to reflect that we want to know (via the `buildpack.yaml[metadata.default-versions]`) what version of the dependency will be installed, and that the dependency layer will be preserved for reuse.

text/python/0000-dependencies-update.md
joshuatcasey: Sounds good to me.

robdimsdale (approved):

thitch97 (approved):

PR #231, opened 2022-07-29
Add rfc to include libs to tiny that are needed for node

c0d1ngm0nk3y:
[Readable](https://github.com/sap-contributions/rfcs/blob/node-tiny-stack/text/stacks/0000-node-on-tiny.md)

## Summary

> At the moment, Node.js applications are not compatible with the tiny builder. Therefor the `nodejs` buildpack does not support the tiny stack.

>The tiny build image is already capable of building the Node.js application.

>The tiny run image is missing `libstdc++6`and `libgcc1`.

This RFC should address this.

## Use Cases

Simple node.js app are forced to use the `base` stack with its increased size and CVE surface. Supporting them on `tiny` could be an improvement.

## Checklist

### DIFF
--- a/text/stacks/0000-node-on-tiny.md
+++ b/text/stacks/0000-node-on-tiny.md
@@ -26,18 +26,25 @@ apps/node-base                        latest     022cb9cea147   41 years ago   2
 apps/node-full                        latest     1324267830ec   41 years ago   803MB
 ```
 
+Additionally image scans would find less CVEs which would make it easier to assess.
+
+| Image | Size Run Image (MB) | Total CVEs | CVEs >= 7 |
+| --- | --- | --- | --- |
+| run-node-tiny | 19.2 |  27 | 14 |
+| paketobuildpacks/run:tiny | 17.4 | 28 | 14 |
+| paketobuildpacks/run:base | 86.7 | 58 | 40 |
+
 ## Detailed Explanation
 
 Install the [`libstdc++6`](https://packages.ubuntu.com/bionic/libstdc++6) and [`libgcc1`](https://packages.ubuntu.com/bionic/libgcc1) packages for the Ubuntu based tiny run image. It looks like this will add around 1.8 MB.
 
-```bash
-run-node-tiny                        latest     43acd28c97af   41 years ago   19.2MB
-paketobuildpacks/run:tiny            latest     fa34fc0b3d7b   41 years ago   17.4MB
-```
+The image scan results when writing this pr "only" showed one additional CVE (5.5) which would need to be assessed.
 
 ## Rationale and Alternatives
 
-1. Node is written in C++, so it needs `libstdc++6`and `libgcc1` at runtime.
+- Node is written in C++, so it needs `libstdc++6`and `libgcc1` at runtime.
+- This would also allow other apps to use the tiny stack, e.g. Rust, C++, SAP JVM
+- This might be solved with the use of `extensions` which is currently discussed
 
 ## Implementation
 
@@ -50,4 +57,4 @@ None
 ## Unresolved Questions and Bikeshedding
 
 - Is it worth to influence other users of the stack?
-- Currently the `npm start` buildpack requires a `bash`, so any Node.js application using a start script would fail.
+- Currently the `npm start` buildpack requires a `bash`, so any Node.js application using a start script would fail, but this might be changed in `npm start`
--- a/text/stacks/0000-node-on-tiny.md
+++ b/text/stacks/0000-node-on-tiny.md
@@ -1,0 +1,53 @@
+# Add to Tiny Run Image to Enable Node Apps
+
+## Summary
+
+At the moment, Node.js applications are not compatible with the tiny builder. Therefor the `nodejs` buildpack does not support the tiny stack.
+
+The tiny build image is already capable of building the Node.js application.
+
+The tiny run image is missing `libstdc++6`and `libgcc1`.
+
+## Motivation
+
+To support running Node.js applications with the tiny builder.
+
+While not all Node.js applications could be supported by that, this could be a feasible approach:
+
+- `tiny`: Run Node.js applications **without** any native extenstions
+- `base`: Run Node.js applications **without** many native extensions
+- `full`: Run Node.js applications **with** many native extensions
+
+This should be desirable as some types of Node.js application would benefit of the reduced image size.
+
+```bash
+apps/node-tiny                        latest     7546c8d45780   41 years ago   130MB
+apps/node-base                        latest     022cb9cea147   41 years ago   202MB
+apps/node-full                        latest     1324267830ec   41 years ago   803MB
+```
+
+## Detailed Explanation
+
+Install the [`libstdc++6`](https://packages.ubuntu.com/bionic/libstdc++6) and [`libgcc1`](https://packages.ubuntu.com/bionic/libgcc1) packages for the Ubuntu based tiny run image. It looks like this will add around 1.8 MB.
+
+```bash
+run-node-tiny                        latest     43acd28c97af   41 years ago   19.2MB
+paketobuildpacks/run:tiny            latest     fa34fc0b3d7b   41 years ago   17.4MB
+```
+
+## Rationale and Alternatives
+
+1. Node is written in C++, so it needs `libstdc++6`and `libgcc1` at runtime.
+
+## Implementation
+
+Install the `libstdc++6` and `libgcc1` packages for the Ubuntu based tiny run image.
+
+## Prior Art
+
+None
+
+## Unresolved Questions and Bikeshedding
+
+- Is it worth to influence other users of the stack?
+- Currently the `npm start` buildpack requires a `bash`, so any Node.js application using a start script would fail.
--- a/text/stacks/0000-node-on-tiny.md
+++ b/text/stacks/0000-node-on-tiny.md
@@ -15,7 +15,7 @@ To support running Node.js applications with the tiny builder.
 While not all Node.js applications could be supported by that, this could be a feasible approach:
 
 - `tiny`: Run Node.js applications **without** any native extenstions
-- `base`: Run Node.js applications **without** many native extensions
+- `base`: Run Node.js applications **with** some native extensions
 - `full`: Run Node.js applications **with** many native extensions
 
 This should be desirable as some types of Node.js application would benefit of the reduced image size.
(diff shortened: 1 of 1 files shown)


text/stacks/0000-node-on-tiny.md
```diff
@@ -0,0 +1,53 @@
...
+
+To support running Node.js applications with the tiny builder.
+
+While not all Node.js applications could be supported by that, this could be a feasible approach:
+
+- `tiny`: Run Node.js applications **without** any native extenstions
+- `base`: Run Node.js applications **without** many native extensions
+- `full`: Run Node.js applications **with** many native extensions
+
+This should be desirable as some types of Node.js application would benefit of the reduced image size.
```
loewenstein, line 21: Although this can only be a snapshot, I suppose it is worth pointing out the correctly known CVEs it would remove for Node apps / add for users of the tiny stack.

I guess this is the more crucial metric compared to image size.

text/stacks/0000-node-on-tiny.md
```diff
@@ -0,0 +1,53 @@
...
+Install the `libstdc++6` and `libgcc1` packages for the Ubuntu based tiny run image.
+
+## Prior Art
+
+None
+
+## Unresolved Questions and Bikeshedding
+
+- Is it worth to influence other users of the stack?
+- Currently the `npm start` buildpack requires a `bash`, so any Node.js application using a start script would fail.
```
dmikusa-pivotal, line 53: I want to say yes to this, but these are the two points that I'm stuck on.

1. Adding this won't unlock the ability to run on Tiny for all Node.js use cases, in fact just a fraction because there are limitations like no start scripts & no native extensions. This leaves me wondering if it's worth it since there is an increase we're imposing on other users of about 10% in size.

    Perhaps you could expand more on what use cases this would enable in the motivations section. What situations or use cases are impractical with the present base stack, beyond image size? What could be unlocked by running on Tiny? 

2. I see the work upstream with the buildpacks project on extensions, which once implemented, would be an easy way to solve this problem. Given the increase in size of the tiny image, would it make more sense to recommend Node.js users wait for that work to be available instead of modifying Tiny.

text/stacks/0000-node-on-tiny.md
c0d1ngm0nk3y: We looked at existing RFC and they were written like that. As you mentioned stating the CVE would only be a snapshot anyway.

text/stacks/0000-node-on-tiny.md
c0d1ngm0nk3y: > 1. like no start scripts

We are currently evaluating a potential pr to `npm-start` to not use `bash`. This might open the possibilities to use start scripts with the tiny stack. I agree that without start scripts, the fraction of nodejs apps able to use this stack would probably be very low.

text/stacks/0000-node-on-tiny.md
```diff
@@ -50,4 +57,4 @@ None
 ## Unresolved Questions and Bikeshedding
 
 - Is it worth to influence other users of the stack?
-- Currently the `npm start` buildpack requires a `bash`, so any Node.js application using a start script would fail.
+- Currently the `npm start` buildpack requires a `bash`, so any Node.js application using a start script would fail, but this might be changed in `npm start`
```
loewenstein, line 60:
```suggestion
- Currently the `npm start` buildpack requires a `bash`, so any Node.js application using a start script would fail, but this can be changed in `npm start`
```

Maybe this makes more clear that we actively look into solving thod.

dmikusa-pivotal:
2022-08-02 WG Notes:

- There are open questions around the use case & what workloads this would enable to run on Tiny. See conversations above.
- The team wants to be careful as this adds about 10% size increase to the image.
- We would like to see if there are other use cases outside of Node.js that this would enable to run on Tiny. The `libstdc++6` should be helpful is someone is running a pre-compiled c++ app (Paketo doesn't have a C/C++ buildpack). Are there other runtimes that use C++ that could also take advantage? For example, Ruby, Python, PHP?

text/stacks/0000-node-on-tiny.md
dmikusa-pivotal: 3.) Would it be better to just have the Node.js dist buildpack manage this. For example, if the stack is Tiny. Then install the required libraries with the Node.js distribution. This would enable all the same Node.js use cases. We took a similar tactic in the Rust buildpack. Instead of including the library, it just checks if the stack is Tiny & then compiles out the need for the additional library.

text/stacks/0000-node-on-tiny.md
```diff
@@ -0,0 +1,60 @@
...
+
+Install the [`libstdc++6`](https://packages.ubuntu.com/bionic/libstdc++6) and [`libgcc1`](https://packages.ubuntu.com/bionic/libgcc1) packages for the Ubuntu based tiny run image. It looks like this will add around 1.8 MB.
+
+The image scan results when writing this pr "only" showed one additional CVE (5.5) which would need to be assessed.
+
+## Rationale and Alternatives
+
+- Node is written in C++, so it needs `libstdc++6`and `libgcc1` at runtime.
+- This would also allow other apps to use the tiny stack, e.g. Rust, C++, SAP JVM
+- This might be solved with the use of `extensions` which is currently discussed
```
ekcasey, line 47: Not suggesting that I prefer this, but another alternative in addition to using extensions might be language-specific stacks.

In the past we have floated the idea of introducing language-specific builders in order to avoid the layer limit on builder images. This idea is predicated on the assumption that most users know which composite buildpack they should be using and detecting the difference between say a nodejs and a golang application, isn't really a meaningful feature.

If we had language-specific builders it might make sense to create more variations on the the tiny stack that provide the minimal packages typically needed for each language family.

text/stacks/0000-node-on-tiny.md
fg-j: @dmikusa-pivotal if the approach you're describing is possible for Nodejs, I think that'd be an attractive option.

text/stacks/0000-node-on-tiny.md
fg-j: Given that `jam create-stack` has now simplified the process of creating and maintaining stacks, and Paketo has new [automation](https://github.com/paketo-buildpacks/github-config/tree/main/stack) to facilitate stack maintenance, this is more possible now than it ever has been. However, as a Builders maintainer, I'm hesitant to add many more builder repositories to the official Paketo offerings. The burden on the Paketo buildpacks Github org's resources is already significant. I imagine @paketo-buildpacks/stacks-maintainers may feel the same about adding Paketo stacks.

While I could potentially see adding stacks to `paketo-community` based on user demand, I'm more inclined to enable users to maintain their own stacks.

davidmirror-ops:
WG 08-09-2022 Notes:

FGJ: not sure if it's necessarily feasible to sign off on maintaining more stacks and builders or let users do this themselves with some support from the core team. We're sort of arriving at the point where stacks/builders offerings are a bit too generic
EC: tooling is shared but releasing/maintaining buildpacks is a per language effort. Another downside is platforms (other than pack) that have a default builder. If everyone is excited about minimal images, the best UX would be minimal stacks. Maybe the maintenance is a burden but it's ideal for the end user

joshuatcasey:
2022-08-16 Working Group Notes: Not sure that we want to increase the image size for other language families so that node apps can have a smaller stack.

ForestEckhardt:
WG 2022-08-23 Notes:
Update from stakeholders is upcoming soon. Tested all NodeJS samples in Paketo against tiny++ stack and all seemed to pass.

dmikusa-pivotal:
FYI, I came across this issue today. I'd forgotten about it. A user has reported that one of the JVMs we support, Eclipse OpenJ9, requires this libstdc++6 package as well. That would be another use case we could enable if we added this to the image. https://github.com/paketo-buildpacks/eclipse-openj9/issues/144

c0d1ngm0nk3y:
> Tested all NodeJS samples in Paketo against tiny++ stack and all seemed to pass.

How have you done it? We wanted to do the same, but realised that most of the NodeJS apps use a start script, so `npm-start` will fail because it assumes a `bash`. So we tried 2 things:

1. Change `npm-start` to not use `bash` (see [this pr](https://github.com/paketo-buildpacks/npm-start/pull/224)) and the integration test worked.

2. We built our own stack including in addition a `bash`. Running the integration tests with that builder worked as well.

c0d1ngm0nk3y:
> * The team wants to be careful as this adds about 10% size increase to the image.

I see see the point, but it is "only" 2MB, right? Whereas the resulting image for a simple NodeJS app would 70MB smaller than with `base`.

davidmirror-ops:
WG 2022-08-30 Notes: reach out to Paketo Node users to gauge interest on this. Part of the team in favor of scratch image.

davidmirror-ops:
WG 2022-09-06 Notes: related to a new RFC regarding scratch stack.

sophiewigmore:
I think with the possible introduction of a scratch stack in #243 I'd be more open to this option. I especially like that a new Java use case would also be unlocked with the added package. 

Just wondering, did anyone ever look into[ Dan's idea](https://github.com/paketo-buildpacks/rfcs/pull/231#discussion_r935923948) of having the buildpack just be responsible for installing the package when the Tiny stack is detected?

fg-j:
WG 2022-09-13 Notes:
The intention is to proceed with this RFC. The "static/scratch" stack will help address concerns about the growing Tiny stack.

text/stacks/0000-node-on-tiny.md
robdimsdale: If we proceed with adding these libraries it should eliminate this concern around builder proliferation.

text/stacks/0000-node-on-tiny.md
robdimsdale: My guess is that it isn't going to be easy to dynamically compile nodejs dependencies based on stack, because the nodejs distribution is precompiled, whereas it appears from a quick investigation that the rust distribution is compiled as part of the rustup buildpack's build phase.

Given that Stacks maintainers are happy with adding these libraries to the tiny stack (because we are making progress on creating a separate 'static' stack with fewer libraries), I'm inclined to resolve this conversation.

Any objections to resolving this conversation?

davidmirror-ops:
WG 2022-09-20 Notes:
The `static` RFC was approved and merged, this RFC is open for final reviews.

text/stacks/0000-node-on-tiny.md
dmikusa: It's not a matter of dynamically compiling, but just basically vendoring the required libraries with the buildpack. So if Node.js needs libstdc++6 and libgcc1 then you grab those and include them for all the supported stacks with the buildpack binaries.

At any rate, I'm OK resolving this because of other changes that have occurred recently.

dmikusa (approved):

text/stacks/0000-node-on-tiny.md
loewenstein: I am wondering what the difference between the current packages in tiny and the proposed to be added ones is? I like the idea of buildpacks just bringing their dependencies with them. Like we could use the new static stack for everything if the go-, java-, node-, python- etc. buildpacks would bring all libs they need, wouldn't we?

sophiewigmore (approved):

ryanmoran:
WG 2022-09-27 Notes:
Waiting for more approvals from @paketo-buildpacks/stacks-maintainers

robdimsdale (approved):

brayanhenao (approved):

loewenstein:
Are we good to merge this now with four approvers?

dmikusa:
I think so, and I can merge this.

Let's wait till Tuesday's meeting and just give one last chance for folks to provide any feedback or additional votes.

fg-j:
From 10/4/22 WG meeting: No objections, good to merge!

robdimsdale:
@paketo-buildpacks/steering-committee i think we're good to merge this.

PR #234, opened 2022-08-05
Adds Graceful Stack Upgrades RFC

ryanmoran:
## Summary

[Readable](https://github.com/paketo-buildpacks/rfcs/blob/graceful-stack-upgrades/text/0051-graceful-stack-upgrades.md)

## Checklist


text/0051-graceful-stack-upgrades.md
```diff
@@ -0,0 +1,41 @@
+# Graceful Stack Upgrades
+
+## Summary
+
+All Paketo Buildpacks should support graceful stack upgrades between the
+current Bionic and Jammy stacks. For most buildpacks, this should not require
```
robdimsdale, line 6: Do you have a concrete example of a buildpack (or at least a type of buildpack) that does not work correctly?

I was initially thinking the problem would be dynamic linking against libraries that have changed, but in those cases we have introduced a new dependency with a recompiled binary dynamically linked against the new stack.

ryanmoran: I can update the RFC to call out more specifically where I think the issues are likely to occur. I agree that any buildpack that just installs a dependency will likely not have an issue since the dependency should be guaranteed to run on the stack, or be replaced with another version that should.

For buildpacks with possible issues, I think the canonical example would be instances where a package manager installs dependencies that link out to C code provided by the OS. I don't have an example right in front of me, but the worry would be that say `bundle-install` installs a bunch of gems that have C bindings that are then broken by the stack upgrade.

ryanmoran: RFC updated with more details about the types of buildpacks we will likely need to modify.

robdimsdale (approved):

ekcasey (approved):

fg-j (approved):

dmikusa-pivotal (approved):

davidmirror-ops:
WG 08-09-2022 Notes:
RM: For most buildpacks, the implementation of this proposal shouldn't be an issue. Maybe package-manager or process buildpacks might experience issues, maintainers should decide the behavior the buildpack it should have when the stack ID is changed

matthewmcnew:
> For some buildpacks, we will want to introduce code that tracks the stack ID between builds and recreates layers built with a stack ID that differs from the current CNB_STACK_ID value.

Is there an expectation that for this to work correctly the buildpack would need to run against a cached layer provided by a buildpack that recorded the stack id? Or in the absence of the stack id recorded would the buildpack force the layer to be rebuilt to be conservative?

ryanmoran:
> Is there an expectation that for this to work correctly the buildpack would need to run against a cached layer provided by a buildpack that recorded the stack id? Or in the absence of the stack id recorded would the buildpack force the layer to be rebuilt to be conservative?

@matthewmcnew At the cost of a one-time rebuild, its probably reasonable to go the conservative route and force rebuilding of layers where the stack ID was not previously recorded. I will outline that strategy in the RFC.

matthewmcnew:
> its probably reasonable to go the conservative route and force rebuilding of layers where the stack ID was not previously recorded

Makes sense to me. Would that lead to all layers with this requirement being rebuilt even on an upgrade on the same stack?

ryanmoran:
> Makes sense to me. Would that lead to all layers with this requirement being rebuilt even on an upgrade on the same stack?

Yes. Considering that we are talking about a rather limited set of the buildpacks and that this penalty would occur once, it seems reasonable to choose that more conservative path.

joshuatcasey:
2022-08-16 Working Group Notes: ready to be merged since it's been open long enough to attract comments and approvals.

PR #239, opened 2022-08-23
RFC: Make All Builder Buildpackless

ForestEckhardt:
[Readable](https://github.com/paketo-buildpacks/rfcs/blob/only-buildpackless-builders/text/0000-all-builder-buildpackless.md)

### DIFF
--- a/text/0000-all-builder-buildpackless.md
+++ b/text/0000-all-builder-buildpackless.md
@@ -1,0 +1,76 @@
+# Make All Builders Buildpackless
+
+## Summary
+
+Builders that conglomerate buildpacks are unsustainable in the long run as
+there is a layer cap set in place by Docker of 127 layers. We should stop
+adding buildpack to our builder and instead only offer builders that contain
+the stack and lifecycle. 
+
+## Motivation
+
+With the current size of the `full` builder we are already seeing users running
+into issues with exceeding the layer limit of the buildpack when trying to add
+custom buildpacks to their builds. This will only be exacerbated going forward
+with the addition of new buildpacks into the Paketo project. Not only that the
+`full` builders are completely unsustainable in the near future with proposals
+such as the addition of APMs and the restructuring of existing buildpacks which
+will both increase the layer count further.
+
+Currently the purpose of the builders that include buildpacks is to allow first
+time users to quickly plug in a large number of applications into the build
+process and see what comes out. However using these loaded builders is not the
+optimal solution if you are running builds for production images and the
+builder images are large and bloated for most users who only have applications
+in a couple of language domains.
+
+This layer cap will only continue to become more of a problem and there is no
+clear workaround so unfortunately a drastic restructuring is required.
+
+## Rationale and Alternatives
+
+### Language Specific Builders
+Have one builder per language family that only supports that language family.
+
+#### Pro
+This guarantees that the builder will remain small as they will be confined to
+a much smaller domain.
+
+#### Cons
+There is not much difference not much between this workflow and specifying the
+buildpack that you want during build with a builder that contains no
+buildpacks.
+
+## Implementation
+
+Announce this RFC in as many channels where there are potentially interested
+parties and try to collect as much feedback as possible.
+
+Announce a TBD deprecation window of the buildpack-full builders to allow for
+the following actions:
+- Maintainers spin up language specific builders if the maintainers of the
+  language family buildpack find it necessary.
+- Give integrators time to switch over.
+- Examine our documentation and make any changes that would be required.
+- Write up a builder to buildpack compatibility.
+
+Then the following steps should be made for the transition:
+- The buildpackless builder repos should be archived.
+- The buildpack-full builders should have all buildpacks removed and a major
+  version cut released.
+
+## Prior Art
+
+- Our currently existing buildpackless builders are our prescribed work around
+  when hitting the layer limit.
+
+## Unresolved Questions and Bikeshedding
+
+- How would `pack` suggest a Paketo builder?
+- Will this unduly affect users of the Paketo builders that are using systems
+  that integrate with `pack` and `kpack`?
+- Is there anything that can be done in the upstream?
+- How long should the deprecation window period be?
+- Who needs to be informed?
+
+{{REMOVE THIS SECTION BEFORE RATIFICATION!}}


text/0000-all-builder-buildpackless.md
```diff
@@ -0,0 +1,76 @@
...
+Have one builder per language family that only supports that language family.
+
+#### Pro
+This guarantees that the builder will remain small as they will be confined to
+a much smaller domain.
+
+#### Cons
+There is not much difference not much between this workflow and specifying the
+buildpack that you want during build with a builder that contains no
+buildpacks.
```
robdimsdale, line 42: I think the real "con" here is that we would have to maintain ~10x more builders than we currently do, and as you point out, the effort of finding a particular language builder for an app is about the same level of effort as running a buildpackless builder with a `--buildpack` flag.

samj1912:
@ForestEckhardt I believe an alternative might be for pack to introduce a `--squash` flag in `pack buildpack` and `pack builder` which squashes build layers for a builder or a meta buildpack.

dmikusa-pivotal:
> @ForestEckhardt I believe an alternative might be for pack to introduce a `--squash` flag in `pack buildpack` and `pack builder` which squashes build layers for a builder or a meta buildpack. We should be able to create this RFC upstream for pack.

I had asked about something similar in Slack, but didn't get a response. I would love to see a `--squash` flag.

https://cloud-native.slack.com/archives/C0331B61A1Y/p1660915410266049

Another possibility in the meantime, we could probably start running `crane flatten` on our composite images when we publish them.

```
> # a test copy of the latest paketo-buildpacks/java with 25 layers
> crane manifest docker.io/dmikusa2pivotal/paketo-buildpacks-java:7.0.0 | jq '.layers | length'
25

# flatten it to a new tag (or leave off `-t` to squash it in place)
> crane flatten -t docker.io/dmikusa2pivotal/paketo-buildpacks-java:7.0.0-flat docker.io/dmikusa2pivotal/paketo-buildpacks-java:7.0.0
2022/08/29 14:21:00 pushed blob: sha256:3925dfe4529265c12156ccae6323875c077ac38ab386f25e49ea7e5a05076fee
2022/08/29 14:21:00 existing blob: sha256:3925dfe4529265c12156ccae6323875c077ac38ab386f25e49ea7e5a05076fee
2022/08/29 14:21:01 pushed blob: sha256:7cdc96bc63490357381178d29f6e1c28841183e1dfa1fac978750d27db8d4301
2022/08/29 14:21:01 docker.io/dmikusa2pivotal/paketo-buildpacks-java:7.0.0-flat: digest: sha256:b011d793bdbe82c5cdaf5ea7048c0545eb02ab74e214004a26af68780477d96f size: 430

# down to one layer now
> crane manifest docker.io/dmikusa2pivotal/paketo-buildpacks-java:7.0.0-flat | jq '.layers | length'
1
```

text/0000-all-builder-buildpackless.md
dmikusa-pivotal: From a user stand point, using `-b` is slower. It dynamically creates a new builder every time you do a `pack build`. If you have small buildpacks, like our Paketo Buildpacks, it generally goes quick. If you embed dependencies in your buildpack images, it'll be very, very slow.

dmikusa-pivotal: >I think the real "con" here is that we would have to maintain ~10x more builders than we currently do

This sounds like a process scalability problem. Perhaps it is an opportunity to rearchitect how we manage and ship builders in a more scalable way? Certainly not a short-term solution, but perhaps something to consider.

text/0000-all-builder-buildpackless.md
```diff
@@ -0,0 +1,76 @@
...
+into issues with exceeding the layer limit of the buildpack when trying to add
+custom buildpacks to their builds. This will only be exacerbated going forward
+with the addition of new buildpacks into the Paketo project. Not only that the
+`full` builders are completely unsustainable in the near future with proposals
+such as the addition of APMs and the restructuring of existing buildpacks which
+will both increase the layer count further.
+
+Currently the purpose of the builders that include buildpacks is to allow first
+time users to quickly plug in a large number of applications into the build
+process and see what comes out. However using these loaded builders is not the
```
dmikusa-pivotal, line 22: I think this is still a use case we'd want to support & we should plan how we do that. IMO, we want the first-touch experience with Paketo to be as easy as possible. I think that if we updated our docs/samples/getting started guides with language-specific builder references, the process could still be pretty easy.

The one spot that comes to mind which might not be easy is `pack builder suggest`. I suppose we could ask if we could add the language-specific builders, but even then you'd get a list of 7+ builders and it might be tricky for users to pick one. I don't really know how many users discover us through this mechanism though, or is it more of just a reference for users that already have some familiarity with Paketo, in which case maybe it's not as big of a deal.

text/0000-all-builder-buildpackless.md
```diff
@@ -0,0 +1,76 @@
...
+optimal solution if you are running builds for production images and the
+builder images are large and bloated for most users who only have applications
+in a couple of language domains.
+
+This layer cap will only continue to become more of a problem and there is no
+clear workaround so unfortunately a drastic restructuring is required.
```
dmikusa-pivotal, line 28: I'm pretty sure you can safely squash buildpack layers, at least from the standpoint that buildapack A should not contain files that overlap with files from buildpack B or any other buildpack. Thus squashing them down should be safe.

A quick glance, see my other comment, it does look like we'd need a custom tool though because buildpack metadata in the labels would need to be adjusted, or we'd need proper support in `pack`.

If you could squash layers, you could have one per composite, which would lower the limit to the point where the max layer count would never be an issue.

ForestEckhardt:
I would be happy to try and work this up the chain into the upstream. I would like to discuss this idea a bit more at the working group meeting if at all possible.

text/0000-all-builder-buildpackless.md
ekcasey: I see the performance implications of always using `-b` as a pretty sizeable con. We have a lot of evidence that performance is a major concern to buildpack users and one of the first things potential users evaluate when comparing tools.

ekcasey: > I think the real "con" here is that we would have to maintain ~10x more builders than we currently do

I agree with Dan that this sounds like a process/automation problem? Whether we publish the language-specific builders or only buildpackless builders we still need to test and maintain the same combinatorial matrix of buildpack+stack+lifecycle. Ideally, publish more images to improve performance for end users shouldn't materially increase increase that burden.

davidmirror-ops:
WG 2022-08-30 Notes: 
[EC] uncomfortable with just buildpackless due to potential performance issues. We could start with squashing.
[SK] We could tackle the layer limit issue and probably that would relieve the need for a proposal like this. Not sure how this proposal could affect polyglot apps

Right now the move to language specific builders is preeminent, so probably next step would have to be to close this proposal. Next step could be creating an upstream issue

PR #241, opened 2022-08-28
Add an RFC about builder config

samj1912:
Signed-off-by: Sambhav Kothari <skothari44@bloomberg.net>

## Summary

This RFC proposes an easy way to configure builders to allow specifying a `config.toml` file that allows updating the Buildpack `detect` and `build` environment based on the configuration file.

[Readable](https://github.com/samj1912/rfcs-1/blob/builder-config/text/0000-builder-config.md)

## Checklist

### DIFF
--- a/text/0000-builder-config.md
+++ b/text/0000-builder-config.md
@@ -16,7 +16,7 @@ Some examples include -
 
 The buildpack logic in `Paketo` largely remains the same, except these environment variables might need to be injected during the `build` and `detect` phases.
 
-The environment variables may ideally also take precendence over any user provided values to ensure that the operators have full control over their builders.
+The environment variables may be able to take precedence over user-provided values, if configured to, which ensures that operators can take full control over their builders if that is required.
 
 ## Detailed Explanation
 
--- a/text/0000-builder-config.md
+++ b/text/0000-builder-config.md
@@ -1,0 +1,120 @@
+# Builder Config
+
+## Summary
+
+This RFC proposes an easy way to configure builders to allow specifying a `config.toml` file that allows updating the Buildpack `detect` and `build` environment based on the configuration file.
+
+## Motivation
+
+Often times, especially in enterprise settings, organizations often have to update the buildpacks to use internal mirrors, proxies and other settings which can be configured easily with environment variables.
+
+Some examples include -
+- `GOPROXY`
+- `PIP_INDEX_URL`
+- `npm_config_registry`
+- `BUNDLE_MIRROR__URL`
+
+The buildpack logic in `Paketo` largely remains the same, except these environment variables might need to be injected during the `build` and `detect` phases.
+
+The environment variables may ideally also take precendence over any user provided values to ensure that the operators have full control over their builders.
+
+## Detailed Explanation
+
+The RFC proposes the introduction of the following file `/builder/config.toml`.
+
+The `config.toml` has an optional section `build.env`. This section allows defining key/value pairs of environment variables.  Environment variables are defined with optional suffixes.  The following table describes the possible definitions for an environment variable called `KEY`:
+
+```toml
+api = "0.1"
+
+[[build.env]]
+name = "KEY"
+# If value is not specified, it is assumed to be an empty string.
+value = "VALUE"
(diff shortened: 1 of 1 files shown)


loewenstein:
I am wondering about two things.
First, I am very skeptical about a precedence order that does not prioritize direct user input 
Second, and maybe even more important, shouldn't this be proposed via CNB spec?

samj1912:
> I am wondering about two things. 

> First, I am very skeptical about a precedence order that does not prioritize direct user input 

It gives builder/operators a choice which they currently don't have very easily. If they wish to prioritize user input they can use `default`. If they wish to override it, they can use `override`.  The fact that builder/buildpacks cannot control user input can be a major security concern. IMHO It should be entirely upto the buildpack and the builder operator to enforce this. This is no different than a buildpack using `clear-env` flag to omit using user provided platform env vars or a buildpack choosing to override/ignore platform env vars. 


> Second, and maybe even more important, shouldn't this be proposed via CNB spec?

It can be. CNB changes take a while to trickle down/be implemented. This is also something that the lifecycle currently allows buildpacks to achieve with the existing API,so I don't see why it can't be implemented at a paketo level.

samj1912:
Created https://github.com/buildpacks/rfcs/pull/230 in case there is appetite for it upstream.

text/0000-builder-config.md
```diff
@@ -0,0 +1,120 @@
...
+
+Some examples include -
+- `GOPROXY`
+- `PIP_INDEX_URL`
+- `npm_config_registry`
+- `BUNDLE_MIRROR__URL`
+
+The buildpack logic in `Paketo` largely remains the same, except these environment variables might need to be injected during the `build` and `detect` phases.
+
+The environment variables may ideally also take precendence over any user provided values to ensure that the operators have full control over their builders.
```
dmikusa-pivotal, line 19:
```suggestion
The environment variables may be able to take precedence over user-provided values, if configured to, which ensures that operators can take full control over their builders if that is required.
```

I would suggest this because I don't think this should be the default value. I think by default, user-configured values should be the highest precedence. That said, you have the `override` mode, so if set to `override` then an operator could take control of that value.

My rationale for this is a.) we do that in pretty much every other situation, so I would say it's the generally expected behavior b.) I think it would be difficult to debug if something wasn't working right. To know that you're setting an env variable as a user and it's getting overridden by the stack.

I would further suggest that when the mode is set to `override` for a builder env variable we should log a message (probably debug, but possibly info) to indicate this happened.

dmikusa-pivotal:
>> Second, and maybe even more important, shouldn't this be proposed via CNB spec?
>
>It can be. CNB changes take a while to trickle down/be implemented. This is also something that the lifecycle currently allows buildpacks to achieve with the existing API,so I don't see why it can't be implemented at a paketo level if we can get immediate benefits from it.

This is my concern as well. Should we do this at the Paketo level, or wait for it to be done in the lifecycle? I'm not opposed to implementing it now in Paketo **if** we can do it in such a way that we could later back this out of Paketo and just let the lifecycle handle it. If we can't, my fear is that we could end up stuck with something we have to support in Paketo that works one way and then later have a very similar, but not the same bit of functionality in lifecycle, which is very confusing for users.

davidmirror-ops:
WG 2022-08-30 Notes: 
[JL] should we push it upstream? 
[SK] Doesn't truly matter where it's implemented: lifecycle or Paketo
Cautions about security (secrets/bindings) and `pack`

davidmirror-ops:
WG 2022-09-06 Notes: there seems to be response from the CNB project, keeping an eye on to see if this RFC is needed here

fg-j:
WG 2022-09-13 Notes:
Given that this idea is gaining traction upstream, Paketo is reluctant to implement at the buildpack level. This RFC is on hold until the CNB project's intentions around this proposal become more clear.

davidmirror-ops:
WG 2022-09-20 Notes: no updates

ryanmoran:
WG 2022-09-27 Notes:
@samj1912 can this be moved back into a draft state since the upstream RFC seems to be making some progress?

samj1912:
Upstream rfc accepted.

PR #243, opened 2022-08-31
Add static stack RFC

robdimsdale:
## Summary

An RFC to propose adding support for a ~scratch~ static stack. [Readable](https://github.com/paketo-buildpacks/rfcs/blob/scratch-stack/text/0000-add-scratch-stack.md).


text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,161 @@
+# Create scratch stack
+
+## Summary
+
+The Paketo stacks maintainers should create and maintain a scratch stack that
+can run statically-linked applications (e.g. golang). The Builder team does not
```
dmikusa, line 6:
```suggestion
can run statically-linked applications (e.g. golang, Rust). The Builder team does not
```

robdimsdale: Happy to add Rust in the summary. Initially I omitted it from the summary because it's not a community-supported language whereas golang is a fully-supported buildpack.

davidmirror-ops:
WG 2022-09-06 Notes: feedback is needed, including about the "scratch" name (could it be confusing?)

text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,161 @@
...
+packages - in turn reducing the surface area for security issues.
+
+Currently, in order to build these statically-linked applications using Paketo
+buildpacks (and aiming for the smallest possible resultant run image) users
+must either opt for the [tiny
+stack](https://github.com/paketo-buildpacks/jammy-tiny-stack) or build their
+own scratch stack. We have evidence of users doing both of these things.
+Neither are satisfactory to users, however, as the tiny stack contains more
+packages than is necessary for those applications, and users would prefer not
+to maintain their own stacks if Paketo is willing and able to do so.
```
sophiewigmore, line 32: I would argue though, that maintaining a scratch stack is not that heavy of a burden - considering there are few to no packages in it. By that same reasoning, as a stacks maintainer I'm probably fine with adopting a scratch stack in the project.

text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,161 @@
...
+We will create a new Github repository at
+`paketo-buildpacks/jammy-scratch-stack` and a new pair of images on Dockerhub
+at `paketobuildpacks/build-jammy-scratch` and
+`paketobuildpacks/run-jammy-scratch`.
+
+We will adopt the tag system used for Jammy - specifically semver (i.e.
+'X.Y.Z') and 'latest' tags.
+
+We will not create a builder repository or image, though we reserve the option
+for the Builder team to do so in the future if that is requested and approved.
```
sophiewigmore, line 57: I'm kind of wondering why no builder will be created? Users of the scratch stack will then have to create their own builder. To me, if we're going to support this we might as well just support it end-to-end.

text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,161 @@
...
+### Image Contents
+
+The scratch stack run image will contain the `tzdata` and `ca-certificates`
+packages, the standard CNB/Paketo metadata (including users and /etc/passwd
+entries), and nothing else. These two packages are often provided in scratch
+images as they are small and have minimal security risk.
+
+The build image will contain the same packages as the tiny stack (probably with
+the addition of the packages proposed in [the tiny packages
+RFCS](https://github.com/paketo-buildpacks/rfcs/pull/231/files)).
```
sophiewigmore, line 73: Is there a clever way we could just reuse the tiny build image for the scratch build image, rather than making a duplicate?

sophiewigmore: Ah, I guess this isn't that useful because we'll still need metadata changes.

robdimsdale: Currently `jam create-stack` requires the two `Dockerfile`s on disk to build the images. We could explore a way that it pulls an existing image rather than rebuilding a new one, but that's a feature that doesn't currently exist.

Beyond that, we could use symlinks (e.g. `../../../../../jammy-tiny-stack/stack/build.Dockerfile` but I think that's a really bad solution.

text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,161 @@
...
+Various literature (e.g.
+[this](https://blog.baeke.info/2021/03/28/distroless-or-scratch-for-go-apps/),
+and
+[this](https://chemidy.medium.com/create-the-smallest-and-secured-golang-docker-image-based-on-scratch-4752223b7324)
+describe the use of "scratch" images for statically-linked golang applications,
+with or without the `tzdata` and `ca-certificates` packages.
+
+## Unresolved Questions and Bikeshedding
+
+* Is the name "scratch" misleading, given we propose adding one or two
```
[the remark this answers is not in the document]
sophiewigmore: +1 scratch is misleading I think

text/0000-add-scratch-stack.md
robdimsdale: I mostly didn't want to sign the builder team up for work in this RFC. I agree that it is easier to create a scratch Builder with the same process we currently use for other builders. I could see value in at least a buildpackless builder.

@paketo-buildpacks/builders-maintainers do you think the burden of creating a buildpackless builder for the proposed scratch image is acceptable and something you would be willing to take on?

text/0000-add-scratch-stack.md
robdimsdale: I think it is somewhat unreasonable to ask Paketo consumers who want to build the smallest-possible images from their statically-linked golang/rust applications to maintain their own stack and builder for this purpose. Application developers shouldn't need to know about stacks and builders to be able to use `pack build` (or equivalent).

The corollary of that is that it is a significantly lower burden for stacks maintainers to create the stack and keep it up to date - as you point out.

text/0000-add-scratch-stack.md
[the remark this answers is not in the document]
robdimsdale: Thanks for the feedback. I'm inclined to agree, but naming is hard.

I'd like to avoid increasing the scope of this RFC to include multiple language-family-specific stacks (e.g. `tiny-static`, `tiny-java`) for the reasons you point out @loewenstein .

Ideally I would like to see a name that is distinct from `tiny`. My preference is for `static` as that is widely adopted in the distroless context, and as such will be familiar to the target audience for this stack.

I'll update the RFC once we get to consensus.

text/0000-add-scratch-stack.md
sophiewigmore: Yea, the build image Dockerfile could possibly have `FROM paketobuildpacks/build-jammy-tiny` maybe in its Dockerfile.

robdimsdale: Yes, if `jam create-stack` were able to correctly update the metadata. Which I'm not sure it can do in the general case. What is a reasonable scope of existing metadata for `jam` to know about and overwrite? E.g. if the previous cnb user had UID of 1002 at build time and the new image should be 1004, how would `jam` know to remove the old one?

The more I think about it, the less in favor of this feature I am. I think it's reasonable to keep the `stack.toml` and `Dockerfile`s in sync some other way and not modify `jam` at all.

loewenstein: I agree with @robdimsdale here. However, I tried to [start a conversation](https://cloud-native.slack.com/archives/C033DV9EBDF/p1662042914368639) with upstream CNB if we could loosen the cardinality of stacks (in particular run images) and builders somehow.

I can imagine there are huge overlaps between tiny and base build image as well. Wouldn't it be great to have an Ubuntu Jammy Stack with multiple build and run images and builders could select one of the build images, but different run images per e.g. order group.

We could reduce the number of stacks we need to maintain drastically, i.e. to a single one and potentially even the number of builders (with the use of extensions we might even be able to bring it down to a single builder). It would need a change of the CNB spec though. WDYT? I'd be very interested if this is just impossible to do / has been thought but was dropped / crazy thoughts or if it would be worth pursuing?

robdimsdale: I like this idea of optimizing for a shared `build` image with tailored `run` images though I haven't fully thought through the ramifications yet.

We could do this today at the builder level by having multiple builders, each with different `run` images that share the same stack ID, but I guess your point is that having something more clear in the spec could make this more discoverable and extensible.

robdimsdale: Whether we pursue that idea or not, I don't think we should block this RFC on that conversation

text/0000-add-scratch-stack.md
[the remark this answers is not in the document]
modulo11: Maybe using one of the [metric prefixes](https://en.wikipedia.org/wiki/Metric_prefix#List_of_SI_prefixes) would come in handy when naming the stack. Something like `paketo-buildpacks/jammy-nano-stack` leaving room for stacks containing more or less packages.

text/0000-add-scratch-stack.md
fg-j: Given that we'd only be adding one builder (Jammy, with no buildpacks), I'd be open to the addition. The maintenance burden here would be extremely low and it'd smooth the onboarding process for static-stack users. @paketo-buildpacks/builders-maintainers any other opinions on this?

text/0000-add-scratch-stack.md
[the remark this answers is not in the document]
robdimsdale: @modulo11 I see the value in that scheme. It offers flexibility in sizing and doesn't lock us into any particular set of packages.

However, I think I prefer the `static` name because it already has meaning with the target audience for this stack (developers of statically-linked applications looking for a minimal image to run their applications on).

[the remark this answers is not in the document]
robdimsdale: Let's go with `static`. I'll update the RFC contents, but I'll leave the filename as `scratch` until we merge the PR so as not to lose the comment history.

text/0000-add-scratch-stack.md
robdimsdale: I'll update the text to say that the builder team will make a buildpackless builder. @paketo-buildpacks/builders-maintainers can still voice any objections to the new text :)

text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,159 @@
+# Create static stack
+
+## Summary
+
+The Paketo stacks maintainers should create and maintain a stack (called
+"static") that can run statically-linked applications (e.g. golang, Rust).
+The Paketo builder maintainers will make a buildpackless builder for this
```
robdimsdale, line 7: @paketo-buildpacks/builders-maintainers any objections to creating and maintaining a new buildpackless builder?

ForestEckhardt: I think that if we are going to build out a static builder at all, we would also want a buildpackless builder for consistency sake. It sounds like to support Go we need a builder therefore we should probably have a buildpackless builder.

text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,159 @@
+# Create static stack
+
+## Summary
+
+The Paketo stacks maintainers should create and maintain a stack (called
+"static") that can run statically-linked applications (e.g. golang, Rust).
```
ForestEckhardt, line 6:
```suggestion
"static") that can run statically-linked applications (e.g. Go, Rust).
```
Nitpick: Just in case Rob Pike finds this RFC

text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,159 @@
...
+The Paketo builder maintainers will make a buildpackless builder for this
+stack.
+
+Additionally, the golang buildpack should gain support for this new static
```
ForestEckhardt, line 10:
```suggestion
Additionally, the Go buildpack should gain support for this new static
```

text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,159 @@
...
+stack, both in order to provide a useful experience to golang application
```
ForestEckhardt, line 11:
```suggestion
stack, both in order to provide a useful experience to Go application
```

text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,159 @@
...
+developers as well as to enable integration tests for this new stack.
+
+## Motivation
+
+Some applications consist of a single
+[statically-linked](https://en.wikipedia.org/wiki/Static_library) executable.
+This is most common in golang, but is also often utilized in rust. It is also
```
ForestEckhardt, line 18:
```suggestion
This is most common in Go, but is also often utilized in rust. It is also
```

text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,159 @@
...
+
+Although there are few CVEs for the `tzdata` and `ca-certificates` packages,
+and hence the run image will be updated infrequently, we want to ensure the
+build image is also kept up to date with CVE fixes for all of its packages.
+
+Also, on a practical note, leveraging the existing automation that is used in
+all other stacks is simpler for the stacks maintainers than creating a new
+system just for this stack.
+
+### Golang Buildpack
```
ForestEckhardt, line 95:
```suggestion
### Go Buildpack
```

text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,159 @@
...
+
+We will add support for this new stack in the golang buildpack. This will
```
ForestEckhardt, line 97:
```suggestion
We will add support for this new stack in the Go buildpack. This will
```

text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,159 @@
...
+enable the following:
+
+1. Golang application developers can utilize this new stack out-of-the-box,
```
ForestEckhardt, line 100:
```suggestion
1. Go application developers can utilize this new stack out-of-the-box,
```

text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,159 @@
...
+   without having to modify the Golang buildpack to support this new stack.
```
ForestEckhardt, line 101:
```suggestion
   without having to modify the Go buildpack to support this new stack.
```

text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,159 @@
...
+1. The static stack can use the Golang buildpack for its integration tests.
```
ForestEckhardt, line 102:
```suggestion
1. The static stack can use the Go buildpack for its integration tests.
```

text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,159 @@
...
+   This will enable meaningful end-to-end integration tests for this stack. We
+   want to be able to build and run a test application during the stack
+   integration test, and golang is the best choice for this.
```
ForestEckhardt, line 105:
```suggestion
   integration test, and Go is the best choice for this.
```

text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Rationale and Alternatives
+
+1. We could not create a static stack at all, and continue adding more
+   packages to the tiny stack. This would likely be unaceptable to developers
+   of applications that can be statically-linked, and will likely turn them
+   away from the Paketo project. We have already heard feedback that the tiny
+   stack is larger than some golang developers would like and is already a
```
ForestEckhardt, line 113:
```suggestion
   stack is larger than some Go developers would like and is already a
```

text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,159 @@
...
+   reason not to adopt Paketo Buildpacks.
+1. We could keep the tiny stack as-is - not consider adding further packages to
+   it. This would satisfy current users at the expense of making the tiny stack
+   more useful. It also does not make Paketo more attractive to users who find
+   the tiny stack is already too large for their needs.
+1. We could create the new stack but not add support in the Golang buildpack.
```
ForestEckhardt, line 119:
```suggestion
1. We could create the new stack but not add support in the Go buildpack.
```

text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,159 @@
...
+   would be almost identical. Additionally, Bionic will be end-of-life in April
+   2023 so this proposed stack would only be present for a few months.
+
+## Implementation
+
+1. Create a new Github repository at: `paketo-buildpacks/jammy-static-stack`
+1. Copy most of the existing contents from `paketo-buildpacks/jammy-tiny-stack` into this repository.
+1. Update metadata (e.g. stack name, Support URL, etc).
+1. Remove all packages from the `run` image package list except `tzdata` and `ca-certificates`.
+1. Ensure the integration tests statically links the golang test application.
```
ForestEckhardt, line 135:
```suggestion
1. Ensure the integration tests statically links the Go test application.
```

text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,159 @@
...
+
+Some Paketo consumers already use an image with minimal packages installed in
+commercial/internal environments. Their feedback is that a static image with
+`tzdata` and `ca-certificates` packages is sufficient for their use-cases.
+
+Various literature (e.g.
+[this](https://blog.baeke.info/2021/03/28/distroless-or-scratch-for-go-apps/),
+and
+[this](https://chemidy.medium.com/create-the-smallest-and-secured-golang-docker-image-based-on-scratch-4752223b7324)
+describe the use of "static" images for statically-linked golang applications,
```
ForestEckhardt, line 148:
```suggestion
describe the use of "static" images for statically-linked Go applications,
```

text/0000-add-scratch-stack.md
robdimsdale: @ForestEckhardt just to be clear, your comment above sounds like a tacit approval of supporting a jammy-static-buildpackless-builder, yes?

ForestEckhardt: Yes, I approve of the buildpackless builder

sophiewigmore (approved):

robdimsdale:
@paketo-buildpacks/steering-committee we have approvals from a majority of stacks maintainers, and there was no additional feedback from the community in the last working group meeting. Can we merge this?

⚠️ When we merge the PR and rename the file, can we rename it to: **1234-add-static-stack.md** to reflect the change in name from `scratch` to `static`.

ryanmoran (approved):

text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,161 @@
...
+This may require approval from the upstream Cloud-Native Buildpacks project.
+
+### Image Contents
+
+The scratch stack run image will contain the `tzdata` and `ca-certificates`
+packages, the standard CNB/Paketo metadata (including users and /etc/passwd
+entries), and nothing else. These two packages are often provided in scratch
+images as they are small and have minimal security risk.
+
+The build image will contain the same packages as the tiny stack (probably with
```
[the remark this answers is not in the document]
robdimsdale: I don't think it will save much work because pragmatically we will leverage `jam create-stack` which I believe requires a `Dockerfile` for both `build` and `run`, and will build both images every time.

We could explore a way to instruct `jam create-stack` to instead use an upstream image, but I think that's out of the scope of this RFC.

text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,161 @@
...
+enable the following:
+
+1. Golang application developers can utilize this new stack out-of-the-box,
+   without having to modify the Golang buildpack to support this new stack.
+1. The scratch stack can use the Golang buildpack for its integration tests.
+   This will enable meaningful end-to-end integration tests for this stack. We
+   want to be able to build and run a test application during the stack
+   integration test, and golang is the best choice for this.
+
+## Rationale and Alternatives
```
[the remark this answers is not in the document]
robdimsdale: I don't think that it would be a good idea to make different images under the same stack ID because I think it would be confusing to say: "this buildpack works with the XYZ stack except if you provide XYZ stack with ABC run image swapped out".

Maybe my mental model of stacks is wrong, but I've found it helpful to think of stacks as a set of interfaces - i.e. a list of ABI-compatible libraries on top of a specific distribution. I know that full and base have the same stack IDs and honestly I think that's confusing too.

I think it is more intuitive to be able something like the following:
* go and rust work on `static`, `tiny`, `base`, and `full`
* java (and hopefully soon nodejs too) work on `tiny`, `base`, and `full`
* python works on `base`, and `full`
* php works on only on `full`

I think the best way to express that compatibility relationship between buildpacks and stacks is programmatically, which means in the `buildpack.toml`.

Also, practically, the tooling we have today won't make it particularly easier to achieve what you described. `jam create-stack` will still create a stack with a pair of images, so we'd have to discard the build image at the builder level.

I think this comes down to what one considers to be the interface of a given stack identifier. Perhaps you have a different set of expectations around stacks?

text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,161 @@
...
+for the Builder team to do so in the future if that is requested and approved.
+
+### Stack ID
+
+We propose creating a new stack ID: `io.buildpacks.stacks.jammy.scratch`.
+This may require approval from the upstream Cloud-Native Buildpacks project.
+
+### Image Contents
+
+The scratch stack run image will contain the `tzdata` and `ca-certificates`
```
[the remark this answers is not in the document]
robdimsdale: This [blog post](https://blog.baeke.info/2021/03/28/distroless-or-scratch-for-go-apps/) is a little old, but I think the information it contains is still relevant.

My personal (historic) experience aligns with what I've read on this topic, which is that quite a few Go applications use `tzdata` without necessarily realizing it, and they need `ca-certificates` to make SSL based requests. I don't actually know how Go handles the trust store, but it seems to Just Work™ with the `ca-certificates` package installed.

I could try to construct an example of a statically linked Go app that requires these packages if you want, but it'll take a bit of time.

text/0000-add-scratch-stack.md
```diff
@@ -0,0 +1,161 @@
...
+
+We will adopt the tag system used for Jammy - specifically semver (i.e.
+'X.Y.Z') and 'latest' tags.
+
+We will not create a builder repository or image, though we reserve the option
+for the Builder team to do so in the future if that is requested and approved.
+
+### Stack ID
+
+We propose creating a new stack ID: `io.buildpacks.stacks.jammy.scratch`.
```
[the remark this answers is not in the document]
robdimsdale: I modeled this on the `jammy` stack.

I initially thought that Paketo stacks should be under a paketo namespace, but we decided against that when we created `jammy`.

PR #246, opened 2022-08-31
RFC: New Dependency Strategy for .NET

ForestEckhardt:
[Readable](https://github.com/paketo-buildpacks/rfcs/blob/dotnet-core-depenendencies-rework/text/dotnet-core/0007-dependencies.md)

### DIFF
--- a/text/dotnet-core/0007-dependencies.md
+++ b/text/dotnet-core/0007-dependencies.md
@@ -41,17 +41,18 @@ and using the information given on the version release pages to find an
 official Microsoft release URL. Currently the .NET SDK [upload workflow
 directly uploads the artifact from
 Microsoft](https://github.com/paketo-buildpacks/dep-server/blob/d7402591d0581a5019b1bd620ed1367d5f155213/.github/data/dependencies.yml#L27).
-Therefore, this RFC proposes removing the .NET SDK as a Paketo-hosted
-dependency.
+This is in accordance with [RFC
+0004](https://github.com/paketo-buildpacks/rfcs/blob/main/text/dotnet-core/0004-extend-dotnet-sdk.md)
+in which we decided to consume the .NET SDK directly from upstream. Therefore,
+this RFC proposes removing the .NET SDK as a Paketo-hosted dependency.
 
 ### .NET Core ASP.NET Core Runtime
 
 This is contingent on the implmentation of [.NET RFC
 0006](https://github.com/paketo-buildpacks/rfcs/blob/1d615afaa355f235b216a8fa9346d227299b388f/text/dotnet-core/0005-simplify-runtime-dependency.md)
 in which we will start to consume the .NET Core ASP.NET Core Runtime as it is
-provided by Microsoft. This will be achieved in the same wat as the .NET Core
-SDK was the release index leads to a release page that also contains the URLs
-for the Microsoft hosted dependnecy.
+provided by Microsoft. This will be achieved in the same way as the .NET Core
+SDK by obtaining the upstream download link from the release index.
 
 ### vsdbg
 
--- a/text/dotnet-core/0007-dependencies.md
+++ b/text/dotnet-core/0007-dependencies.md
@@ -1,0 +1,70 @@
+# Decide Which .NET Dependencies Will Be Paketo-hosted
+
+## Proposal
+
+The following dependencies should be kept as Paketo-hosted dependencies:
+* [ICU](https://github.com/paketo-buildpacks/icu/blob/main/buildpack.toml)
+
+
+The following dependencies should be removed as Paketo-hosted dependencies:
+* [.NET Core SDK](https://github.com/paketo-buildpacks/dotnet-core-sdk/blob/main/buildpack.toml)
+
+The following will be a new dependencies that are not currently Paketo-hosted
+and should not be added:
+* .NET Core ASP.NET Core Runtime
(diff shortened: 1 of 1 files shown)


davidmirror-ops:
WG 2022-09-06 Notes: open for feedback

text/dotnet-core/0007-dependencies.md
```diff
@@ -0,0 +1,70 @@
...
+
+### ICU
+
+Keep this as a Paketo-hosted dependency.
+
+ICU currently [only builds for one version of
+Ubuntu](https://github.com/unicode-org/icu/releases/tag/release-71-1) therefore
+in order to support a wide range of stacks we should continue to compile and
+host this independent of upstream. The current build process can be seen
+[here](https://github.com/cloudfoundry/buildpacks-ci/blob/7c76257aa285bf148767a14dd62953b1f0b163b0/tasks/build-binary-new/builder.rb#L462).
```
sophiewigmore, line 27: I think you're probably right about continuing to compile it. 

I am wondering though if we are doing something strange by compiling all new versions of ICU for different stacks. Would a user of ICU in the wild even want/be able to use a new version of ICU on an old OS? 

Also fwiw, it does look like ICU dropped a Jammy Jellyfish-compatible version in June 2022: https://launchpad.net/ubuntu/+source/icu (70.1 version line)

text/dotnet-core/0007-dependencies.md
```diff
@@ -0,0 +1,70 @@
...
+
+Currently ICU uses `dep-server`, `binary-builder`, and `buildpacks-ci` to
+build, but language family maintainers will transition this to the new Github
+Action workflow once that has been approved.
+
+### .NET Core SDK
+
+Remove the Paketo-hosted dependency.
+
+It looks like .NET Core SDL is downloaded directly from a Microsoft by scanning
```
sophiewigmore, line 37:
```suggestion
It looks like .NET Core SDK is downloaded directly from a Microsoft by scanning
```

text/dotnet-core/0007-dependencies.md
```diff
@@ -0,0 +1,70 @@
...
+their [release
+index](https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/releases-index.json)
+and using the information given on the version release pages to find an
+official Microsoft release URL. Currently the .NET SDK [upload workflow
+directly uploads the artifact from
+Microsoft](https://github.com/paketo-buildpacks/dep-server/blob/d7402591d0581a5019b1bd620ed1367d5f155213/.github/data/dependencies.yml#L27).
+Therefore, this RFC proposes removing the .NET SDK as a Paketo-hosted
+dependency.
```
sophiewigmore, line 45: Not to mention that in https://github.com/paketo-buildpacks/rfcs/blob/main/text/dotnet-core/0004-extend-dotnet-sdk.md, we decided to start using the dependency directly from Microsoft as well. Maybe we should mention that here?

text/dotnet-core/0007-dependencies.md
```diff
@@ -0,0 +1,70 @@
...
+
+### .NET Core ASP.NET Core Runtime
+
+This is contingent on the implmentation of [.NET RFC
+0006](https://github.com/paketo-buildpacks/rfcs/blob/1d615afaa355f235b216a8fa9346d227299b388f/text/dotnet-core/0005-simplify-runtime-dependency.md)
+in which we will start to consume the .NET Core ASP.NET Core Runtime as it is
+provided by Microsoft. This will be achieved in the same wat as the .NET Core
+SDK was the release index leads to a release page that also contains the URLs
+for the Microsoft hosted dependnecy.
```
sophiewigmore, line 54: This sentence is a little confusing

fg-j:
As mentioned in Working Group, we should investigate an artifact verification method for each of these. Are checksums available? Is another form of verification possible?

sophiewigmore:
@fg-j @ForestEckhardt it looks like the SHA512 is published. We can still calculate our own SHA256 and use it for the purposes of making sure that Microsoft doesn't change the dependency for a version under our noses. We'll lose the extra usefulness of using a SHA to cross-check what we've downloaded, unless we can do something clever with the SHA512.

ForestEckhardt:
Could we not just download the package, verify it against the sha512, and then calculate the sha256 once we hav verified the package and use that sha256 going forward?

ForestEckhardt:
I also just put this forward as a POC/Proposal in `packit` https://github.com/paketo-buildpacks/packit/pull/389

text/dotnet-core/0007-dependencies.md
ForestEckhardt: I think that I fixed this please let me know if it is still rough.

sophiewigmore:
@ForestEckhardt The POC is a great idea

text/dotnet-core/0007-dependencies.md
fg-j: Since we currently make new versions of ICU available on Bionic, I think we should continue doing that until Bionic goes out of support. (That way, we don't have to downgrade the available versions of ICU.) It's worth investigating whether users will still expect the latest ICU versions on Bionic when there's another LTS version of Ubuntu available.

davidmirror-ops:
WG 2022-09-20 Notes:
This proposal has a dependency on a buildpack implementation (`packit`), open PR in the jam repository.  .NET maintainers team to review

fg-j (approved):

sophiewigmore (approved):

PR #247, opened 2022-09-06
RFC: New Dependency Strategy for Go

ForestEckhardt:
[Readable](https://github.com/paketo-buildpacks/rfcs/blob/rework-go-dependnecies/text/go/0002-dependencies.md)

### DIFF
--- a/text/go/0002-dependencies.md
+++ b/text/go/0002-dependencies.md
@@ -1,0 +1,20 @@
+# Decide Which Go Dependencies Will Be Paketo-hosted
+
+## Proposal
+
+The following dependencies should be removed as Paketo-hosted dependencies:
+* [Go Distribution](https://github.com/paketo-buildpacks/go-dist/blob/main/buildpack.toml)
+
+
+## Rationale
+
+### Go Distribution
+
+Remove the Paketo-hosted dependency.
+
+The buildpack can use the dependency that is provided by Go on their [download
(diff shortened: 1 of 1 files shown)


fg-j:
As mentioned in the Working Group meeting, I think it'll be important to set up some way to verify that the downloaded artifact is authentic. This might involve scraping and using those checksums from the Downloads page. Hopefully there's a cleaner way.

ryanmoran:
We should use this JSON feed instead of screen-scrapping the HTML page: https://go.dev/dl/?mode=json&include=all

fg-j:
@ryanmoran for future reference, where is the JSON feed you linked discoverable?

ryanmoran:
@fg-j I have no idea where or if at all it is documented. I know about it because my friend, @zachgersh, implemented it (https://github.com/golang/tools/pull/21) back when we were both working on Cloud Foundry together.

zachgersh:
Wanted to also drop a link to the godoc that was added when all of this got refactored and expanded (since you asked for future discoverability).

https://pkg.go.dev/golang.org/x/website/internal/dl

Thanks for the callout @ryanmoran

ForestEckhardt:
@fg-j @ryanmoran I have reworked the RFC. I now specify that it should use the JSON payload page and that on that page that is a provided SHA256 for every artifact.

fg-j (approved):

ryanmoran (approved):

PR #258, opened 2022-11-01
Adds Kerberos Buildpack RFC

ryanmoran:
## Summary

[Readable](https://github.com/paketo-buildpacks/rfcs/blob/kerberos-buildpack/text/0054-kerberos-buildpack.md)

[Exploration](https://github.com/paketo-community/explorations/pull/15)

## Checklist

### DIFF
--- a/text/0054-kerberos-buildpack.md
+++ b/text/0054-kerberos-buildpack.md
@@ -1,0 +1,70 @@
+# Introduce Kerberos Buildpack
+
+## Summary
+
+Introduce a buildpack that installs the [MIT
+Kerberos](https://kerberos.org/dist/) libraries and executables such that
+applications authenticating with Kerberos can do so on any of the supported
+stacks.
+
+## Motivation
+
+Kerberos is currently supported for applications built on top of the Full
+stack. This is enabled through the addition of the `krb5-user` package in our
+stack definition. Unfortunately, the Full stack also includes a number of other
+packages that many application developers would like to avoid installing.
+
+Creating a new Kerberos buildpack will allow these developers to make a more
+focused addition to their application image be ensuring that only those
+packages relating to Kerberos are installed. It will also allow the project to
+support Kerberos on top of stacks that are not the Full stack without needing
+to also include that package in these other stack definitions.
+
+## Detailed Explanation
+
+A new repository will be created at `paketo-buildpacks/kerberos`. This repo
+will house a buildpack that has the singular purpose of installing the Kerberos
+libraries and executables into a layer made available to other buildpacks or at
+runtime. The buildpack will include Kerberos as a precompiled package that can
+be dropped into a layer with support for Jammy only. The buildpack will follow
+the existing provide/requires conventions, exposing a dependency called
+`kerberos`.
+
+Note on Bionic: Lacking any direct feedback about a desired for this support on
+Bionic, we will explicitly ignore Bionic support for the initial implementation
+of this buildpack.
+
+## Rationale and Alternatives
+
+An alternative would be to add the `krb5-user` package to the Base stack
+images. Avoiding this would be preferrable as we should try not to increase the
+scope or vectors of attack for our most commonly used stack image.
+
+Another alternative would be to use the new Stack Extensions feature of the CNB
+spec to install the `krb5-user` package dynamically. Unfortunately, the
+implementation that has shipped at this point is insufficient to enable this
+outcome without significant maintenance burden on stack maintainers.
+Specifically, stack run images cannot be modified, only replaced entirely. This
+would mean that the stacks team would be required to produce images that are
+simply `<stack-variant> + krb5-user` for every stack we support going into the
+future. The CNB team is interested in solving the run image extension problem,
+but the timeline for a solution and its implementation is very vague at this
+point.
+
+## Implementation
+
+The implementation will use all of the new dependency management infrastructure
+and so be contained to this one repository. The dependency will need to be
+compiled against a Jammy stack and hosted by ourselves as no precompiled
+distribution has been found.
+
+## Prior Art
+
+The most immediately similar example of this kind of library-installing
+buildpack is the ICU buildpack that is used within the .NET Core language
+family.
+
+## Unresolved Questions and Bikeshedding
+
+* Should we explore using stack extensions to implement this?
+* What team should own this buildpack?


text/0054-kerberos-buildpack.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Prior Art
+
+The most immediately similar example of this kind of library-installing
+buildpack is the ICU buildpack that is used within the .NET Core language
+family.
+
+## Unresolved Questions and Bikeshedding
+
+* Do we need to backport support to Bionic?
```
sophiewigmore, line 54: I think it would be worth supporting Bionic for the next few months if its reasonably easy to compile the dependency against both stacks and get support. Are you aware of if different libraries will be needed for Bionic as well?

text/0054-kerberos-buildpack.md
```diff
@@ -0,0 +1,55 @@
...
+* Should we explore using stack extensions to implement this?
```
sophiewigmore, line 55: I am very interested in this, do you happen to have any insights into specific stack extension availability timelines? It looks like in the upstream the [issue to implement support](https://github.com/buildpacks/rfcs/issues/224) extensions is in progress. I wouldn't want to block this work on waiting for upstream completion, but I feel like this is a great candidate for that feature. I guess we can weigh whether its worth it to re-work this buildpack later with stack extensions.

text/0054-kerberos-buildpack.md
ryanmoran: I don't believe there are any major differences in compiling this for Bionic vs Jammy. Just that its a bunch of C code and so likely needs to be compiled independently on each version.

robdimsdale: I'd probably start with Jammy, and depending on how long it takes to release and how long we have left before Bionic goes out of support (i.e. April 2023), we could evaluate the benefit of shipping a bionic dependency too.

text/0054-kerberos-buildpack.md
dmikusa: Stack extensions was the first thing that came to mind when I read this RFC. I get that we may not want to wait for that, but I think we should consider it as we have to a.) make this buildpack and b.) continue to support it. 

If extensions are fully available in 3-6 months and can do all this, it puts us in a bad spot. We have a.) invested engineer time & resources into the buildpack and b.) have to either very quickly pivot and deprecate this buildpack or continue supporting something that has been superseded by upstream features.

dmikusa: I also wonder if it is sufficient to recommend apt-buildpack in the short term? Has that approach been tried? Even if it's not perfect, if it provides a workable solution now, then we could focus future efforts on using the stack extensions, which seems like the long term solution.

text/0054-kerberos-buildpack.md
fg-j: Worth noting that users relying on .NET 3.1 apps can't use jammy, because the runtime itself isn't compatible. So by not supporting bionic, we're saying that Kerberos and .NET 3.1 support are mutually exclusive features. This isn't terrible, since [.NET 3.1 technically EOLs December 13, 2022](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core). But I wouldn't be surprised if there are still .NET 3.1 users in the wild long after its EOL date.

text/0054-kerberos-buildpack.md
fg-j: > If extensions are fully available in 3-6 months and can do all this

To Ryan's point, it's also worth considering if lifecycle consumers (i.e. platform providers) will be ready to support this in 3-6 months. Does anyone have a window into how this is looking for `pack` or `kpack`?

text/0054-kerberos-buildpack.md
ryanmoran: But wouldn't we remove 3.1 after that date anyway? I feel like the overlap of a Kerberos buildpack existing and us still delivering support for .NET 3.1 would be infinitesimally small if non-existent, no?

fg-j: I think that's true. Just want to be clear about expectations-setting here for users.

davidmirror-ops:
WG 11-15-2022: [RM] do we want to use extensions? Not sure about when they will be available on CNB

text/0054-kerberos-buildpack.md
ryanmoran: I think we should ignore Bionic until we get feedback that users want it. I think that by the time we get a buildpack out, it will be close to the EOL date for Bionic anyway and any users trying this out will already be looking to run on Jammy. I'll update the RFC to state that we are explicitly not going to support Bionic in the absence of user demand.

text/0054-kerberos-buildpack.md
ryanmoran: After witnessing a demo of the current stack extension implementation during the CNB Working Group meeting on 11/17, I feel it is insufficient in its current state to fulfill the requirements of this RFC. Specifically, we cannot "extend" a run image by running extra commands against the base image. Instead, we must produce an entirely new run image that can be referenced with a `FROM` directive. This means we would need to have new run images for each stack that just include the `krb5-user` library. Extrapolate this out for any other stack we might need to support in the future and the added maintenance burden becomes clear.

I will update the RFC with this context.

sophiewigmore:
Due to the lengthy timelines we discussed at WG around being able to use extensions, I think we should proceed with this proposal so we can support this use case sooner rather than later.

sophiewigmore (approved):

text/0054-kerberos-buildpack.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Detailed Explanation
+
+A new repository will be created at `paketo-buildpacks/kerberos`. This repo
+will house a buildpack that has the singular purpose of installing the Kerberos
+libraries and executables into a layer made available to other buildpacks or at
+runtime. The buildpack will include Kerberos as a precompiled package that can
+be dropped into a layer with support for Jammy only. The buildpack will follow
+the existing provide/requires conventions, exposing a dependency called
+`kerberos`.
```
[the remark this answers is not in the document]
ryanmoran: I guess I wasn't just thinking a `dpkg --install` to some layer. I was expecting that we compile this from source, hopefully in a way that enables it to be run on more than just Ubuntu stacks. Additionally, I'm eventually gonna want to write an RFC that suggests we include this in some language family buildpacks. Are you suggesting that we pull buildpacks we don't maintain into those metabuildpacks?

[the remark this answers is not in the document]
dmikusa: I think my thought is that compiling and distributing software is a lot of work, and there are entire teams of people already doing that. If we can skip that part and use what's already being offered, it gives us more time to focus on buildpacks.

My thought would be to perhaps make an official apt-buildpack under Paketo. The OSS one is getting old, and I'm not sure it's being updated (I don't think it works with Jammy). Possibly we fork from that? At any rate, we can jump to the point of being able to install the Kerberos software in the Paketo stacks much faster, IMO. Not on point for this RFC, but it also has a broader use beyond just this particular package, as you can install many other packages this way too.

As for other stacks, if we have to add an rpm-buildpack (or another package format) to support other stacks, then we could explore that. 

My only hesitation would be if there are known limitations with the existing packages, such that they won't work when installed in non-standard locations. Has anyone tried using the existing apt-buildpack and the required packages? I'm wondering if it works? or if there are known problems with that approach?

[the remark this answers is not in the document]
ryanmoran: I guess I worry that this expands the scope of what I really view as a stop-gap to address a need in the short-term. If we're talking about being able to install all sort of OS-level packages, I'd really like to see that in a stack extension, and I'd be happy to wait until we have more fleshed out support in CNB to make that happen. But, for now, I'd either like to see us implement something more limited or we can choose to do nothing and I can explore enabling these types of applications outside of Paketo.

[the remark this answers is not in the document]
dmikusa: That's what worries me about this RFC, the stop-gap nature of this. Building, maintaining, and distributing the software isn't trivial, it also includes a buildpack to build, maintain and release. After upstream releases stack extensions we're not going to want to keep using it, but you can't release something and then deprecate it in 6 months or really even a year. Users will expect a longer-term availability, which puts the project in a bad position. Either we spend resources to keep maintaining it, or we deprecate it in short order.

If there is intended to be just a stop-gap thing, I'd rather see it go into paketo-community. That will temper user any expectations.

My thought with the apt-buildpack (and similar) is that a.) we don't have to build/maintain the software and b.) it might have some usable life even after stack extensions are available. The stack extensions RFC is not a perfect solution, it'll have advantages and disadvantages. Perhaps an apt-buildpack can fill in some of those disadvantages, but even if it doesn't the ongoing maintenance of such a buildpack is much less because we don't have to maintain the software, just the buildpack itself.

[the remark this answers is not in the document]
ryanmoran: Ok. I think this is leading towards me finding another route to get this outcome. I'm not gonna go the `apt` buildpack route. I just don't have any interest in opening up and maintaining that rabbit hole of complexity.

PR #260, opened 2022-11-02
Adds Squash Stacks RFC

ryanmoran:
## Summary

[Readable](https://github.com/paketo-buildpacks/rfcs/blob/squash-stacks/text/stacks/0007-squash-stack.md)

## Checklist

### DIFF
--- a/text/stacks/0007-squash-stack.md
+++ b/text/stacks/0007-squash-stack.md
@@ -1,0 +1,46 @@
+# Squash Stack Images
+
+## Summary
+
+The build and run images delivered as a stack for our users should each be
+squashed into a single layer to reduce the total number of layers in the
+builder and application images and remove duplicated files that appear in
+multiple layers.
+
+## Motivation
+
+Stack images act as the base image for applications and our own builder images.
+Their internal layer implementations should not be important to end users.
+Recently, we have been fielding a lot of bug reports wherein users cannot
+create their application images because the total number of layers for the
+image has exceeded the limit enforced by `containerd`.
+
+Squashing each stack image down into a single layer will create more headroom
+for both the builders and the images they create.
+
+## Detailed Explanation
+
+The `jam create-stack` command will include a new flag, `--squash`, that will
+squash the build and run image layers each into a single layer. These squashed
+images will be what the project ships as the canonical stack images for Bionic
+and Jammy.
+
+## Rationale and Alternatives
+
+As an alternative, we could choose to add this functionality to `jam
+create-stack`, but not use it as the default for the stack we ship, but allow
+those with layer limit issues to squash the stacks themselves.
+
+## Implementation
+
+The [`crane`
+CLI](https://github.com/google/go-containerregistry/tree/main/cmd/crane) has a
+`flatten` command that does [this exact
+operation](https://github.com/google/go-containerregistry/blob/1711cefd7eec057d3892d0bbce1bcd3f8c46d606/cmd/crane/cmd/flatten.go#L201-L254),
+but against an image hosted already on a registry. We can reference this
+existing code as a starting point for our own implementation that would operate
+against an image as implemented within `jam create-stack`.
+
+## Prior Art
+
+* [`crane flatten`](https://github.com/google/go-containerregistry/blob/main/cmd/crane/doc/crane_flatten.md)
--- a/text/stacks/README.md
+++ b/text/stacks/README.md
@@ -6,7 +6,7 @@ This directory contains RFCs that pertain to the [Stacks Subteam](https://github
 
 * [0004: Stacks based on Ubuntu 2022.04: Jammy Jellyfish](0004-jammy-jellyfish.md)
 * [0006: Add to Tiny Run Image to Enable Node Apps](0006-node-on-tiny.md)
-
+* [0007: Squash Stack Images](0007-squash-stack.md)
 
 ## Implemented RFCs
 


text/stacks/0007-squash-stack.md
```diff
@@ -0,0 +1,45 @@
+# Squash Stack Images
+
+## Summary
```
robdimsdale, line 3: Are there any concerns with the fact that a single layer will effectively disable any caching? I'm assuming not, given that the bulk of the stacks (i.e. the package installation) is currently all in a single layer and the layers above are minimal in size.

text/stacks/0007-squash-stack.md
```diff
@@ -0,0 +1,45 @@
...
+for both the builders and the images they create.
+
+## Detailed Explanation
+
+The `jam create-stack` command will include a new flag, `--squash`, that will
+squash the build and run image layers into a single layer. These squashed
+images will be what the project ships as the canonical stack images for Bionic
+and Jammy.
+
+## Rationale and Alternatives
```
robdimsdale, line 27: Instead of `jam create-stack --squash` we could modify the pipelines to run `crane flatten`.

This has this advantage of not requiring us to maintain more code in `jam create-stack`.

text/stacks/0007-squash-stack.md
```diff
@@ -0,0 +1,45 @@
...
+## Motivation
+
+Stack images act as the base image for applications and our own builder images.
+Their internal layer implementations should not be important to end users.
+Recently, we have been fielding a lot of bug reports wherein users cannot
+create their application images because the total number of layers for the
+image has exceeded the limit enforced by `containerd`.
+
+Squashing the stack images down into a single layer will create more headroom
+for both the builders and the images they create.
```
sophiewigmore, line 18: What do you envision the `builder.toml` to look like when using a single layer? Will the separate `build-image` and `run-image` lines go away in favour of a `stack-image` line? Have you tried this out in a builder to see if it works as expected?

text/stacks/0007-squash-stack.md
ryanmoran: I think the main issue here is that you need to run this against an image that is already in a registry. So, we'd need to push to some registry, flatten, pull, and then push everywhere else? And then secondarily, nobody else in the community would really benefit from this. Is that fine?

text/stacks/0007-squash-stack.md
ryanmoran: Sorry. This is just clumsily worded. There are still 2 images. Its just that each image only has 1 layer.

text/stacks/0007-squash-stack.md
ryanmoran: I think users already experience little in the way of caching since the package layer is just one big layer and we pretty much change it on every release. Technically, yes, this makes caching the stack worse. But in reality, I believe that most users benefited very little from that caching to begin with.

text/stacks/0007-squash-stack.md
sophiewigmore: Oh, ok, yea that didn't come through. I was a confused about how that would work

ryanmoran: I think I should say "Squashing each stack image down into a single layer..."

Is that clearer?

text/stacks/0007-squash-stack.md
robdimsdale: Oh, ok. I didn't dig into the mechanism of `crane flatten`. We could explore pushing the image to a local registry, flattening, then pushing to dockerhub/gcr. But I agree that this path is sounding less attractive than adding support in `jam create-stack`.

text/stacks/0007-squash-stack.md
robdimsdale: Yeah, that's pretty much what I thought, too. I just wanted to be explicit about that consequences of this decision; why/how we are ok with the lack of caching.

text/stacks/0007-squash-stack.md
sophiewigmore: Definitely!

dmikusa:
💯 This will be be fantastic!

dmikusa (approved):

fg-j:
I was curious how many layers this would save (i.e. how many layers are in each current stack) - it seems like each stack currently contains ~7 layers. So, while it won't permanently solve our problems regarding builders with too many layers, it'll create some wiggle room for people creating custom builders that add to the Paketo Full builders.

dmikusa:
> I was curious how many layers this would save (i.e. how many layers are in each current stack) - it seems like each stack currently contains ~7 layers. So, while it won't permanently solve our problems regarding builders with too many layers, it'll create some wiggle room for people creating custom builders that add to the Paketo Full builders.

Hmm, interesting. I think I misunderstood this RFC. Is this do just squash the run/build image layers? or is it to squash down the builder layers? or both?

I agree, squashing the run/build images isn't going to save a ton, but squashing the builder layers, that would solve the layer problem for us.

fg-j:
Perhaps a discussion of squashing builders should be added here. Right now, we create builders with `pack builder create`, which doesn't have a `--squash` option. A potentially more reusable implementation here is to add a `jam squash-image` command that flattens an arbitrary image. This could be used for both builders and stacks.

fg-j:
Something to bear in mind with potentially squashing builders: If builders squash all buildpacks into a single layer, whenever any buildpack in the builder changes, the entire cached layer will be invalidated. This will significantly hurt builder-pull performance on any platform that uses on OCI layer caching. Maybe it would be more prudent to squash composite buildpacks before distributing them in builders.

dmikusa:
> Something to bear in mind with potentially squashing builders: If builders squash all buildpacks into a single layer, whenever any buildpack in the builder changes, the entire cached layer will be invalidated. This will significantly hurt builder-pull performance on any platform that uses on OCI layer caching. Maybe it would be more prudent to squash composite buildpacks before distributing them in builders.

That's a fair point. I think it would be interesting to try both and see how it goes. I'd be curious to see how big the layer ends up being with all the buildpacks squashed into one layer. There are quite a few individual layers now with the builder image and that's not exactly speedy. Sometimes downloading one larger resource is faster than downloading many smaller resources.

fg-j:
All in all, if this RFC introduces a more general `jam squash-image` subcommand and sets the intent to (at least) squash stacks, we can experiment and decide on a best builder-squashing approach after implementation. I'm all for experimenting on this.

ryanmoran:
I don't think we'd likely want a builder squashed into a single layer. More likely, we'd want to keep the base stack layer, and then squash each of the buildpack groups into a layer. It feels like it'll be harder to do that from the outside.

davidmirror-ops:
WG 11-15-2022: [RM] think about a more intelligent way to squash layers
[FGJ] What about generalizing this, and implementing it as a CLI command?
[RM] API design would be needed

fg-j (approved):

ryanmoran:
@paketo-buildpacks/stacks-maintainers I'm looking for any more feedback on this.

sophiewigmore:
I'm fine with proceeding with this as-is. We can extend support further for builders, etc in the future if we want to.

robdimsdale (approved):
Sorry, I thought I already approved this :/

PR #264, opened 2022-11-17
Make nodejs a paketo-hosted dependency

TisVictress:


text/nodejs/0015-dependencies.md
```diff
@@ -11,13 +11,13 @@ The following dependencies should be removed as Paketo-hosted dependencies:
...
+Keep this as a Paketo-hosted dependency.
 
 The Paketo node-engine buildpack supports using the system's CA store, which is not a default feature in node.
 This feature is currently enabled by compiling node from source with the `--openssl-use-def-ca-store` flag.
-However, this could also be achieved by simply setting the `NODE_OPTIONS` environment variable to `--use-openssl-ca`, which
-may make it easier to use this dependency directly from upstream.
-Therefore, this RFC proposes removing node as a Paketo-hosted dependency.
+But it can also be enabled by setting `NODE_OPTIONS` to `--use-openssl-ca`.
+While older versions of node can be installed directly from upstream, versions 18 and up are not compatible with bionic.
+Therefore, this RFC does not propose removing node as a Paketo-hosted dependency.
```
ryanmoran, line 20: I'm confused about what we're gonna be doing. Are we saying that because Node.js v18+ doesn't work with Bionic that we need to now compile and host our own Node.js dependencies for everything? I don't understand how one connects to the other. Can you provide some more context here on why you think this is the right path forward?

TisVictress: @ryanmoran  No worries. Node can be installed directly from upstream on Bionic except for versions 18+ because they are not compatible. So trying to install v19.1.0 from upstream fails with an error: `node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by node)`.  This doesn't happen on Jammy but it might be less confusing if both dependencies were compiled as opposed to compiling for bionic for versions 18+, pulling from upstream for versions below 18, and pulling from upstream for jammy. So we'd be consistent in the way we update dependencies for node.

robdimsdale: I thought I commented on this, but I can't see it anymore, so I'll write it again.

Another option is to say that we don't support nodejs 18+ on Ubuntu 18.04 (Bionic). If this isn't support out of the box upstream it seems reasonable to say we won't support it either.

Then we can stick to just using upstream dependencies everywhere - up to 18 for Bionic, and all versions that run on jammy (16+? 14+?) for Jammy.

According to https://endoflife.date/nodejs, support for Node v14 ends in April 2023 - around the time Bionic goes out of support. So we will continue to have two minor lines of nodejs support on Bionic, and (presumably) at least two minor lines of nodejs supported on Jammy.

ryanmoran: Just for clarity, v18 does not work on Bionic when we pull from upstream?

TisVictress: @ryanmoran That's correct. Sorry for the confusion. For v18, pulling directly from upstream causes this error:
`node: /lib/x86_64-linux-gnu/libc.so.6: version GLIBC_2.28 not found (required by node)` on bionic.

ryanmoran: Hmm. This complicates things as we just switched to using v18 as the default version. I wonder if we could have compiled dependencies for Bionic, and use upstream for Jammy. This would be painful short-term, but would put us in a better spot once Bionic expires. Thoughts?

ryanmoran (approved):

PR #265, opened 2022-11-18
Adds RFC to remove Ruby from the Jammy Full stack

ryanmoran:
## Summary

[Readable](https://github.com/paketo-buildpacks/rfcs/blob/full-stack-remove-ruby/text/stacks/0007-jammy-full-remove-ruby.md)

## Checklist

### DIFF
--- a/text/stacks/0008-jammy-full-remove-ruby.md
+++ b/text/stacks/0008-jammy-full-remove-ruby.md
@@ -1,0 +1,45 @@
+# Remove Ruby from the Jammy Full Stack
+
+## Summary
+
+The Jammy Full stack should not install the `ruby` package.
+
+## Motivation
+
+The `ruby` (Ruby 3.0) package will reach the end of its support period (March
+2024) before Jammy does (April 2027). This means that for a large portion of
+the support period for the Jammy Full stack, it will contain an unsupported
+Ruby interpreter.
+
+Additionally, removing the `ruby` package will have little impact on existing
+Jammy Full stack users since they will be able to still get a Ruby interpreter
+through the `mri` buildpack or by employing a stack extension to install the
+`ruby` package.
+
+Finally, the existence of the `ruby` package in the stack is a relic of its
+ancestry having been developed from the Cloud Foundry `cflinuxfs*` line of stack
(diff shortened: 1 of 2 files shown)


sophiewigmore:
Based off of this RFC, removing Ruby will have no impact on any of _our_ buildpacks right? If that's the case I'm supportive of this.
 t will be unfortunate if this change breaks any stacks consumers who have their own buildpacks written in Ruby, but A) I'm not aware of any people doing this at the moment and B) I think the workarounds of using a custom stack, MRI buildpack, or extensions in the future are all reasonable.

sophiewigmore:
I wonder if there are any other legacy-type packages in the stack that fall under the same category?

robdimsdale (approved):

robdimsdale:
@sophiewigmore I think we'd probably want to remove `python` and a few other libraries too (`ubuntu-minimal`?).

I'd be happy with one RFC per library, or one RFC for all. I don't mind either way.

sophiewigmore:
Notes from working group 11/29:
- We include Ruby and Python as a legacy thing, buildpacks in Cloud Foundry were written in those languages once upon a time
- This is no longer the case for Paketo buildpacks, its best to remove it
- Bump the major version of the stack accordingly
- There are enough mitigations to get around this issue.
- Removing Python is a longer-term goal, it has more related dependencies we'll need to investigate

- Question - the stacks aren't yet at a 1.0.0 version - what version bump should this change be?

robdimsdale:
As far as the version, I'd like to remove all the packages we think we want to before making a `v1.0.0` release. Otherwise we will make a series of breaking changes (and major version bumps) as we identify more packages.

I don't want to go too far the other way - deferring v1.0.0 indefinitely - either, but I think I'd like to keep pre-v1.0.0 once Ruby is removed.

sophiewigmore:
I am fine with this plan, and we can address the more complex Python-related concerns in a separate RFC.

sophiewigmore (approved):

robdimsdale:
@ryanmoran I think we have 👍  from all the stacks maintainers, so feel free to merge whenever you want.

PR #267, opened 2022-12-12
Propose additional JVMs in Java buildpack

kevin-ortega:
[Readable](https://github.com/kevin-ortega/rfcs/blob/use-alternate-jvm-with-environment-variable/text/java/0000-alternate-jvms-in-java-buildpack.md)

### DIFF
--- a/text/java/0000-alternate-jvms-in-java-buildpack.md
+++ b/text/java/0000-alternate-jvms-in-java-buildpack.md
@@ -46,6 +46,8 @@ In a composite buildpack all the JVM buildpacks will detect true to indicate the
 
 Additional JVMs can easily be added to the Java buildpack by implementing this RFC.  
 
+Initially, this RFC should be implemented in the new language family builder.  If and when the layer limit issue is resolved more JVM providers can be added to the main Paketo builders.    
+
 ## Prior Art
 
 - Offering alternate Java application servers in the composite java buildpack was implemented using an environment variable.  
--- a/text/java/0013-alternate-jvms-in-java-buildpack.md
+++ b/text/java/0013-alternate-jvms-in-java-buildpack.md
@@ -29,11 +29,7 @@ Alternatives:
 
 ## Implementation
 
-In a composite buildpack all the JVM buildpacks will detect true to indicate they can provide a JRE/JDK.  At build time, 
-buildpacks execute in order, according to the order group.  The first JVM buildpack in the order group, bellsoft-liberica, will
-have the first opportunity to contribute the JVM regardless if `BP_JVM_VENDOR` is set.  If not set, the buildpack will be free
-to contribute.   If set, it should only contribute if `BP_JVM_VENDOR` is set to `liberica`.  Other buildpacks are given the 
-opportunity to provide the JVM and should only if the buildplan entry is not already met and the JVM name matches `BP_JVM_VENDOR`.  Otherwise, the buildpack will exit successfully without doing anything.  
+In a composite buildpack all the JVM buildpacks will detect true to indicate they can provide a JRE/JDK.  At build time, all JVM buildpacks are free to contribute if the buildplan entry is not met and `BP_JVM_VENDOR` is set to their tag or not set at all. Buildpacks execute in the order they are defined. The bellsoft-liberica buildpack comes first, which preserves the current default behaviour.
 
 | JVM                | tag          	| Buildpack                           |
 | ------------       | ------------ 	| ------------------------------------|
--- a/text/java/0013-alternate-jvms-in-java-buildpack.md
+++ b/text/java/0013-alternate-jvms-in-java-buildpack.md
@@ -29,7 +29,7 @@ Alternatives:
 
 ## Implementation
 
-In addition to the current criteria for each buildpack, the detect method will also validate itself against the value of `BP_JVM_VENDOR` and contribute only if it matches. BellSoft Liberica will remain the default JVM.
+In addition to the current criteria for each buildpack, the detect method will also take the value of `BP_JVM_VENDOR` into account. If it is set, a buildpack should only detect if the value matches. BellSoft Liberica will remain the default JVM.
 
 | JVM                | tag          	| Buildpack                           |
 | ------------       | ------------ 	| ------------------------------------|
--- a/text/java/0013-alternate-jvms-in-java-buildpack.md
+++ b/text/java/0013-alternate-jvms-in-java-buildpack.md
@@ -1,0 +1,55 @@
+# Add additional JVMs to the Java buildpack
+
+## Summary
+
+The default JVM in the Java buildpack is the Liberica JVM.  To use an alternative JVM requires specifying two --buildpack arguments on the `pack build` command.  
+
+This RFC proposes a simpler method by defining an environment variable to specify which JVM provider to use.  This would be similar to how the Java Buildpack uses `BP_JAVA_APP_SERVER` to specify an alternate java application server to use instead of the default, apache tomcat.  
+
+## Motivation
+
+There are a few motivations for this change:
+
+1. Using an alternate JVM is tedious.  
+2. The preferred JVM for the Liberty application server is eclipse openj9 because there are features in it that the liberty server takes advantage of we would like to exploit in the liberty buildpack.
+3. The alternate JVM buildpack runs before the CA certs buildpack and therefore traffic from the alternate JVM vendor buildpack won’t trust any additional CA certs.  
+4. Easier adoption of the other JVM buildpacks.
+
+## Detailed Explanation
+
+Currently, the Java buildpack only contributes the BellSoft Liberica JVM.  If we can accommodate multiple JVMs then we can provide a more
+well-rounded java buildpack, reach more Java communities, generate more interest in Paketo and Cloud Native buildpacks, and ultimately make it easier for users to build images for Java applications.
+
+## Rationale and Alternatives
+
+Alternatives:
+
+- Do nothing. Users of the Java buildpack will have to manually create a builder or manually specify the buildpack order, which is tedious.
+- Create new composite buildpacks other JVMs as the default.
+
+## Implementation
+
+In addition to the current criteria for each buildpack, the detect method will also validate itself against the value of `BP_JAVA_JVM` and contribute only if it matches. BellSoft Liberica will remain the default JVM.
+
+| JVM                | tag          	| Buildpack                           |
+| ------------       | ------------ 	| ------------------------------------|
+| BellSoft Liberica  | liberica       |  bellsoft-liberica  buildpack       |
+| Eclipse Openj9     | openj9         |  eclipse-openj9 buildpack           |
+| Microsoft OpenJDK  | ms-openjdk     |  microsoft-openjdk buildpack        |
+| Oracle             | oracle         |  oracle buildpack                   |
+| Amazon Corretto    | corretto       |  amazon-corretto                    |
+| Adoptium           | adoptium       |  adoptium buildpack                 |
+| SAP Machine        | sapmachine     |  sap-machine buildpack              |
+| Alibaba Dragonwell | dragonwell     |  alibaba-dragonwell buildpack       |
+| GraalVM            | graalvm        |  graalvm buildpack                  |
+| Azul Zulu          | azul-zulu      |  azul-zulu buildpack                |
+
+Additional JVMs can easily be added to the Java buildpack by implementing this RFC.  
+
+## Prior Art
+
+- Offering alternate Java application servers in the composite java buildpack was implemented using an environment variable.  
+
+## Unresolved Questions and Bikeshedding
+
+- None
--- a/text/java/0013-alternate-jvms-in-java-buildpack.md
+++ b/text/java/0013-alternate-jvms-in-java-buildpack.md
@@ -29,7 +29,7 @@ Alternatives:
 
 ## Implementation
 
-In addition to the current criteria for each buildpack, the detect method will also validate itself against the value of `BP_JAVA_JVM` and contribute only if it matches. BellSoft Liberica will remain the default JVM.
+In addition to the current criteria for each buildpack, the detect method will also validate itself against the value of `BP_JVM_VENDOR` and contribute only if it matches. BellSoft Liberica will remain the default JVM.
 
 | JVM                | tag          	| Buildpack                           |
 | ------------       | ------------ 	| ------------------------------------|
--- a/text/java/0013-alternate-jvms-in-java-buildpack.md
+++ b/text/java/0013-alternate-jvms-in-java-buildpack.md
@@ -29,7 +29,11 @@ Alternatives:
 
 ## Implementation
 
-In addition to the current criteria for each buildpack, the detect method will also take the value of `BP_JVM_VENDOR` into account. If it is set, a buildpack should only detect if the value matches. BellSoft Liberica will remain the default JVM.
+In a composite buildpack all the JVM buildpacks will detect true to indicate they can provide a JRE/JDK.  At build time, 
+buildpacks execute in order, according to the order group.  The first JVM buildpack in the order group, bellsoft-liberica, will
+have the first opportunity to contribute the JVM regardless if `BP_JVM_VENDOR` is set.  If not set, the buildpack will be free
+to contribute.   If set, it should only contribute if `BP_JVM_VENDOR` is set to `liberica`.  Other buildpacks are given the 
+opportunity to provide the JVM and should only if the buildplan entry is not already met and the JVM name matches `BP_JVM_VENDOR`.  Otherwise, the buildpack will exit successfully without doing anything.  
 
 | JVM                | tag          	| Buildpack                           |
 | ------------       | ------------ 	| ------------------------------------|


text/java/0013-alternate-jvms-in-java-buildpack.md
```diff
@@ -0,0 +1,55 @@
...
+
+Additional JVMs can easily be added to the Java buildpack by implementing this RFC.  
+
+## Prior Art
+
+- Offering alternate Java application servers in the composite java buildpack was implemented using an environment variable.  
+
+## Unresolved Questions and Bikeshedding
+
+- None
```
dmikusa, line 55: So there is one technical hurdle here. Right now, the Paketo Builders (especially the full builder) are very close to the max layer limit for container images. This has temporarily put a hold on adding new buildpacks because each new buildpack pushes the layer limit up more.

There has been a proposal to squash down some of the layers, but I'm not sure where we are in terms of implementing that and how much of a reduction in the number of layers in the builders it would actually offer.

tl;dr - Adding ten more buildpacks isn't going to be feasible at this point.

The question I would propose is, do we wait for this to be resolved in the Paketo builders or do we strike out and make a Java-specific builder (this would require its own RFC also)? The Java-specific builder would have fewer layers as we'd remove the other language families.

Thoughts?

text/java/0013-alternate-jvms-in-java-buildpack.md
```diff
@@ -0,0 +1,55 @@
...
+## Rationale and Alternatives
+
+Alternatives:
+
+- Do nothing. Users of the Java buildpack will have to manually create a builder or manually specify the buildpack order, which is tedious.
+- Create new composite buildpacks other JVMs as the default.
+
+## Implementation
+
+In addition to the current criteria for each buildpack, the detect method will also validate itself against the value of `BP_JAVA_JVM` and contribute only if it matches. BellSoft Liberica will remain the default JVM.
```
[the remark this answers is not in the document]
kevin-ortega: I just threw a name out there without giving it much thought.  I'm don't have any attachment to any name.    

`BP_JAVA_VENDOR` makes sense.

text/java/0013-alternate-jvms-in-java-buildpack.md
kevin-ortega: Yes, I figured the technical hurdle would apply here but I at least wanted to get this discussion started. 

I do like the idea of a java-specific builder.  We don't have to add all the JVMs.  Do we know which are the most popular?

loewenstein:
Just for clarity, this will mainly affect the java meta buildpack (adding the alternative JVMs), right? Are jvm buildpacks free to detect if the environment variable is not set? I am particularly interested in having the sap-machine buildpack detecting positive because we package that and a proprietary jbm buildpack to our Java meta-buildpack and rely on the requested Java version to distinguish both (the proprietary is 8 only and sap-machine is 11+ only).

text/java/0013-alternate-jvms-in-java-buildpack.md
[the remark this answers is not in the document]
pivotal-david-osullivan: My 2c here would be `BP_JVM_VENDOR` as this slots in with the existing `BP_JVM_TYPE` and `BP_JVM_VERSION`

kevin-ortega:
> Just for clarity, this will mainly affect the java meta buildpack (adding the alternative JVMs), right? Are jvm buildpacks free to detect if the environment variable is not set? I am particularly interested in having the sap-machine buildpack detecting positive because we package that and a proprietary jbm buildpack to our Java meta-buildpack and rely on the requested Java version to distinguish both (the proprietary is 8 only and sap-machine is 11+ only).
> 
> Edit: I guess it is not "for clarity" only after all. The more I think about it, the more I think this RFC will break our current setup.

Yes, JVM buildpacks are free to detect if the variable is not set.   When we added additional Java app servers to the Java meta buildpack we allowed for the buildpacks to detect if the variable is not set to cover the scenario with a custom builder with only one java app server buildpack.  I would assume we'd do the same thing for JVM vendors.

text/java/0013-alternate-jvms-in-java-buildpack.md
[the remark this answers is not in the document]
kevin-ortega: I will change the name to `BP_JVM_VENDOR`.

davidmirror-ops:
2022-12-13 WG Notes: waiting for feedback and rackle technical challenges associated with it. Related to a topic for 2023 Roadmap posted by Dan Mikusa

text/java/0013-alternate-jvms-in-java-buildpack.md
```diff
@@ -0,0 +1,55 @@
...
+## Rationale and Alternatives
+
+Alternatives:
+
+- Do nothing. Users of the Java buildpack will have to manually create a builder or manually specify the buildpack order, which is tedious.
+- Create new composite buildpacks other JVMs as the default.
+
+## Implementation
+
+In addition to the current criteria for each buildpack, the detect method will also validate itself against the value of `BP_JVM_VENDOR` and contribute only if it matches. BellSoft Liberica will remain the default JVM.
```
loewenstein, line 32:
```suggestion
In addition to the current criteria for each buildpack, the detect method will also take the value of `BP_JVM_VENDOR` into account. If it is set, a buildpack should only detect if the value matches. BellSoft Liberica will remain the default JVM.
```

This would make it more clear, that a JVM providing buildpack should continue to detect, when all other criteria are being fulfilled, if there is no value provided for `BP_JVM_VENDOR`.
Although, this leads to the question of who will be providing `BP_JVM_VENDOR=liberica` to handle
> BellSoft Liberica will remain the default JVM.

loewenstein:
> > Just for clarity, this will mainly affect the java meta buildpack (adding the alternative JVMs), right? Are jvm buildpacks free to detect if the environment variable is not set? I am particularly interested in having the sap-machine buildpack detecting positive because we package that and a proprietary jbm buildpack to our Java meta-buildpack and rely on the requested Java version to distinguish both (the proprietary is 8 only and sap-machine is 11+ only).
> > Edit: I guess it is not "for clarity" only after all. The more I think about it, the more I think this RFC will break our current setup.
> 
> Yes, JVM buildpacks are free to detect if the variable is not set. When we added additional Java app servers to the Java meta buildpack we allowed for the buildpacks to detect if the variable is not set to cover the scenario with a custom builder with only one java app server buildpack. I would assume we'd do the same thing for JVM vendors.

I added a suggestion to make this more clear: https://github.com/paketo-buildpacks/rfcs/pull/267#discussion_r1047631858

text/java/0013-alternate-jvms-in-java-buildpack.md
dmikusa: The way we did this with `BP_JAVA_APP_SERVER` is that it's based on buildpack order. 

>* `$BP_JAVA_APP_SERVER` is `tomcat` or if `$BP_JAVA_APP_SERVER` is unset or empty and this is the first buildpack to provide a Java application server.

The way this works is that at build time, the buildpack checks to see if some other buildpack has already met that buildplan entry. If it has not been met, the buildpack is free to satisfy it.

loewenstein: I am not entirely sure I get this. The order is static, how are the alternative buildpacks meant to come into effect?

Anyway, your explanation does sound like it would still satisfy our requirement. As long as sap-machine keeps providing it's jvm dependencies even without `BP_JVM_VENDOR` set to `sap-machine`, we should be fine.

dmikusa: @loewenstein It works off the buildpack order defined in the composite buildpack.

So a proposed new order could be something like:

- bellsoft-liberica
- JVM vendor A
- JVM vendor B
- JVM vendor ...
- JVM vendor N
- <existing buildpack order starts here>

At detect, all of the buildpacks run. The JVM vendor providing buildpacks say "yes, I can provide you with a JVM". They all pass detection. The build process runs in sequence based on the order of buildpacks defined in the order group and that passed detection. Because all of the JVM vendor buildpacks pass by default without any user intervention, Bellsoft Liberica would run first and contribute the JVM. Then JVM Vendor A would run, it would see that the buildplan entry for a JDK/JRE has been satisfied and not do anything (just exit successfully). Then JVM Vendor B would do the same, and C and D, and so on. 

It's not a lot different than how things work now, if you just load multiple JVM vendor buildpacks into an order group. The main difference is that they would all ship this way by default, so a user wouldn't need to add `-b` flags to `pack build` and insert the JVM vendor buildpacks in the proper order.

Now, the other main difference is that a user can change this order by setting `BP_JVM_VENDOR=sap-machine` (or their vendor of choice. This works similarly. Detect runs in parallel, all the JVM vendor buildpack pass. The difference is at build time. The Bellsoft Liberica buildpack would run, it would see `BP_JVM_VENDOR=sap-machine` and opt-out, more specifically it'll mark the JRE/JDK buildplan entries as Unmet. This is a special condition that tells the lifecycle to essentially try another buildpack. The next JVM Vendor buildpack would run, look at `BP_JVM_VENDOR=sap-machine`, and so on until the SAP Machine buildpack runs. It would see `BP_JVM_VENDOR=sap-machine`, say that's me and install it's JRE/JDK. If there are any JVM vendor buildpacks left after it, they would then run and no-op (like in the default case).

So out of the box with this proposal, the paketo-buildpacks/java buildpack will ship with Bellsoft Liberica as the default, and it will include all of the other JVM vendors so a user can pick with `BP_JVM_VENDOR`. With the proposal as it stands, there is no way to change the default aside from changing the buildpack order which requires a custom composite buildpack (or repackage of paketo-buildpacks/java) or a custom builder that redefines the order groups (you don't need a composite buildpack, you could define the order groups at the builder level).

Having said that, there is an upstream buildpacks RFC that will allow for [builders to set environment variables](https://github.com/buildpacks/rfcs/pull/230). That has been approved but not implemented. When that is implemented, we would be able to consider adding some additional logic, perhaps a `BP_DEFAULT_JVM_VENDOR` (and similar for app servers) environment variable that would be intended to be set at the builder level. This could then further influence the logic, so rather than the default being the first, builder authors could more easily set a default. That's down the road in a different Paketo RFC though.

loewenstein: Alright, then it's about `build` not `detect`. But the same logic would still work, wouldn't it?
I.e. 
Liberica is first in the order, so while all jvm buildpacks detect, the build will skip early, because the build plan entry is already satisfied.
Looked at individually, the jvm buildpacks can all install as long as `BP_JVM_VENDOR` is either not set at all or set to their marker, e.g. `sap-machine`.
Our composite buildpack can keep packaging too different jvm buildpacks (sap-machine and a proprietary one) with non-overlapping versions and our users don't need to set `BP_JVM_VENDOR` at all.

Sound correct?

kevin-ortega: That sounds correct and I will update the RFC to reflect the logic is in build, not detect.

kevin-ortega: @loewenstein I have updated the RFC providing more details on how this is implemented.

text/java/0013-alternate-jvms-in-java-buildpack.md
```diff
@@ -0,0 +1,59 @@
...
+- Do nothing. Users of the Java buildpack will have to manually create a builder or manually specify the buildpack order, which is tedious.
+- Create new composite buildpacks other JVMs as the default.
+
+## Implementation
+
+In a composite buildpack all the JVM buildpacks will detect true to indicate they can provide a JRE/JDK.  At build time, 
+buildpacks execute in order, according to the order group.  The first JVM buildpack in the order group, bellsoft-liberica, will
+have the first opportunity to contribute the JVM regardless if `BP_JVM_VENDOR` is set.  If not set, the buildpack will be free
+to contribute.   If set, it should only contribute if `BP_JVM_VENDOR` is set to `liberica`.  Other buildpacks are given the 
+opportunity to provide the JVM and should only if the buildplan entry is not already met and the JVM name matches `BP_JVM_VENDOR`.  Otherwise, the buildpack will exit successfully without doing anything.
```
loewenstein, line 36:
```suggestion
In a composite buildpack all the JVM buildpacks will detect true to indicate they can provide a JRE/JDK.  At build time, all JVM buildpacks are free to contribute if the buildplan entry is not met and `BP_JVM_VENDOR` is set to their tag or not set at all. Buildpacks execute in the order they are defined. The bellsoft-liberica buildpack comes first, which guarantees the current default behaviour.
```

I changed more than I intended, sorry. The main point is that for all buildpacks `BP_JVM_VENDOR` should either match their tag or not be set at all.

kevin-ortega: @loewenstein Thanks for the suggestions.

mstoodle:
For clarity, I would adjust your table of "JVM"s because it's really a table of "JDK"s. You can highlight which JVM is used (which highlights one reason why IBM Semeru Runtimes really should be in the list as the only OpenJ9 representative).

Vendor | JDK | JVM | tag | Buildpack
-- | -- | --
BellSoft | Liberica | Hotspot | liberica | bellsoft-liberica  buildpack
IBM | Semeru Runtimes Open Edition | Eclipse Openj9 | semeru-open | semeru-open buildpack
Microsoft | OpenJDK | Hotspot | ms-openjdk | microsoft-openjdk buildpack
Oracle | OracleJDK | Hotspot | oracle | oracle buildpack
Amazon | Corretto | Hotspot | corretto | amazon-corretto
Eclipse Adoptium | Temurin | Hotspot | temurin | temurin buildpack
SAP | Machine | Hotspot | sapmachine | sap-machine buildpack
Alibaba | Dragonwell | Hotspot | dragonwell | alibaba-dragonwell buildpack
Oracle | GraalVM | Graal? | graalvm | graalvm buildpack
Azul | Zulu | Hotspot | azul-zulu | azul-zulu buildpack

Also note that, so far as I'm aware, all of these builds (with the exception of `oracle` and `graalvm`) are open source licensed JDK binaries based on OpenJDK (Oracle's being proprietary). I'm not exactly sure what's meant by `graal`: does it refer to GraalVM Community Edition providing OpenJDK with the Graal JIT inside? Or is it a build back for native images?

Did you leave out Red Hat OpenJDK on purpose? :)

kevin-ortega:
> 
The list is of the currently available Paketo buildpacks that [contribute JREs/JDKs](https://paketo.io/docs/howto/java/#use-an-alternative-jvm) to the lifecycle.   

A Red Hat OpenJDK buildpack does not exist and therefore is not listed.  I don't know the details of each of these buildpacks but most (all?) can contribute a JRE and/or a JDK.    

This RFC is to simply make them easier to use.

mstoodle:
Gotcha, thanks for explaining. Is there another place I could put my "for clarity" comment ?

dmikusa:
>does it refer to GraalVM Community Edition providing OpenJDK with the Graal JIT inside?

Yes. It installs both of the assets, `graalvm-ce-java17-*` and `native-image-installable-svm-*`, from [GraalVM's Github releases](https://github.com/graalvm/graalvm-ce-builds/releases/tag/vm-22.3.0). The assets can be used to build/run Java apps, or build native image binaries.

sophiewigmore:
Jan 4 working group: more technical discussions needed between @kevin-ortega and @dmikusa

text/java/0013-alternate-jvms-in-java-buildpack.md
dmikusa: If we had a Java-specific builder, we could add all of the JVMs. That wouldn't be a problem.

kevin-ortega:
@dmikusa How can we move this RFC forward?  Do we need another RFC to create a language specific builder?

loewenstein:
Just a thought... The problem with including this is the number of layers, isn't it? And that is because builders are build with one layer per component buildpack, right? So instead of requiring a Java specific builder, could we instead have an additional Java meta-buildpack that simply configures all JRE/JDK dependencies? I suppose the creation could easily be automated with references to the single vendor buildpacks - copying in all the dependencies into a single buildpack.toml

It would not require alignment with the builder maintainers (to add yet another builder to their plate). Versioning wise we should be relatively fine, as there's a quarterly release cadence for Java that most of the vendors will just follow. I would expect only a few intermittent dependency changes from vendor specific fixes.

dmikusa:
> Just a thought... The problem with including this is the number of layers, isn't it? And that is because builders are build with one layer per component buildpack, right? So instead of requiring a Java specific builder, could we instead have an additional Java meta-buildpack that simply configures all JRE/JDK dependencies? I suppose the creation could easily be automated with references to the single vendor buildpacks - copying in all the dependencies into a single buildpack.toml

My understanding is that this won't work. The buildpack tools just inline composite/meta buildpacks, so adding one doesn't reduce the layer count. Right now, the builders order groups already link to the language family composites. When the builder is created, those get expanded into all of the layers from each language family composite. What we really need is for the buildpack tools to compact the layers when they generate the builder. Alternatively, we might be able to flatten the composite image in advance, and then generate the builder. If that doesn't break anything with the buildpack tools then that might work too.

I just logged an issue for this, cause I don't think we have yet: https://github.com/buildpacks/pack/issues/1595. Upvotes help.

loewenstein:
> My understanding is that this won't work. The buildpack tools just inline composite/meta buildpacks, so adding one doesn't reduce the layer count.

Well, I was not suggesting to add another buildpack. It was more about creating an alternative one. One that contains all JDK/JRE vendors and allows selecting the one a user chooses. This could replace the BellSoft Liberica one in the existing Java meta-buildpack.

dmikusa:
> > My understanding is that this won't work. The buildpack tools just inline composite/meta buildpacks, so adding one doesn't reduce the layer count.
> 
> Well, I was not suggesting to add another buildpack. It was more about creating an alternative one. One that contains all JDK/JRE vendors and allows selecting the one a user chooses. This could replace the BellSoft Liberica one in the existing Java meta-buildpack.

Oh, ok. I think I understand now, correct me if I'm wrong. You're suggesting one buildpack that knows how to install all of the different JVM vendors. That would technically work, and it wouldn't increase the layer count. It would require some differences in the implementation proposed by this RFC.

1. We'd have to create a new buildpack.
2. We'd have to propose changes to the `paketo-buildpacks/java` buildpack order.
3. We probably also have to address packaging for offline consumption. Putting all of the binaries in one buildpack would make it harder for offline packaging. We'd need to make sure that one could still do an offline package with a subset of JVM vendors.
4. I haven't thought through it all, but I think we'd need to consider if there are any breaking changes this might cause.

I'd also be curious to see if anyone has a preference, for multiple buildpacks/one per vendor, or one buildpack that can install all vendors, and if anyone can think of any advantages/disadvantages to either approach. 

@pivotal-david-osullivan please correct me if I'm wrong, but didn't we talk about moving to a single buildpack for all JVM vendors? I know it would decrease the number of repositories we have to manage, which would be a plus. We might even be able to do away with libjvm.

loewenstein:
> > > My understanding is that this won't work. The buildpack tools just inline composite/meta buildpacks, so adding one doesn't reduce the layer count.
> > 
> > 
> > Well, I was not suggesting to add another buildpack. It was more about creating an alternative one. One that contains all JDK/JRE vendors and allows selecting the one a user chooses. This could replace the BellSoft Liberica one in the existing Java meta-buildpack.
> 
> Oh, ok. I think I understand now, correct me if I'm wrong. You're suggesting one buildpack that knows how to install all of the different JVM vendors. That would technically work, and it wouldn't increase the layer count. It would require some differences in the implementation proposed by this RFC.
> 
> 1. We'd have to create a new buildpack.
> 2. We'd have to propose changes to the `paketo-buildpacks/java` buildpack order.
> 3. We probably also have to address packaging for offline consumption. Putting all of the binaries in one buildpack would make it harder for offline packaging. We'd need to make sure that one could still do an offline package with a subset of JVM vendors.
> 4. I haven't thought through it all, but I think we'd need to consider if there are any breaking changes this might cause.
> 
> I'd also be curious to see if anyone has a preference, for multiple buildpacks/one per vendor, or one buildpack that can install all vendors, and if anyone can think of any advantages/disadvantages to either approach.
> 
> @pivotal-david-osullivan please correct me if I'm wrong, but didn't we talk about moving to a single buildpack for all JVM vendors? I know it would decrease the number of repositories we have to manage, which would be a plus. We might even be able to do away with libjvm.

Being interested in only one vendor, I have a pretty clear preference. I was also not suggesting to remove the individual vendor buildpacks, just add an additional one for the "easy to pick the vendor" experience.
I thought adding a separate buildpack for it should be roughly comparable in cost to adding a whole separate builder for only Java. The main difference would be that the Paketo Builder team wouldn't't need to carry the burden of just how many different vendors the Java language has ;)

pivotal-david-osullivan:
@dmikusa Yes, I had POC'd having one giant buildpack.toml with all JVM vendor dependencies and using a variable (with a default of Bellsoft) to choose which vendor/deps to build with.

I think the question was whether to do this or have separate buildpack.toml's per vendor under 1 repo with libjvm (and again a variable to choose vendor) and adjust the CI to deal with this.

kevin-ortega:
Good discussion.  I have been working on a separate RFC to create a Java-specific builder.   I'm wondering now if I should switch gears.   If we go the route of creating a single JVM buildpack that doesn't necessarily resolve the layer limit and doesn't help in improving the size/speed of building using the full builder that a Java-specific builder can address.

loewenstein:
> Good discussion. I have been working on a separate RFC to create a Java-specific builder. I'm wondering now if I should switch gears. If we go the route of creating a single JVM buildpack that doesn't necessarily resolve the layer limit and doesn't help in improving the size/speed of building using the full builder that a Java-specific builder can address.

How would that not resolve the layer limit? There is currently one JVM buildpack (providing Bellsoft Liberica) and there would be one JVM buildpack but providing a choice of JVM implementations. I.e. we keep the number of layers as it is but add additional JVMs with ease of use.

loewenstein:
> please correct me if I'm wrong, but didn't we talk about moving to a single buildpack for all JVM vendors? I know it would decrease the number of repositories we have to manage, which would be a plus. We might even be able to do away with libjvm.

If we replace the multiple JVM buildpacks with just one, we will probably have to maintain a fork (to reduce the JVMs to just the sap-machine and add another proprietary one).

One thing that just came to my mind, although probably not a short-term solution, wouldn't the dependency outsourcing ideas also solve the problem?

kevin-ortega:
> > Good discussion. I have been working on a separate RFC to create a Java-specific builder. I'm wondering now if I should switch gears. If we go the route of creating a single JVM buildpack that doesn't necessarily resolve the layer limit and doesn't help in improving the size/speed of building using the full builder that a Java-specific builder can address.
> 
> How would that not resolve the layer limit? There is currently one JVM buildpack (providing Bellsoft Liberica) and there would be one JVM buildpack but providing a choice of JVM implementations. I.e. we keep the number of layers as it is but add additional JVMs with ease of use.

Yes, we replace one JVM buildpack for another but it doesn't reduce the overall number of buildpacks in the builder.  I'm not saying having a single JVM buildpack is not a reasonable alternative, it just doesn't resolve all the issues that a Java-specific builder does.

SaschaSchwarze0:
Just my few cents ...

As being interested in multi-JVM support, and having followed the discussion ... I think that the problem needs to be resolved outside of the JVM buildpacks. Paketo has chosen an approach with very fine-grained buildpacks. This leads to many buildpacks being required to be bundled into a builder. If there is a limitation that is being hit, then this limitation will not just be hit by adding more JVM buildpacks, but would also eventually be hit if, for example, buildpacks of other languages graduate such as rust, and want to become part of the full-builder. That makes me conclude that the main problem is the limitted number of buildpacks that can be added implied by the limitation of the number of image layers. Everything else that is being discussed here are all good ideas but in my opinion only workarounds.

dmikusa:
Ok, now that #274 is merged I think we can unblock this.

My suggestion as the easiest path forward would be to target the new language family builder for inclusion of all of our JVM providers. This builder won't need to worry about the layer limits.

Folks can start using that and as other technology problems are addressed we can talk about bringing more JVM providers into the main Paketo builders.

kevin-ortega:
> Ok, now that #274 is merged I think we can unblock this.
> 
> My suggestion as the easiest path forward would be to target the new language family builder for inclusion of all of our JVM providers. This builder won't need to worry about the layer limits.
> 
> Folks can start using that and as other technology problems are addressed we can talk about bringing more JVM providers into the main Paketo builders.

I have update the RFC to indicate this should be initially implemented in the language family builder.

text/java/0000-alternate-jvms-in-java-buildpack.md
```diff
@@ -0,0 +1,57 @@
...
+| Amazon Corretto    | corretto       |  amazon-corretto                    |
+| Adoptium           | adoptium       |  adoptium buildpack                 |
+| SAP Machine        | sapmachine     |  sap-machine buildpack              |
+| Alibaba Dragonwell | dragonwell     |  alibaba-dragonwell buildpack       |
+| GraalVM            | graalvm        |  graalvm buildpack                  |
+| Azul Zulu          | azul-zulu      |  azul-zulu buildpack                |
+
+Additional JVMs can easily be added to the Java buildpack by implementing this RFC.  
+
+Initially, this RFC should be implemented in the new language family builder.  If and when the layer limit issue is resolved more JVM providers can be added to the main Paketo builders.
```
loewenstein, line 49: Is this meant to be implemented in a separate composite buildpack?

kevin-ortega: > Is this meant to be implemented in a separate composite buildpack?

No, this intended for the current java composite buildpack via the new language family builder.

loewenstein: How would that same composite buildpack be added to the normal builders without running into the layer limit? Or is the plan that Java is going to be removed from the other builders?

kevin-ortega: > How would that same composite buildpack be added to the normal builders without running into the layer limit? Or is the plan that Java is going to be removed from the other builders?

The java language family builder will include all the JVM providers in addition to the existing java composite buildpack.  The normal Paketo builders contain only the java composite buildpack.

loewenstein: Get it, but then the first paragraph in implementation is a bit misleading, isn't it? The composite buildpack would continue to only include the bellsoft liberica buildpack, right?

With all the other JVM buildpacks coming first now, how do we plan to handle defaulting to the liberica JVM?

loewenstein: Or is the idea to not have a composite buildpack for that case at all, but instead directly put in the component buildpacks into the builder's order groups?

dmikusa: I think the way it originally reads is OK. 

My understanding and preference is this:

1. No additional composite buildpack. I'm not strongly opposed, but I'm not sure it's necessary and there is overhead to keep an additional composite buildpack up-to-date.
2. No changes to the existing Java composite buildpacks
3. We add these additional buildpacks to the new Java language family builder. They'll be added directly to that builder. I think a buildplan like this should work:

    - Adoptium
    - Alibaba Dragonwell
    - ... rest of the JVM provider buildpacks in alphabetic order
    - Java composite buildpack

    This will duplicate Bellsoft Liberica but I don't think that should be an issue. By the time it runs the second time, one of the other buildpacks will have already run and contributed a JVM.
4. As Kevin mentioned, `BP_JVM_VENDOR` will be used to pick the JVM vendor of choice.

dmikusa (approved):

text/java/0000-alternate-jvms-in-java-buildpack.md
loewenstein: > 4\. As Kevin mentioned, `BP_JVM_VENDOR` will be used to pick the JVM vendor of choice.

So, the idea is that the Bellsoft buildpack comes first and picks the JRE provider spot unless BP_JVM_VENDOR is set to something other than bellsoft?

What do the other component buildpacks do?

Background is that I want to avoid that e.g. the sap-machine buildpack is forced to yield unless BP_JVM_VENDOR is set.

dmikusa: >So, the idea is that the Bellsoft buildpack comes first and picks the JRE provider spot unless BP_JVM_VENDOR is set to something other than bellsoft?

Oh, yes. Right. The default needs to come first in the order group. The only reliable way to set the default is if that env variable is empty. This is the same as what we're doing with app servers. That pattern has proven to work well.

>What do the other component buildpacks do?

The idea is for this to work like what we do with app servers. Taking from Tomcat CNB's `detect.go`.

```
	appServer, _ := cr.Resolve("BP_JAVA_APP_SERVER")
	if appServer != "" && appServer != JavaAppServerTomcat {
		d.Logger.Debugf("failed to match requested app server of [%s], buildpack supports [%s]", appServer, JavaAppServerTomcat)
		return libcnb.DetectResult{Pass: false}, nil
	}
```

Essentially, what we'll do is add in logic like this to libjvm's `detect.go`.

1. If `BP_JVM_VENDOR` is empty then 👍  proceed to detect as normal
2. If `BP_JVM_VENDOR` is not empty, but its value matches the key for this JVM vendor then 👍 proceed to detect as normal.
3. Any other cases, fail detection.

Because this happens at detection time, I believe that the only time you should see two JVM vendor buildpacks run their build's is with the default, i.e. Liberica. This is because it'll technically be in the order group twice. Once at the beginning to enforce the default and once in the Java composite. That shouldn't cause problems though. The second time it runs should be a no-op. Libjvm [checks the buildplan before outputting anything though](https://github.com/paketo-buildpacks/libjvm/blob/main/build.go#L98-L100) so it shouldn't be visible in the output.

>Background is that I want to avoid that e.g. the sap-machine buildpack is forced to yield unless BP_JVM_VENDOR is set.

After explaining that more, does that work for you? If not, can you expand on your requirement here? 

Thanks

loewenstein: I think that fits. Thanks @dmikusa.

pivotal-david-osullivan (approved):

dmikusa:
Votes are in. This RFC is approved.

PR #269, opened 2022-12-21
Adds new Java RFC for composite draft branches

pivotal-david-osullivan:
[Readable](https://github.com/paketo-buildpacks/rfcs/blob/composite-draft-branch/text/java/0000-composite-draft-branch.md)

## Summary
This RFC proposes a change to the pipeline-builder CI which will target the composite/meta buildpack repositories, e.g. paketo-buildpacks/java, and change the branch that component buildpack bump PRs are created against. 

## Use Cases
This will allow integration tests to be run on merges to the main branch once component buildpack bumps are in and ready for release. This should also reduce the maintenance burden as the component bumps can be auto-merged with review possible before the merge to main.

## Checklist

### DIFF
--- a/text/java/0000-composite-draft-branch.md
+++ b/text/java/0000-composite-draft-branch.md
@@ -1,0 +1,39 @@
+# Create a draft branch for composite/meta buildpacks
+
+## Summary
+
+Currently component buildpack bumps are created as PRs against the main branch, e.g. see [here](https://github.com/paketo-buildpacks/java/pull/873).
+This RFC proposes changing the CI to create these dependency PRs against a new branch, e.g. `draft`.    
+
+## Motivation
+
+This will allow us to achieve the following improvements:
+
+1. Auto-merging of component buildpack PRs into this draft branch 
+2. Manual review can be performed with a single PR before merging to main
+2. Integration tests can be run on merging to main ensuring all changes in the release are covered
+
+## Detailed Explanation
+
+Currently the component dependency PRs must be manually approved and merged separately, even though their changes do not cause merge conflicts. Auto-merging these will reduce the maintainence burden involved here.
+Integration tests can configured to run on merges to the main branch, which could happen once per week before a release, or more frequently if necessary for a specific purpose.
+
+## Rationale and Alternatives
+
+1. Do not change the current CI/branch configuration - integration tests could be run via a manual step before release.
+2. Do not change the current CI/branch configuration - create new release tags as a separate initial step before a release is published. Integration tests could be run on tag creation instead, with the publish action potentially triggered from passed tests. 
+
+## Implementation
+
+For the composite/meta buildpack repositories:
+
+1. The pipeline-builder CI workflow step that updates package [dependencies](https://github.com/paketo-buildpacks/pipeline-builder/blob/main/octo/package_dependencies.go) would be changed to checkout a draft branch rather than the default of main. The PR for containing the update would then be created targeting this branch. 
+2. We would likely need a new workflow which triggers on merges to the 'draft' branch and opens/updates a PR to merge the dependency changes into main. The integration tests could be triggered here also.
+
+## Prior Art
+
+N/A
+
+## Unresolved Questions and Bikeshedding
+
+N/A
--- a/text/java/0000-composite-draft-branch.md
+++ b/text/java/0000-composite-draft-branch.md
@@ -3,7 +3,7 @@
 ## Summary
 
 Currently component buildpack bumps are created as PRs against the main branch, e.g. see [here](https://github.com/paketo-buildpacks/java/pull/873).
-This RFC proposes changing the CI to create these dependency PRs against a new branch, e.g. `draft`.    
+This RFC proposes changing the CI to create these dependency bumps as commits to a PR against a new branch, e.g. `draft`.    
 
 ## Motivation
 
@@ -32,7 +32,7 @@ For the composite/meta buildpack repositories:
 
 ## Prior Art
 
-N/A
+This is how `buildpack.toml` is updated in non-Java composite buildpacks, see [this PR](https://github.com/paketo-buildpacks/python/pull/555/files)
 
 ## Unresolved Questions and Bikeshedding
 


text/java/0000-composite-draft-branch.md
```diff
@@ -0,0 +1,39 @@
...
+
+Currently component buildpack bumps are created as PRs against the main branch, e.g. see [here](https://github.com/paketo-buildpacks/java/pull/873).
+This RFC proposes changing the CI to create these dependency PRs against a new branch, e.g. `draft`.    
+
+## Motivation
+
+This will allow us to achieve the following improvements:
+
+1. Auto-merging of component buildpack PRs into this draft branch 
+2. Manual review can be performed with a single PR before merging to main
```
robdimsdale, line 13: It sounds like the main goal of this proposal is to batch the various buildpack bumps to (presumably) reduce the human burden by reviewing fewer, larger changes instead of more, smaller changes.

If so, I don't think you need to open a PR against a `draft` branch. Instead I think you can just push each change to this draft branch (e.g. `buildpack-bumps`) and have a single PR from this branch to `main`. As long as you rebase each change on top of this branch before pushing it, the single PR from `buildpack-bumps` to `main` will contain all the changes, and can be reviewed as a single PR on whatever cadence is useful for the java maintainers. You can run your integration tests on this PR the same as if you opened it manually.

This is how the other Paketo buildpacks work (i.e. the buildpacks that don't use `pipeline-builder`). See [this Python Buildpack PR](https://github.com/paketo-buildpacks/python/pull/555/files) as an example - it contains automated bumps to multiple component buildpacks in a single PR, with automated integration tests. A human can review this PR and manually merge on whatever schedule works for the maintainers of that repository.

I think your proposal will work, and it might reduce the frequency on which Java maintainers review buildpack bumps, but I think you can save yourself even more work if you adopt the flow identified above.

sophiewigmore:
Notes from working group 12/21 - this is a good opportunity to make the Java and non-Java workflows more similar. This RFC has the potential to make those workflows more separate, so we should think about the ramifications of that.

text/java/0000-composite-draft-branch.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Detailed Explanation
+
+Currently the component dependency PRs must be manually approved and merged separately, even though their changes do not cause merge conflicts. Auto-merging these will reduce the maintainence burden involved here.
+Integration tests can configured to run on merges to the main branch, which could happen once per week before a release, or more frequently if necessary for a specific purpose.
+
+## Rationale and Alternatives
+
+1. Do not change the current CI/branch configuration - integration tests could be run via a manual step before release.
+2. Do not change the current CI/branch configuration - create new release tags as a separate initial step before a release is published. Integration tests could be run on tag creation instead, with the publish action potentially triggered from passed tests.
```
loewenstein, line 24: Should we maybe add - and of course evaluate - using [Renovate](https://docs.renovatebot.com) as an alternative or additional means to achieve our goals here?

1. There is a nice [Dependency Dashboard](https://docs.renovatebot.com/key-concepts/dashboard/), showing a repo's dependency and deferred updates and allows to control which update PRs are done when (see [Dependency Dashboard Approval workflow](https://docs.renovatebot.com/key-concepts/dashboard/#dependency-dashboard-approval-workflow).
2. You can e.g. [group update PRs](https://docs.renovatebot.com/key-concepts/pull-requests/#grouped-updates-with-different-versions) to reduce the number of individual PRs. Applying all patch updates in a batch PR is an easy option.
3. Renovate can add labels, so we could control to what extend CI will execute tests depending on the updates we do.

The "easy to use" might hit it's limit when we look at updating dependencies like Tomcat, because there is a limited number of [datasources](https://docs.renovatebot.com/modules/datasource/) Renovate supports. On the other hand side, [Tomcat seems to at least have proper version tags](https://github.com/apache/tomcat/tags).

TL;DR We should investigate how far we can lower the CI resources used for testing and review effort required merging with using Renovate. If this evaluation runs successfully, it might be worth rolling this out to other Paketo repositories via a separate, project scoped RFC.

cc @pbusko

loewenstein:
> Notes from working group 12/21 - this is a good opportunity to make the Java and non-Java workflows more similar. This RFC has the potential to make those workflows more separate, so we should think about the ramifications of that.

See also my comment on evaluating Renovate, potentially for the whole project.

text/java/0000-composite-draft-branch.md
robdimsdale: Ah, great! Glad we're on the same page.

text/java/0000-composite-draft-branch.md
robdimsdale: @loewenstein I think renovate could be a useful tool to explore. If we are able to get all the paketo buildpacks using the same tooling/workflows, then refactoring to use a tool like Renovate would impact all buildpacks equally.

Without digging into whether renovate can solve this particular problem in this RFC - bumping component buildpack versions in the meta buildpack - I think a good first step would be to unify the workflows of the buildpacks.

Also, the way I read this RFC, it doesn't try to address how dependencies (e.g. Tomact) are bumped within the component buildpacks.

loewenstein: @robdimsdale well, the RFC sounds like there is effort involved changing CI and while maybe not too much still significant. So I thought rather than aligning first and then changing we could just as well explore the alternative and align in that - if it turns out to be the superior choice.

Vacation season has started over here, but we have a bit of experience with Renovate, so we could probably provide some insights and spend some cycles, if this is considered useful.

dmikusa: I would like to explore Renovate. I'm not happy with Dependabot, and it not updating transitive dependencies. I would also like to see if we could get Renovate to replace our CI jobs that run to update composite buildpacks (i.e. teach it to know about dependent buildpacks and submit PRs for those as well).

My $0.02 is that it should be something separate we explore though, to not hold up the changes here. Also as @robdimsdale mentioned, it would really be project-wide. I think a good first step would be to do an exploration, possibly in the [explorations repo](https://github.com/paketo-community/explorations) or maybe as a separate repo since we will probably need to adjust CI & integrate with the Renovate services.

Another concern with Renovate is whether can we get a managed service of it that provides all the functionality we require, or if there is something we need to host. If it's the latter, we have to be very careful with the amount of maintenance that requires.

pbusko: > I would also like to see if we could get Renovate to replace our CI jobs that run to update composite buildpacks (i.e. teach it to know about dependent buildpacks and submit PRs for those as well).

We're using Renovate for updating composite buildpack dependencies, as well as builder dependencies, like lifecycle version, stack versions and included buildpacks with the [regex manager](https://docs.renovatebot.com/modules/manager/regex/) feature + [github releases datasource](https://docs.renovatebot.com/modules/datasource/#github-releases-datasource).

> Another concern with Renovate is whether can we get a managed service of it that provides all the functionality we require, or if there is something we need to host. If it's the latter, we have to be very careful with the amount of maintenance that requires.

In a nutshell, renovate is just a CLI, which makes it extremely easy to self-host (github action (official one is available)/k8s job/any runner capable of calling renovate CLI). Also there's a [free GitHub App](https://github.com/apps/renovate) ready for use. Functionality-wise all options are equal.

ryanmoran:
Jan 4 working group:  one of the goals for the new year is to get the java/utilities buildpacks in alignment with the non-java buildpacks. This RFC has project-level changes, so it would be good to get some feedback from people in the Java and non-Java side.

text/java/0000-composite-draft-branch.md
dmikusa: I would like to propose that we move the Renovate topic to a new RFC. It really is something that should be Paketo-wide and this is a Java-specific RFC so I don't think it has the visibility necessary. Also, we have [specific needs for this RFC](https://github.com/paketo-buildpacks/java/pull/828) which we don't want to block by a larger discussion on Renovate.

If everyone is OK with this, please 👍 this comment. Thanks

dmikusa:
@ryanmoran @loewenstein @sophiewigmore - Aside from Renovate, which [I'm requesting we table for a separate RFC](https://github.com/paketo-buildpacks/rfcs/pull/269#discussion_r1063048837), I'm not sure I see how this would take the Java buildpacks more out of alignment. 

The idea here is to enable a workflow that would allow us to trigger integration tests more easily off the Java composite buildpack. The implementation details may be slightly different, but adding integration tests, which Java buildpacks don't presently have, brings them more inline with other Paketo buildpacks. I'm also hopeful it'll help flush out issues earlier and not in the builder pipelines, which has been a problem.

I believe @pivotal-david-osullivan was going to adjust some of the text based on @robdimsdale's comment, hopefully that'll clarify as well, but if you have any other concerns let me or David know. Thanks

loewenstein:
@dmikusa My [main take on this](https://github.com/paketo-buildpacks/rfcs/pull/269#discussion_r1054707766) was
> well, the RFC sounds like there is effort involved changing CI and while maybe not too much still significant.

I agree that getting integration tests is very valuable, that's why we started a PR to contribute some ;)

TL;DR If you think spending a little extra now instead of moving straight to Renovate will significantly speed up having integration tests, maybe it's worth it.

dmikusa:
@loewenstein OK, I resolved that conversation. I do think the best way forward is to keep the scope on this small, get it approved, then we can merge your PR and circle back to Renovate & dependency management.

text/java/0000-composite-draft-branch.md
pivotal-david-osullivan: I made a small update to clarify how I intended this to work, as you said @robdimsdale, it would be multiple commits to one draft PR, as per non-Java buildpacks. Thanks!

robdimsdale: Thanks. I'm glad we're in alignment regarding the high-level goal. The RFC still references a two-step approach of automatic merge of PRs to the `draft` branch followed by a manual merge of PRs to `main`. Maybe I'm missing something, but I think that can be simplified to automatically merging certain PRs directly to `main`. Or at least batching up the changes into a single PR to `main` which is manually merged. This is how it's done in the example you linked to in the Python buildpack.

Do we need the intermediate `draft` branch? Am I missing something?

dmikusa: We can't merge directly to main, but this is my understanding:

>Or at least batching up the changes into a single PR to main which is manually merged.

I think `draft` here is just an implementation detail. You need to have a branch to make a PR, so the proposal is calling that `draft`.

robdimsdale: Right, ok. I think I'm getting hung up on the semantics. In order to "merge" a PR, there already must be a branch open. You're proposing calling that branch `draft` - in the python example lined that branch is called `automation/buildpack.toml/update`. I thought you were proposing opening one branch e.g. `automation`, merging that to `draft`, and then merging that to `main`.

As long as there's only one branch (and hence PR to `main`), I think this is structurally identical to what we have on the other buildpacks.

pivotal-david-osullivan: That's correct Rob 👍 sorry for the confusion. Are you happy for me to resolve this thread?

dmikusa:
Looks like commenting has settled down. I'd like to start the final call for comments & votes with Feb 7th, 2023 being the final day for comment/vote. Thanks!

dmikusa (approved):

dmikusa:
OK, thanks everyone for the comments and feedback. Votes are in and sufficient to pass. I'll go ahead and merge then assign a RFC number.

PR #273, opened 2023-01-11
RFC: Selecting the next default Java Version

modulo11:
## Summary

This RFC proposes a general pattern/process on when to update the default Java version used in the respective buildpack. See https://github.com/paketo-buildpacks/maven/issues/217#issuecomment-1345375874 for more details.

## Use Cases

Users can expect the default version of Java to advance in a regular manner.

## Checklist

CC @loewenstein @pbusko

### DIFF
--- a/text/0000-selecting-default-java-version.md
+++ b/text/0000-selecting-default-java-version.md
@@ -22,7 +22,9 @@ The default should be changed according to a defined process rather than at some
 
 ## Implementation
 
-To fulfill this RFC, the default version should be changed to Java 17. All future versions should be bumped according to this RFC.
+To fulfill this RFC, the default version should be changed to Java 17 immediately. 
+
+Going forward, the default Java version should be bumped according to this RFC.
 
 ## Prior Art
 
--- a/text/0000-selecting-default-java-version.md
+++ b/text/0000-selecting-default-java-version.md
@@ -10,7 +10,7 @@ Currently, there is no process how the default version of Java is picked in the 
 
 ## Detailed Explanation
 
-The default should be changed according to a defined process rather than at some undefined point in time to some undefined version. To do so, the default should be changed once the latest releases LTS version of Java is at least one year old. 
+The default should be changed according to a defined process rather than at some undefined point in time to some undefined version. To do so, the default should be changed once the latest releases LTS version of Java is at least one year old. Setting this as the baseline, exceptions of faster (or slower) updates are acceptable in urgent cases.
 
 ## Rationale and Alternatives
 
@@ -22,13 +22,9 @@ The default should be changed according to a defined process rather than at some
 
 ## Implementation
 
-To fulfill this RFC, the default version should be changed to Java 17 immediately. 
+To fulfill this RFC, the default version should be changed to Java 17 immediately. Additionally, the process on how the community is notified about the update should be documented on paketo.io/docs.
 
 Going forward, the default Java version should be bumped according to this RFC.
-
-## Prior Art
-
-n/a
 
 ## Unresolved Questions and Bikeshedding
 
--- a/text/0000-selecting-default-java-version.md
+++ b/text/0000-selecting-default-java-version.md
@@ -10,7 +10,7 @@ Currently, there is no process how the default version of Java is picked in the 
 
 ## Detailed Explanation
 
-The default should be changed according to a defined process rather than at some undefined point in time to some undefined version. To do so, the default should be changed once the latest releases LTS version of Java is at least one year old. Setting this as the baseline, exceptions of faster (or slower) updates are acceptable in urgent cases.
+The default should be changed according to a defined process rather than at some undefined point in time to some undefined version. To do so, the default should be changed once the latest released LTS version of Java is at least one year old. Exceptions are permitted for a shorter change cycle when deemed sufficiently urgent by a majority vote of Java subteam contributors and maintainers.
 
 ## Rationale and Alternatives
 
--- a/text/0000-selecting-default-java-version.md
+++ b/text/0000-selecting-default-java-version.md
@@ -1,0 +1,34 @@
+# Selecting the next default Java Version for Paketo Buildpacks
+
+## Summary
+
+This RFC introduces a process to select the next default version of the Paketo Java Buildpack.
+
+## Motivation
+
+Currently, there is no process how the default version of Java is picked in the Paketo world. As of today (January 2023), the default still is Java 11 (released September 2018) although Java 17 (released September 2021) is already available.
+
+## Detailed Explanation
+
+The default should be changed according to a defined process rather than at some undefined point in time to some undefined version. To do so, the default should be changed once the latest releases LTS version of Java is at least one year old. 
+
+## Rationale and Alternatives
+
+- Do nothing, update default manually from time to time
+- Always use the latest LTS immediately
+  - Could break people immediately
+- This RFC
+  - Gives room to adapt to breaking changes
+
+## Implementation
+
+To fulfill this RFC, the default version should be changed to Java 17. All future versions should be bumped according to this RFC.
+
+## Prior Art
+
+n/a
+
+## Unresolved Questions and Bikeshedding
+
+- Define the time to wait after a new LTS was published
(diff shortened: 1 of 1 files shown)


text/0000-selecting-default-java-version.md
```diff
@@ -0,0 +1,34 @@
...
+
+- Do nothing, update default manually from time to time
+- Always use the latest LTS immediately
+  - Could break people immediately
+- This RFC
+  - Gives room to adapt to breaking changes
+
+## Implementation
+
+To fulfill this RFC, the default version should be changed to Java 17. All future versions should be bumped according to this RFC.
```
dmikusa, line 25:
```suggestion
To fulfill this RFC, the default version should be changed to Java 17 immediately. 

Going forward, the default Java version should be bumped according to this RFC.
```

Just trying to make it clear the immediate action and then what we'll do going forward.

text/0000-selecting-default-java-version.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Rationale and Alternatives
+
+- Do nothing, update default manually from time to time
+- Always use the latest LTS immediately
+  - Could break people immediately
+- This RFC
+  - Gives room to adapt to breaking changes
+
+## Implementation
```
dmikusa, line 23: I don't think we need a ton of action here, but I would like to have the implementation include documenting this process on paketo.io/docs.

I would also like to see a description of the notification channels/process that should be used to alert folks of the upcoming change.

Like at 3 months prior to the change, we'll post on Slack & open a Github issue to upgrade (maybe we also embed a message in the buildpack output?).  At 1 month prior, we'll post a Paketo blog entry on the topic. Then after the change we'll post another blog post when the default change has been released.

I'm not saying it has to be exactly that, but I think we should document what folks can expect in terms of notifications also.

text/0000-selecting-default-java-version.md
```diff
@@ -0,0 +1,34 @@
...
+
+To fulfill this RFC, the default version should be changed to Java 17. All future versions should be bumped according to this RFC.
+
+## Prior Art
+
+n/a
+
+## Unresolved Questions and Bikeshedding
+
+- Define the time to wait after a new LTS was published
```
dmikusa, line 33: I like the way it's worded now "at least one year". It's not always possible to do this at exactly one year, so that gives us some room to adjust if that's necessary.

text/0000-selecting-default-java-version.md
```diff
@@ -0,0 +1,34 @@
...
+- Should this be an automated process?
```
dmikusa, line 34: My vote would be no. I don't see this as being a ton of work, so I'm not sure if it meets the payoff for automating. That said, if someone wants to automate it and submit a PR with a Github Action to do that, then we would accept it.

Even if it's not a full automation, like if it just tracks upstream and opens an issue when a year's time has elapsed, that would be helpful too.

text/0000-selecting-default-java-version.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Implementation
+
+To fulfill this RFC, the default version should be changed to Java 17. All future versions should be bumped according to this RFC.
+
+## Prior Art
+
+n/a
+
+## Unresolved Questions and Bikeshedding
```
dmikusa, line 31: Two more questions:

- Do we want to allow for exceptions to the "at least one year" rule?

I'm not sure what/if that would be necessary, but right now there is no provision to bump the default early, if something exceptional happened.

We could add something like, "If there is a critical reason and a majority vote of all Java contributors and maintainers passes then the LTS version may be switched after 6 months time, but no earlier."

- Do we want to require an RFC to make the switch?

I think this would be more of a formality & to help with alerting folks of the change. What do folks think about this? Is it overkill? Or helpful?

dmikusa (changes requested):

text/0000-selecting-default-java-version.md
kevin-ortega: I think there should be a provision to override the one year rule and a majority vote seems reasonable. 

I don't feel an RFC is necessary.  Java contributors can be notified via slack or blogs of the upcoming change.

c0d1ngm0nk3y: > Is it overkill?

I think so. If a majority vote of all Java maintainers is in favour because of some good reason, why not be pragmatic and change it? The process should more describe the "normal" process, imho. There ca always be exceptions.

text/0000-selecting-default-java-version.md
modulo11: I added a sentence about documenting the process. I like your proposal, but I'm not sure about adding it to the build output.

text/0000-selecting-default-java-version.md
modulo11: I agree with @kevin-ortega and and @c0d1ngm0nk3y. Added a sentence to the RFC. Would that be sufficient?

text/0000-selecting-default-java-version.md
```diff
@@ -10,7 +10,7 @@ Currently, there is no process how the default version of Java is picked in the
 
 ## Detailed Explanation
 
-The default should be changed according to a defined process rather than at some undefined point in time to some undefined version. To do so, the default should be changed once the latest releases LTS version of Java is at least one year old. 
+The default should be changed according to a defined process rather than at some undefined point in time to some undefined version. To do so, the default should be changed once the latest releases LTS version of Java is at least one year old. Setting this as the baseline, exceptions of faster (or slower) updates are acceptable in urgent cases.
```
dmikusa, line 13:
```suggestion
The default should be changed according to a defined process rather than at some undefined point in time to some undefined version. To do so, the default should be changed once the latest released LTS version of Java is at least one year old. Exceptions are permitted for a shorter change cycle when deemed sufficiently urgent by a majority vote of Java subteam contributors and maintainers.
```

I would really recommend having the exception process spelled out more. The purpose of this RFC is to provide a stable process on which folks can depend. A predictable change cycle is a big part of that. 

If we say that the version can be changed in a shorter period of time, but does not provide any requirements then it makes the process opaque and erodes the comfort we're trying to provide for users.

This specifically says that it cannot be done by just a small number of folks. Technically two maintainers, one to commit and one to approve, could decide to change this unless we specify otherwise. Specifically saying that it requires a majority vote, makes the process harder and something most of the Java maintainers & contributors need to agree upon.

pivotal-david-osullivan:
Should we decide on a length of time (e.g. 1 month) to wait before implementing the RFC & making the first default version change, i.e from 11 to 17? We would post in the docs, channels, an issue, maybe a blog post etc. and then make the change.

dmikusa:
@pivotal-david-osullivan - I'm 👍 for that. For an OSS project, 1 month from the date of a blog post announcing the change sounds reasonable to me. During that time, we can update the docs, post a couple of reminders to Slack, and have an issue open on Github as a placeholder for the work.

dmikusa (approved):

dmikusa:
Looks like commenting has settled down. I'd like to start the final call for comments & votes with Feb 5th, 2023 being the final day for comment/vote. Thanks!

anthonydahanne:
Hello!
I've just experienced the issue where using `pack --builder=paketobuildpacks/builder:base` against a new Spring Boot project (defining `java.version`) ends up with a JVM 11 being picked up and a failed build (https://github.com/paketo-buildpacks/maven/issues/217) - and I certainly don't want to delay this RFC that would fix this failing default user experience without any additional configuration (new users aren't familiar with `BP_JVM_VERSION` I presume). 

But... I also wish we could, as a further improvement (RFC?), parse the `pom.xml` or `build.gradle` and try and find a version; for sure it won't be a perfect detection (nor a "clean" implementation since it'd need to happen early) compared to a `MANIFEST.MF` parsing, but I believe it'd be a better guess than picking a 1 year old Java LTS version - after all the next LTS is out in 6 months...

loewenstein:
@anthonydahanne you might be interested to comment on https://github.com/paketo-buildpacks/rfcs/pull/268

pivotal-david-osullivan (approved):

PR #274, opened 2023-01-18
Proposal to create language family builders

kevin-ortega:
[Readable](https://github.com/kevin-ortega/rfcs/blob/create-java-specific-builder/text/0000-create-language-family-builders.md)

### DIFF
--- a/text/0000-create-language-family-builders.md
+++ b/text/0000-create-language-family-builders.md
@@ -59,15 +59,6 @@ description = "Ubuntu bionic base image with buildpacks for Java, including all 
   uri = "docker://gcr.io/paketo-buildpacks/eclipse-openj9"
 
 [[buildpacks]]
-  uri = "docker://gcr.io/paketo-buildpacks/syft"
-
-[[buildpacks]]
-  uri = "docker://gcr.io/paketo-buildpacks/gradle"
-
-[[buildpacks]]
-  uri = "docker://gcr.io/paketo-buildpacks/maven"
-
-[[buildpacks]]
   uri = "docker://gcr.io/paketo-buildpacks/ca-certificates"     
 
 [[buildpacks]]
@@ -92,67 +83,9 @@ description = "Ubuntu bionic base image with buildpacks for Java, including all 
   uri = "docker://gcr.io/paketo-buildpacks/azul-zulu"
 
 [[buildpacks]]
-  uri = "docker://gcr.io/paketo-buildpacks/leiningen"
-
-[[buildpacks]]
-  uri = "docker://gcr.io/paketo-buildpacks/clojure-tools"
-
-[[buildpacks]]
-  uri = "docker://gcr.io/paketo-buildpacks/sbt"             
-
-[[buildpacks]]
-  uri = "docker://gcr.io/paketo-buildpacks/watchexec"
-
-[[buildpacks]]
-  uri = "docker://gcr.io/paketo-buildpacks/executable-jar"
-
-[[buildpacks]]
-  uri = "docker://gcr.io/paketo-buildpacks/apache-tomcat"
-
-[[buildpacks]]
-  uri = "docker://gcr.io/paketo-buildpacks/apache-tomee"
-
-[[buildpacks]]
-  uri = "docker://gcr.io/paketo-buildpacks/dist-zip"    
-
-[[buildpacks]]
-  uri = "docker://gcr.io/paketo-buildpacks/spring-boot"             
-
-[[buildpacks]]
-  uri = "docker://gcr.io/paketo-buildpacks/procfile"  
-
-[[buildpacks]]
-  uri = "docker://gcr.io/paketo-buildpacks/jattach"  
-
-[[buildpacks]]
-  uri = "docker://gcr.io/paketo-buildpacks/azure-application-insights"  
-
-[[buildpacks]]
-  uri = "docker://gcr.io/paketo-buildpacks/google-stackdriver"  
-
-[[buildpacks]]
-  uri = "docker://gcr.io/paketo-buildpacks/datadog"  
-
-[[buildpacks]]
-  uri = "docker://gcr.io/paketo-buildpacks/java-memory-assistant"  
-
-[[buildpacks]]
-  uri = "docker://gcr.io/paketo-buildpacks/encrypt-at-rest"  
-
-[[buildpacks]]
-  uri = "docker://gcr.io/paketo-buildpacks/environment-variables"  
-
-[[buildpacks]]
-  uri = "docker://gcr.io/paketo-buildpacks/image-labels"
-
-[[buildpacks]]
-  uri = "docker://gcr.io/paketo-buildpacks/liberty"  
+  uri = "docker://gcr.io/paketo-buildpacks/java"  
 
 [[order]]
-
-  [[order.group]]
-    id = "paketo-buildpacks/ca-certificates"
-    optional = true
 
   [[order.group]]
     id = "paketo-buildpacks/bellsoft-liberica"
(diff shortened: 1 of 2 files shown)


ryanmoran:
Today we offer 3 different "flavors" of builder based on the stack they align with. How would we deliver a coherent set of language-specific builders as we offer more stacks (or "flavors") going forward?

For instance, what does it mean to have a builder called `paketobuildpacks/builder:nodejs` or `paketobuildpacks/builder:go`? Would the Node.js builder build on top of the `base` Jammy stack while the Go builder built off the `tiny` stack? And what of other stacks? Where does this leave UBI or a future Alpine stack?

kevin-ortega:
> Today we offer 3 different "flavors" of builder based on the stack they align with. How would we deliver a coherent set of language-specific builders as we offer more stacks (or "flavors") going forward?
> 
> For instance, what does it mean to have a builder called `paketobuildpacks/builder:nodejs` or `paketobuildpacks/builder:go`? Would the Node.js builder build on top of the `base` Jammy stack while the Go builder built off the `tiny` stack? And what of other stacks? Where does this leave UBI or a future Alpine stack?

I'm updating the RFC on how to deal with jammy, ubi and future ubuntu-based stacks.  I'm not familiar enough with the other language families and the full, base, tiny flavors and what each family would need.    Is it sufficient for the RFC to suggest each language family will need to choose what stack their respective builder should be based on?

ryanmoran:
I think I can get onboard with the current language about the variants that would be available for each language family.

I also see that the `builder.toml` example provided composes the component buildpacks that make up the current Java buildpack. Is the expectation that we would get rid of the language-family buildpacks and replace them with builders. My concern with this is that we end up with significant duplication, or even incidental skew in the buildpacks included in each builder. It also prevents someone from easily using a "buildpackless" builder and the language-family buildpack to build.

What are the benefits of pushing the language-family order configurations into the builders?

sophiewigmore:
WG notes:
Looking for feedback from maintainers of other language families and builders

kevin-ortega:
> I think I can get onboard with the current language about the variants that would be available for each language family.
> 
> I also see that the `builder.toml` example provided composes the component buildpacks that make up the current Java buildpack. Is the expectation that we would get rid of the language-family buildpacks and replace them with builders. My concern with this is that we end up with significant duplication, or even incidental skew in the buildpacks included in each builder. It also prevents someone from easily using a "buildpackless" builder and the language-family buildpack to build.
> 
> What are the benefits of pushing the language-family order configurations into the builders?

I don't propose to replace language-family buildpacks with builders.   I will update the sample builder.toml to reflect that.

text/0000-create-language-family-builders.md
```diff
@@ -0,0 +1,148 @@
...
+
+Another alternative is adding fewer additional JVM provideds to the existing builders.  Each buildpack adds one layer to the image.  If the 8 other JVM providers were included in the Java buildpack the number of layers would increase to 111.  Still within the 127 max but inching closer. Perhaps a smaller set of JVM providers could be added instead of all 8.
+
+A third alternative is to create a JVM meta-buildpack that contains all the JVM provider buildpacks and configures the JRE/JDK dependencies.   This would replace the current BellSoft Liberica buildpack.  
+
+However, for all 3 alternatives, a language-specific builder still reduces the size (number of layers in the builder) and speed (downloading fewer buildpacks) of building applications and provides growing room to add additional buildpacks like the APMs.    
+
+## Implementation
+
+This RFC proposes that the builder team own the language-specific builders.  The current process of updating the Paketo builders is automated using a builder.toml. Adding additional language-specific builder config files in each builder repo therefore has little impact on maintenance cost for the project.
```
dmikusa, line 33: Is this OK with @paketo-buildpacks/builders-maintainers? Is it technically possible to add additional builders this way?

Also, I would kind of like to see this jointly owned. Where the @paketo-buildpacks/builders-maintainers own the machinery/workflows to produce the images, but the language team owns the actual builder.toml & any tests for validation.

Thoughts?

dmikusa: I think we should also make a note here to say something like, basically to make it clear that ownership is delegated to the language family teams. RFCs and decisions about these builders would go through the language family RFC process.

```
The decision to create language family-specific builders is owned by each language family team. Teams may decide to create zero or more based on their needs and user demand. In addition, each language-family team owns the decisions about what goes into the builder. Again, this can be based on team and user needs.
```

kevin-ortega: @dmikusa I have updated the RFC to more clearly propose joint ownership and delegate the creation to each individual language family.

dmikusa: Thanks Kevin. I'm going to resolve this conversation. I've added @paketo-buildpacks/builders-maintainers as a reviewer, so we can get some feedback.

dmikusa (approved):

johnnyr0x:
@ryanmoran to follow up with team to get more input

dmikusa:
@paketo-buildpacks/builders-maintainers - Would you be able to weigh in on this one? Primarily, the ownership is going to be on the language-family team, but language-family teams will use builder tools so there is some shared ownership. Thanks!

robdimsdale (approved):
This looks good to me.

The only thing I'd call out is that OSS support for Bionic will be dropped in April 2023 - both by Paketo but also by Canonical - so by the time we approve and implement this RFC there may not be much value in implementing a Bionic builder for language families. Therefore we may wish to remove it from the RFC. I don't mind if we keep it in the RFC though in the interests of getting it approved sooner.

The ownership detailed in the Implementation section seems fine, with the addition that some of the machinery for the builders is co-maintained with @paketo-buildpacks/tooling-maintainers (i.e. the [`builder` section of the shared github config](https://github.com/paketo-buildpacks/github-config/tree/main/builder))

kevin-ortega:
> This looks good to me.
> 
> The only thing I'd call out is that OSS support for Bionic will be dropped in April 2023 - both by Paketo but also by Canonical - so by the time we approve and implement this RFC there may not be much value in implementing a Bionic builder for language families. Therefore we may wish to remove it from the RFC. I don't mind if we keep it in the RFC though in the interests of getting it approved sooner.
> 
> The ownership detailed in the Implementation section seems fine, with the addition that some of the machinery for the builders is co-maintained with @paketo-buildpacks/tooling-maintainers (i.e. the [`builder` section of the shared github config](https://github.com/paketo-buildpacks/github-config/tree/main/builder))

Good point.  I have updated the RFC and removed bionic examples.

ryanmoran (approved):

sophiewigmore:
Working group 3/15 - waiting on a final review from @ForestEckhardt

robdimsdale:
I think it will be easier for the builders team to maintain if we have separate repos per builder. So one repo for `builder-jammy-full-java`, one for `builder-jammy-tiny-java`, etc. The existing tooling should work fairly well with minimal overhead for that structure. If we want to adopt a monorepo approach we'll have to spend additional time and effort adapting the existing automation to work with that. Not an inherent problem - just pointing it out.

ForestEckhardt:
I am fine with separate repos from a builder maintainer perspective but that does increase the number of repo for participating language families by a quite a bit. Would just want to make sure that we are comfortable with adding that level of overhead.

robdimsdale:
Great, so in terms of naming, the current (jammy) naming convention is as follows: `builder-<OS>-<varient>` e.g. `builder-jammy-full`.

For the language-specific subtypes, I'd propose adding the subtype as a suffix (rather than somewhere else like in the middle or as a prefix). So that would look like: `builder-jammy-full-java` or `builder-jammy-base-java-with-extras`.

kevin-ortega:
> Great, so in terms of naming, the current (jammy) naming convention is as follows: `builder-<OS>-<variant>` e.g. `builder-jammy-full`.
> 
> For the language-specific subtypes, I'd propose adding the subtype as a suffix (rather than somewhere else like in the middle or as a prefix). So that would look like: `builder-jammy-full-java` or `builder-jammy-base-java-with-extras`.

I agree and I will update the RFC with this naming convention.

text/0000-create-language-family-builders.md
```diff
@@ -30,19 +30,24 @@ However, for all 3 alternatives, a language-specific builder still reduces the s
...
 ```
-builder-{distro}-{language-family}:{version}
+builder-{distro}-{variant}-{language-family}:{version}
 ```
 For example,
-* `paketobuildpacks/builder-jammy-java:latest`
-* `paketobuildpacks/builder-jammy-go:latest`
-* `paketobuildpacks/builder-ubi-java:latest`
+* `paketobuildpacks/builder-jammy-full-java:latest`
+* `paketobuildpacks/builder-jammy-full-base:latest`
```
robdimsdale, line 48: I assume this was meant to read: `builder-jammy-base-java`?

kevin-ortega: > I assume this was meant to read: `builder-jammy-base-java`?

Crap, yes.  Thank you. I’ll fix.

robdimsdale (approved):
LGTM

ForestEckhardt (approved):

dmikusa:
Thanks for the comments and votes. I think we have the votes to proceed, so this is the final chance to comment.

If there are no additional comments or objections by EOD Friday, Mar 24th, then we'll go ahead and merge this. Thanks!

PR #277, opened 2023-02-09
Add rfc to allow pre-release dependencies

c0d1ngm0nk3y:
## Summary
Each buildpack has a list of dependencies that can be provided. This list could also contain pre-releases so that the buildpack can provide early access to some dependencies.

## Use Cases
Some dependencies (e.g. `sap-machine`) provide pre-releases for their releases. By allowing the user to use them via buildpacks, the upcoming version could already be validated.

## Checklist

### DIFF
--- a/text/0000-allow-preview-dependencies-minor-and-patch.md
+++ b/text/0000-allow-preview-dependencies-minor-and-patch.md
@@ -1,8 +1,8 @@
-# Allow users to consume pre-release depedencies
+# Allow users to consume pre-release depedencies for an existing major version
 
 ## Summary
 
-Each buildpack has a list of dependencies that can be provided. This list could also contain pre-releases so that the buildpack can provide early access to some dependencies.
+Each buildpack has a list of dependencies that can be provided. This list could also contain pre-releases for existing major versions so that the buildpack can provide early access to some dependencies.
 
 ## Motivation
 
@@ -31,11 +31,20 @@ Some dependencies (e.g. `sap-machine`) provide pre-releases for their releases. 
    - `true`: Use preview version whereever availalbe
    - comma separated list: Use preview version for named dependencies, e.g. `node,npm` or `jvm`
 
+Example:
+
+There are 2 versions of the dependency for java 19:
+
+- `19.0.2` - the latest released one
+- `19.0.3-rc.1` - a pre-release
+
+The user could decide to use the pre-release one for early validation. By default the released one would be chosen.
+
 ## Rationale and Alternatives
 
 - **Do nothing**
 
-  We could stick to the status quo (pre-releases can't be consumed via buildpacks).
+  We could stick to the status quo (pre-releases for an existing major version can't be consumed via buildpacks).
 
 - **Fetch latest preview in buildpack**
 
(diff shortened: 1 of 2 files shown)


dmikusa:
I could be missing something, but I'm not aware of a rule that preventing the inclusion of prerelease dependencies. In fact, I can think of a couple of buildpacks off the top of my head that do include prerelease software. Maven is one (I think Tomee has or had some pre-release deps too). We added support for Maven 4, which is in milestone status.

I think we generally don't just given that Paketo buildpacks target "production workloads" and because it creates extra work to track and merge prereleases which often move quickly, but my $0.02 is that if it benefits the buildpack and users, add them.

I don't think we need an RFC for this, unless others feel strongly about it. Again, just my $0.02 but I would say include them like regular dependencies. If we want to standardize around a "preprelase" metadata flag, I'm fine with that, but in general I think it's up to the buildpack to a.) default to an appropriate version (probably not a prerelease) and b.) users to know which version they are selecting.

c0d1ngm0nk3y:
> I could be missing something...

Fair point. I think we phrased the rfc poorly because for major releases, it already works like you pointed out. The use case we have in mind is rather a pre-release for an existing major.

Example:

The user needs java 19 and there are 2 dependencies available: `19.0.2` and `19.0.3-rc3`. So the use case would be to enable the user to use `19.0.2` in some pipeline, but to have another one using `19.0.3-rc3` to already validate coming changes.

We "only" had java (sap-machine to be more specific) in mind, but there is no real reason to limit it to java.

We started a RFC for that because it would most likely involve some change to central components (e.g. `libpak`). When downloading a component, a pre-release should only be picked if the user explicitly ask to include pre-releases. Otherwise `19.0.2` would be picked in that example.

You are right, with major versions, that is a completely different story. If the user wants java 20, it is clear that a pre-release would be picked.

I will rework the rfc and hopefully clarify things.

dmikusa:
>We started a RFC for that because it would most likely involve some change to central components (e.g. libpak). When downloading a component, a pre-release should only be picked if the user explicitly ask to include pre-releases. Otherwise 19.0.2 would be picked in that example.

This would be something to try. Right now, my understanding is that it uses semver to pick the "latest" version. I think that should exclude prereleases unless opted into. At any rate, it would be interesting to see if changes are required or if we can just lean on semver to do this for us. I'd rather not create more convention and metadata if semver can handle things in a reasonable way.

c0d1ngm0nk3y:
> I'd rather not create more convention and metadata if semver can handle things in a reasonable way.

I totally agree. That is what `semver` is for anyway. But I checked all buildpacks contributing an `OpenJDK` and most of them have some kind of pre-releases, but NONE is using semver to signal this.

When actually putting the dependency into the `buildpack.toml`, it would be easy to add this property since this information is available in github releases or the foojay api.

> I think that should exclude prereleases unless opted into.

I think so, too. I thought of a environment variable to allow pre-releases to be found.

dmikusa:
>That is what semver is for anyway. But I checked all buildpacks contributing an OpenJDK and most of them have some kind of pre-releases, but NONE is using semver to signal this.

I'm not surprised by that. Let's rectify that though and use semver to do this. Even if we're not now, semver should be capable of doing it. So let's fix up pipeline-builder and buildpack.toml's to include prerelease information where appropriate.

As usual, the hardest part will probably be for things that do not comply with semver & having to stuff those versions into a semver compliant version string.

c0d1ngm0nk3y:
We had a closer look and it seems easier to have a separate buildpack to deal with that instead of coming up with a generic solution that works for everyone. It is probably "only" relevant for "OpenJDKs" since they might break their users with minor/patch versions when removing ciphers. But addressing this for the specific buildpack we are using, should be rather easy. Hence I will close this RFC.

PR #278, opened 2023-02-15
Adds RFC for Steering Committee Elections

ryanmoran:
[Readable](https://github.com/paketo-buildpacks/rfcs/blob/steering-committee-elections/text/0054-steering-committee-elections.md)


text/0054-steering-committee-elections.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Detailed Explanation
+
+### Process Changes
+
+#### Membership
+
+The composition of the Steering Committee will align to the specifications,
+with some modifications, outlined in the [Cloud Foundry Foundation
+Charter](https://github.com/cloudfoundry/community/blob/main/governing-board/charter.md)
```
dmikusa, line 36: Should we make a point to reference a specific commit of the CFF charter, rather than the main branch which could change over time? If we point to main, it seems like we'd essentially be accepting CFF governance changes automatically.

text/0054-steering-committee-elections.md
```diff
@@ -0,0 +1,158 @@
...
+   a. During any Steering Committee election, if the Steering Committee
+      membership would exceed this limit even after the natural cycle of Steering
+      Committee seat term expirations, enough Steering Committee members must
+      resign for it to be possible for the election to yield a diverse enough
+      Steering Committee.
+   b. If a Steering Committee election cannot produce a diverse enough Steering
+      Committee, the limit will not apply until the next election cycle. When a
+      change in employment of a Steering Committee member causes the Steering
+      Committee membership to exceed this limit, that Steering Committee member
+      will NOT be required to resign their Steering Committee membership.
```
dmikusa, line 55:
```suggestion
   a. During any Steering Committee election, if the Steering Committee
      membership would exceed this limit even after the natural cycle of Steering
      Committee seat term expirations, enough Steering Committee members must
      resign for it to be possible for the election to yield a diverse enough
      Steering Committee.
   b. If a Steering Committee election cannot produce a diverse enough Steering
      Committee, the limit will not apply until the next election cycle. 
   c. When a change in employment of a Steering Committee member causes the 
      Steering Committee membership to exceed this limit, that Steering Committee
      member will NOT be required to resign their Steering Committee membership.
```

Minor note. Seems like b.) is two things so I split them out.

text/0054-steering-committee-elections.md
```diff
@@ -0,0 +1,158 @@
...
+   in one year and one will be up for election the following year.
+4. Employees from the same company or Related Companies (as defined in CFF
+   Charter Section 9) should not hold more than one Steering Committee seat.
+   a. During any Steering Committee election, if the Steering Committee
+      membership would exceed this limit even after the natural cycle of Steering
+      Committee seat term expirations, enough Steering Committee members must
+      resign for it to be possible for the election to yield a diverse enough
+      Steering Committee.
+   b. If a Steering Committee election cannot produce a diverse enough Steering
+      Committee, the limit will not apply until the next election cycle. When a
```
dmikusa, line 52: I'm trying to understand this point. Would this essentially happen there are not enough candidates to meet the company diversity requirement?

Like let's say we need to fill two seats, and there are only two candidates both from the same company? In that case, both could be put onto the steering committee simply because there is no one else available.

Also, am I understanding this scenario correctly? If we needed to fill two seats and there were three candidates, two candidates from company A and a candidate from company B. Even if the two candidates from company A get the most votes, we'd still select one candidate from company A and one from company B, right? Because there is enough company diversity to meet the governance requirement.

text/0054-steering-committee-elections.md
```diff
@@ -0,0 +1,158 @@
...
+      change in employment of a Steering Committee member causes the Steering
+      Committee membership to exceed this limit, that Steering Committee member
+      will NOT be required to resign their Steering Committee membership.
+5. One Steering Committee seat is reserved for a “community champion”, defined
+   as a person NOT working for a corporation that is a paying member of the
+   Cloud Foundry Foundation.
```
dmikusa, line 58: And would 4.b also apply to this requirement? If no one from the community steps forward, then we'd waive this requirement until the next election.

text/0054-steering-committee-elections.md
```diff
@@ -0,0 +1,158 @@
...
+align to the rules, with some modifications, outlined in the [Cloud Foundry
+Foundation
+Charter](https://github.com/cloudfoundry/community/blob/main/governing-board/charter.md)
+Sections 7.f. Their modified contents are outlined below.
+
+1. In the case where a Steering Committee member resigns or is removed from the
+   Steering Committee by the Governing Board, the remaining members of the
+   Steering Committee will determine whether the vacant seat either will remain
+   vacant until the next annual election cycle, or will be filled by using the
+   results of the latest election.
```
dmikusa, line 133: Is this sufficient for a group of three? The CF TOC has five members, so there is more redundancy.

With a group of three, you could run into issues like:

1. Two resignations around the same time, leaving only one member left.
2. A single resignation, but no agreement on how to proceed (since there's an even number of members left).

I'm also not a fan of how it's arbitrarily left up to the remaining steering committee members to decide how to proceed. 

Perhaps we need something more concrete like "if it's within X months of the previous election, the remaining Steering committee members may fill the vacancy based on the previous election results"?

Perhaps we also need something like "If a resignation results in reducing the Steering Committee to only one member, then the remaining member will work with the CFF Governing Board (or TOC?) to determine a course of action."?

What do folks think?

sophiewigmore:
WG meeting notes:
Folks should think about if they are interested in becoming a steering committee member going forward. Anyone involved in the project should take a look at this RFC.

text/0054-steering-committee-elections.md
ryanmoran: Yes. I read this as the diversity requirement has precedence but would not prevent less diverse candidates to be elected if the candidate pool itself wasn't diverse enough.

text/0054-steering-committee-elections.md
ryanmoran: Yes. We would waive the requirement. I can add a clause stating that.

text/0054-steering-committee-elections.md
ryanmoran: I've modified this to state the the Steering Committe, in whatever capacity it remains, will work with the Governing Board to determine next steps.

text/0054-steering-committee-elections.md
ryanmoran: Done in the latest version.

text/0054-steering-committee-elections.md
ryanmoran: Done in the latest version.

dmikusa (approved):
Looks good to me.

PR #279, opened 2023-03-03
Proposal to create UBI based stacks.

BarDweller:
## Summary
Paketo stacks are currently based on Ubuntu, this RFC proposes creation of UBI based Builder/Run images, also making use of the new Builder Image extension capabilities to provide runtime dependencies via rpm/dnf.

## Use Cases
Customers who wish to use buildpacks, but also require their applications to run on UBI based containers, using UBI supplied runtimes.

## Checklist

### DIFF
--- a/text/0000-ubi-based-stacks.md
+++ b/text/0000-ubi-based-stacks.md
@@ -40,7 +40,7 @@ same mechanism).
 
 Via extensions, it becomes possible to have a Builder image that can install dependencies such as Java, Maven,
 NodeJS, etc via yum. These extensions are very similar to buildpacks, participating in the 'detect' phase, and 
-being allowed to issue 'provides' to the buildplan (but critically not 'requires').. then extensions that are
+being allowed to issue 'provides' to the buildplan (but critically not 'requires'). Then extensions that are
 selected to become part of a build run a 'generate' task (rather than a buildpacks 'build' task), that may output 
 Dockerfiles to modify the Builder image, and/or substitute the Run image. Just like buildpacks, the detect & 
 generate invocations get access to the source project to determine what actions they wish to take. For more 
--- a/text/0000-ubi-based-stacks.md
+++ b/text/0000-ubi-based-stacks.md
@@ -71,15 +71,15 @@ subsequent images are published.
 The stack will name and tag the release images with the following pattern:
 
 ```
-paketobuildpacks/{phase}-ubi8-{variant}:{version}
+paketocommunity/{phase}-ubi8-{variant}:{version}
 ```
 
 For example we could see the following images for UBI8 stacks:
 
-* `paketobuildpacks/build-ubi8-base:latest`
-* `paketobuildpacks/run-default-ubi8-base:1.2.3`
-* `paketobuildpacks/run-java-ubi8-base:1.2.3`
-* `paketobuildpacks/run-nodejs-ubi8-base:1.2.3`
+* `paketocommunity/build-ubi8-base:latest`
+* `paketocommunity/run-default-ubi8-base:1.2.3`
+* `paketocommunity/run-java-ubi8-base:1.2.3`
+* `paketocommunity/run-nodejs-ubi8-base:1.2.3`
 
 *Note the run image `phase` is narrowed by inclusion of the pre-packaged runtime. The correct image is 
 selected at Build time by the appropriate Extension, if no Extensions select an image, then the
--- a/text/0000-ubi-based-stacks.md
+++ b/text/0000-ubi-based-stacks.md
@@ -152,7 +152,7 @@ The buildpacks team will create two new repos for the initial two Extensions.
 * `ubi-java-extension`
 * `ubi-nodejs-extension`
 
-The *(unknown)* team will create the repo for the Builder image
+The builders-maintainers team will create the repo for the Builder image
 
 * `builder-ubi-base`
 
--- a/text/0000-ubi-based-stacks.md
+++ b/text/0000-ubi-based-stacks.md
@@ -77,11 +77,13 @@ paketobuildpacks/{phase}-ubi8-{variant}:{version}
 For example we could see the following images for UBI8 stacks:
 
 * `paketobuildpacks/build-ubi8-base:latest`
+* `paketobuildpacks/run-default-ubi8-base:1.2.3`
 * `paketobuildpacks/run-java-ubi8-base:1.2.3`
 * `paketobuildpacks/run-nodejs-ubi8-base:1.2.3`
 
 *Note the run image `phase` is narrowed by inclusion of the pre-packaged runtime. The correct image is 
-selected at Build time by the appropriate Extension*
+selected at Build time by the appropriate Extension, if no Extensions select an image, then the
+'default' image will be used instead*
 
 Each stack repository should include a README that outlines the stacks that are
 available including links to each other repository allowing users to discover
@@ -186,8 +188,3 @@ run images for a given builder image. This will likely require minor updates to 
 
 * It may be possible in future builders to support run image customization via rpm/yum, at the cost of 
 losing rebasing support with the final application image. 
-
-* How does the actual building & releasing of the stack image work? is this something configured onto the 
-repo by the 'stacks team' on our behalf? 
-
-* I'm guessing here at the teams that do things.. feel free to jump in and tell me where things actually get done =)
--- a/text/0000-ubi-based-stacks.md
+++ b/text/0000-ubi-based-stacks.md
@@ -1,0 +1,181 @@
+# Stacks & Extensions for UBI base images. (UBI8)
+
+## Summary
+
+A set of stacks based on the UBI base image should be developed, released, and maintained. 
+
+Unlike the existing stacks, that come in "full", "base", and "tiny" the UBI stack will initially start with a 
+single variant, with future variants to follow. The initial variant will have buildpacks for a handpicked set of 
+runtimes that have been tested with UBI, and will make use of a number of CNCF Buildpack 'Extensions' that will be 
+developed, released and maintained by the relevant teams to offer installation of selected runtimes and dependencies 
+via rpm/dnf. These extensions will be packaged with the UBI builder to allow the customer the option of using the 
+UBI provided runtimes. 
+
+## Motivation
+
+Red Hat Universal Base Images (UBI) are designed to be a foundation for cloud-native and web applications use cases 
+developed in containers. UBI Images form a popular choice for base container images, and can even be preferred 
+within environments with existing UBI deployments. In addition, use of a UBI image can sometimes be a requirement 
+for an application qualifying for support from some vendors while in production. That requirement is also extended 
+to any runtime in use, eg for Node or Java, the runtime must be installed via rpm/dnf, rather than via a tarball. 
+
+## Detailed Explanation
+
+### Background
+
+There are a wide variety of UBI images offered by Red Hat, offering pre configured deployments of various runtimes 
+(such as Java), and offering cutdown images designed for extension (ubi-minimal).
+
+While it would be possible to use UBI images in a similar fashion to the existing Ubuntu based images, and have 
+buildpacks perform the download & install of depdencies within layers, the preferred route would be to to use the 
+dependencies supplied via the configured package repositories within the images. Eg, rather than downloading a 
+JVM/JDK, one would be installed via 'yum' or 'dnf'. 
+
+Until recently, CNCF Buildpacks were unable to effect changes to the Builder or Run image at a system level (eg, you
+could not make modifications that required root from a buildpack, such as installing packages via yum). Recent 
+updates to the Buildpacks project have added 'extensions', these can be considered as special variants of buildpacks 
+that are able to modify the Builder image (and substitute the Run image), via generated Dockerfiles created during 
+the execution of the Extensions. (In the future, the goal is to also allow modification of the Run image via the 
+same mechanism).
+
+Via extensions, it becomes possible to have a Builder image that can install dependencies such as Java, Maven,
+NodeJS, etc via yum. These extensions are very similar to buildpacks, participating in the 'detect' phase, and 
+being allowed to issue 'provides' to the buildplan (but critically not 'requires').. then extensions that are
+selected to become part of a build run a 'generate' task (rather than a buildpacks 'build' task), that may output 
+Dockerfiles to modify the Builder image, and/or substitute the Run image. Just like buildpacks, the detect & 
+generate invocations get access to the source project to determine what actions they wish to take. For more 
+information on extensions see.. https://github.com/buildpacks/spec/blob/main/image_extension.md
+
+### Initial offered image
+
+The intent is provide an initial builder image that caters to Java & NodeJS runtimes (without Java Native). 
+Some early prototyping has shown this to be possible using the existing Java & NodeJS builder ecosystems 
+present within Paketo today, with Extensions installing the main runtime dependencies. 
+
+In the case of Java, much work has already been done by Paketo to allow for different buildpacks to supply the 
+JDK/JVM, making it relatively simple to extend this to allow an Extension to perform the same role. For NodeJS, there
+were some small changes required, that have already been merged. 
+
+The initial image will be labelled with a Stack ID of `io.buildpacks.stacks.ubi8` Note that much as Ubuntu
+moves from release to release, so does UBI, and currently we will target UBI8, but want to ensure the naming allows
+for the future addition for UBI9 and so on.
+
+There will be no "intermediate" image as with Bionic, as the base builder image is intended to be a published
+UBI8 image.
+
+The initial image shall be considered as the 'base' variant, for the purposes of differentiation when/if 
+subsequent images are published. 
+
+### Image Naming and Tagging
+
+The stack will name and tag the release images with the following pattern:
+
+```
+paketobuildpacks/{phase}-ubi8-{variant}:{version}
+```
+
+For example we could see the following images for UBI8 stacks:
+
+* `paketobuildpacks/build-ubi8-base:latest`
+* `paketobuildpacks/run-java-ubi8-base:1.2.3`
+* `paketobuildpacks/run-nodejs-ubi8-base:1.2.3`
+
+*Note the run image `phase` is narrowed by inclusion of the pre-packaged runtime. The correct image is 
+selected at Build time by the appropriate Extension*
+
+Each stack repository should include a README that outlines the stacks that are
+available including links to each other repository allowing users to discover
+the stack variants available from any of the stack repository pages.
+
+### Subsequent images..
+
+It is likely that Native compilation for Java may require a different Builder image, as the requirements for 
+native compilation are complex, and may require the Builder to extend an appropriate base image to provide 
+the require capabilities in a supported manner. 
+
+As additional runtimes are added to the initial image over time, it may make sense to follow the same 
+general approach as the Ubuntu stacks have, and aim for a small but manageable number of stacks that meet definable 
+use cases. (eg, 'one size fits all', 'just the most used runtimes')
+
+
+## Rationale and Alternatives
+
+With the goal in mind of creating a Stack based on UBI using system packaged depdencies, these alternatives 
+have been considered...
+
+#### Builders per Runtime
+This option would have 1 Builder image per runtime, paired with a matching run image per runtime, with the runtime
+depdencies baked into the builder/run images. This option did not require Extensions from upstream, but would 
+result in a large collection of single purpose stacks which rapidly becomes unmanageable as you consider versioning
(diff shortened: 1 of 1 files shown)


text/0000-ubi-based-stacks.md
```diff
@@ -0,0 +1,181 @@
...
+Until recently, CNCF Buildpacks were unable to effect changes to the Builder or Run image at a system level (eg, you
+could not make modifications that required root from a buildpack, such as installing packages via yum). Recent 
+updates to the Buildpacks project have added 'extensions', these can be considered as special variants of buildpacks 
+that are able to modify the Builder image (and substitute the Run image), via generated Dockerfiles created during 
+the execution of the Extensions. (In the future, the goal is to also allow modification of the Run image via the 
+same mechanism).
+
+Via extensions, it becomes possible to have a Builder image that can install dependencies such as Java, Maven,
+NodeJS, etc via yum. These extensions are very similar to buildpacks, participating in the 'detect' phase, and 
+being allowed to issue 'provides' to the buildplan (but critically not 'requires').. then extensions that are
```
dmikusa, line 43:
```suggestion
being allowed to issue 'provides' to the buildplan (but critically not 'requires'). Then extensions that are
```

text/0000-ubi-based-stacks.md
```diff
@@ -0,0 +1,181 @@
...
+use of dependencies downloaded via existing buildpacks. The rpm/yum behavior will be the default for the ubi stack. 
+
+* Note that multi-runtime applications (where say both nodejs and java form the final application) are not 
+covered by this stack. Although still achievable via end-user custom builders, it's not a goal for this incarnation. 
+
+* To allow for rebasing, and to work with the current Extension support, it is required to publish **multiple** 
+run images for a given builder image. This will likely require minor updates to the existing stack creation tooling. 
+
+* It may be possible in future builders to support run image customization via rpm/yum, at the cost of 
+losing rebasing support with the final application image.
```
dmikusa, line 176: These two bullets are my primary concerns.

1. The proliferation of run images and how much overhead that creates for the project. 

    @BarDweller Do you have any thoughts on ultimately how run images there will be? If not, is it a capped number or open-ended? 

    I'd also like to know what @paketo-buildpacks/stacks-maintainers think about the amount of overhead/toil this might create. Does it seem like it would be feasible to handle? Do we need more people-power to keep up? Do we perhaps need to look at adjustments to the tooling for efficiency purposes before or as part of accepting this RFC?

2. I'm concerned with the fact that upstream isn't done working on this feature and if we're rushing into something. 

    If we implement this now, we have to support what we're doing going forward for a reasonable amount of time. We can't just pivot and start doing something else, we'll be on the hook for support and publishing new run images for at least a year, if not longer. 

    So if in 6 months or a year, the buildpacks project finishes its work and run images support customization on-the-fly, what becomes of this implementation? Is it still the **best** implementation for our users?

text/0000-ubi-based-stacks.md
```diff
@@ -0,0 +1,181 @@
...
+information on extensions see.. https://github.com/buildpacks/spec/blob/main/image_extension.md
+
+### Initial offered image
+
+The intent is provide an initial builder image that caters to Java & NodeJS runtimes (without Java Native). 
+Some early prototyping has shown this to be possible using the existing Java & NodeJS builder ecosystems 
+present within Paketo today, with Extensions installing the main runtime dependencies. 
+
+In the case of Java, much work has already been done by Paketo to allow for different buildpacks to supply the 
+JDK/JVM, making it relatively simple to extend this to allow an Extension to perform the same role. For NodeJS, there
```
dmikusa, line 56: Are any changes to the Java buildpack required to support this? Or will it Just Work™?

BarDweller: Tested and working within local prototypes with no changes required so far =). 

The flow is basically.. extensions & buildpacks run their detects, both extension-java and buildpack-supplying-java are selected.. (both also claim to 'provides' jdk & jre) the extension runs first, installs java via rpm, then when the buildpack-supplying-java executes the build phase it notices the buildplan no longer requires jdk/jre and becomes a no-op. 
 
It's also possible to just omit the jdk/jre supplying buildpacks from the ubi8 builder image, but I'm trying here to leave as much of the existing meta buildpack content as-is as possible. 
 
 For what it's worth, Node did need a minor change to give it the same behavior, as it wasn't correctly handling the removal of the requires from the buildplan. I believe that change has already been contributed & released.

text/0000-ubi-based-stacks.md
```diff
@@ -0,0 +1,181 @@
...
+were some small changes required, that have already been merged. 
+
+The initial image will be labelled with a Stack ID of `io.buildpacks.stacks.ubi8` Note that much as Ubuntu
+moves from release to release, so does UBI, and currently we will target UBI8, but want to ensure the naming allows
+for the future addition for UBI9 and so on.
+
+There will be no "intermediate" image as with Bionic, as the base builder image is intended to be a published
```
sophiewigmore, line 63: Could you clarify what you mean by no "intermediate" image?

text/0000-ubi-based-stacks.md
```diff
@@ -0,0 +1,181 @@
...
+
+```
+paketobuildpacks/{phase}-ubi8-{variant}:{version}
+```
+
+For example we could see the following images for UBI8 stacks:
+
+* `paketobuildpacks/build-ubi8-base:latest`
+* `paketobuildpacks/run-java-ubi8-base:1.2.3`
+* `paketobuildpacks/run-nodejs-ubi8-base:1.2.3`
```
sophiewigmore, line 81: I think I'm misunderstanding - what are the runtime-specific run images needed for, if the extensions will be providing the runtime installations?

text/0000-ubi-based-stacks.md
BarDweller: This RFC was based in part on the Jammy Stacks RFC, which included this statement, when looking at how Jammy built its stack it felt like that was in keeping with how the UBI stack would be built, so I included this statement here. I'll confess to not knowing what Bionic did with an intermediate image, or why it needed one, but thought that adding the statement would help people who did, to know that UBI is closer to Jammy in approach, than to Bionic =)

text/0000-ubi-based-stacks.md
BarDweller: Because at least in this incarnation, the extensions are only modifying the builder image, not the run image. At present upstream only offers support for builder image modification and run image switching. This is enough to create a viable solution, but does have the downside of requiring pre-canned runtime images. 

Once upstream offers run image modification, then we can offer a solution based around that, but run image modification comes with the cost that it will disable rebasing for any applications that are built using it. 

Run image switching doesn't come with that drawback, so in some ways provides the end user with a better experience at the cost of requiring a greater number of run images. 

The base for each runtime run image already exists as a  published UBI runtime image, except each is lacking the required env vars and labels for it to qualify as a buildpack run image. We are looking at if we can get those added to the upstream images, which would remove the need for Paketo to publish them, but I didn't want to rely on that happening for this RFC, so included the info on Paketo publishing the run images.

sophiewigmore: @barDweller thank you for the added explanation, that cleared up my confusion. 👍

text/0000-ubi-based-stacks.md
dmikusa: That's great. Thanks for digging into that and clarifying.

text/0000-ubi-based-stacks.md
dmikusa: >That much said, I don't expect any great surprises from the way things are going upstream, if run-image extension becomes something that would offer customers value, it's not mutually exclusive within the same builder, there would just have to be env style config to select if the app image should use a precanned run image or allow the building of one during app build. (the latter coming with the cost of disabling rebasing).

I think you answered my next question too. I am wondering if you'd still suggest going ahead with the precanned run image even if, hypothetically, upstream had the dynamic run image extension support available today. It sounds like, yes, you still would. For users where rebasing is important.

It's more work to do both, but the idea of making that something a user could pick is interesting too. Precanned run images vs dynamic.

text/0000-ubi-based-stacks.md
```diff
@@ -0,0 +1,181 @@
...
+every update. A large image + frequent updates creates a less than ideal development experience.
+
+#### UBI without rpms
+This option would just take the existing Paketo buildpacks as-is and run them on a UBI base image. As with all the 
+options here, testing would be required for all contained buildpacks to ensure any system or shell interactions were 
+unaffected by the Ubuntu->UBI switch. 
+
+Although this approach would produce functional application images, the resulting images would not be using UBI 
+installed runtimes, and thus could be unable to qualify for full support in production. This is considered important 
+enough to end users that this option was discarded.
```
dmikusa, line 135: Just curious, but what if you used the buildpacks to perform the installation but they installed UBI runtime files? Hypothetically speaking, what if the buildpack installed the files from a UBI RPM into a UBI image, just in a different location (since the buildpacks can only write to `/layers`)?

I'm not really proposing this as an alternative, but I'm curious about what is required for a user to be officially supported with UBI. Thanks in advance.

text/0000-ubi-based-stacks.md
```diff
@@ -0,0 +1,181 @@
...
+### Repositories
+
+The stacks team will create a new repo for UBI.. 
+
+* `ubi-base-stack`
+
+The buildpacks team will create two new repos for the initial two Extensions.
+
+* `ubi-java-extension`
+* `ubi-nodejs-extension`
```
dmikusa, line 153: Are you proposing these be owned by each language family team? i.e. Java buildpacks team owns `ubi-java-extension` and `ubi-nodejs-extension` is owned by the Node.js team?

Do you have a link to these you can share? I'm not sure if it's part of this RFC, possibly another one, or just an implementation detail, but these are going to be the first extensions we support so I think it would be worth digging into the way these are set up, what tooling is being used, and what sorts of patterns we're trying to establish for future extensions.

text/0000-ubi-based-stacks.md
```diff
@@ -0,0 +1,181 @@
...
+
+The *(unknown)* team will create the repo for the Builder image
```
dmikusa, line 155: Probably @paketo-buildpacks/builders-maintainers ?

text/0000-ubi-based-stacks.md
```diff
@@ -0,0 +1,181 @@
...
+
+There will be no "intermediate" image as with Bionic, as the base builder image is intended to be a published
+UBI8 image.
+
+The initial image shall be considered as the 'base' variant, for the purposes of differentiation when/if 
+subsequent images are published. 
+
+### Image Naming and Tagging
+
+The stack will name and tag the release images with the following pattern:
```
dmikusa, line 71: I don't think it would necessarily impact the naming of stacks, but is it a goal to have multi-architecture images available? Maybe put another way, will these stacks support both AMD64 and ARM64?

text/0000-ubi-based-stacks.md
BarDweller: Absolutely, the precanned enables rebasing, and I think that may be important enough to offer precanned for hand picked selection of runtimes.. for everything else, eventually, run-image customization will be 'enough' to start with, and if we put in the right sort of override, then an end user could force their own pre-canned run-image with the appropriate pre-installed customizations. 

Given where we're starting from though, it feels safe to have an initial plan to do precanned, and evolve it as upstream and requirements change =)

text/0000-ubi-based-stacks.md
BarDweller: Heh.. it's not entirely hypothetical, we looked pretty extensively into if we could install files from various runtime RPMs into `/layers` .. but sadly it's not that simple, as rpm's are intended to install to the system, and we discovered hardcoded path dependencies within some that prevented their install to different locations. From a technical perspective, although this could be corrected, you'd basically be expanding the support offering to allow the installation to non-default paths.. that would likely come with a pile of required testing etc and isn't something that's easy to justify just for buildpacks. 

From my understanding of "supported" you generally today use the precanned ubi8 runtime images, where it's all already installed.. or it's also acceptable to install via rpm. However, I'm just a developer, and my answer shouldn't be taken as a legal response ;) I suspect the actual definition of supported is thrashed out by any support contract by lawyer type people who are far more precise in their definition =)

text/0000-ubi-based-stacks.md
BarDweller: I think multi-arch makes sense, I don't specifically see any complications with that from a ubi perspective, provided ubi images exist for the arch's being considered, because the deps are installed via rpm, so that's just a shell interaction, the appropriate rpm repo for the arch would already be configured in the base image, so I think if there any multi arch concerns for Paketo UBI stacks, they are existing ones shared by Paketo already (eg, the Go detect/build binaries needing to be executable etc).

text/0000-ubi-based-stacks.md
sophiewigmore: To your point Dan, I'm hesitant about adding specifics about the particular tooling and patterns in this RFC. My biggest concern is that we'd pigeonhole ourselves into a specific implementation across the board, which feels especially unnecessary if each extension will be maintained by separate language maintainer groups. It doesn't hurt to align on these things, but I'm not sure if declaring these things at the RFC level is too much.

BarDweller: I should add that just because it makes sense for these two extensions to be owned by their respective language groups doesn't mean that will be true for all extensions.. eg an 'update-ca-certificates' extension (that updates the system certs in the builder image during build to include self-signed certs required for local resources).. wouldn't have any affinity to either java, or node, or any specific runtime. 

It's also maybe worth pointing out that it's doable to have multiple extensions for a given language group, eg Java could have a ubi-maven-extension that 'supplied' maven via rpm (in practice this isn't required because the maven buildpack can defer to an existing on-path maven, but you get the idea). 

The only critical thing about use of extensions (enforced by the human who selects which extensions are packaged into the builder image) is that they don't end up fighting over which run image they want to switch to. This is fine with a hypothetical ca-certificates extension that doesn't want to touch the run image, and just needs to fix up the build env, but less good if say node & java both decide to kick in for the same project. (we're aware of the recent java/node changes with buildpacks)

dmikusa: @sophiewigmore Yes, I agree. I don't want to make the RFC more complicated or overly specific. I do have concerns with having consistent implementations though, not at a fine-grained level, but in a larger sense. In short, I want to try to avoid another libpak vs packit situation.

text/0000-ubi-based-stacks.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Background
+
+There are a wide variety of UBI images offered by Red Hat, offering pre configured deployments of various runtimes 
+(such as Java), and offering cutdown images designed for extension (ubi-minimal).
+
+While it would be possible to use UBI images in a similar fashion to the existing Ubuntu based images, and have 
+buildpacks perform the download & install of depdencies within layers, the preferred route would be to to use the 
+dependencies supplied via the configured package repositories within the images. Eg, rather than downloading a 
+JVM/JDK, one would be installed via 'yum' or 'dnf'.
```
ForestEckhardt, line 32: I am not sure that I fully understand why installation from a configured package this is preferable method? Could you layout what it is superior to using the CNB layer system that we are currently using?

BarDweller: I'm not sure it's that simple, to be able to say one is superior to the other.. The tarball+layers dir approach is perfectly capable of creating a working functional application, and so is the route where the runtime is installed via system packages.

I explain the motivation behind wishing to use system based deps within the section "UBI without rpms" a little further on in the RFC, but in essence, there's a support driven motivation here, that in order to be totally supported, a customer on UBI needs to also be using a UBI installed runtime (eg, java, or node, which are the ones being considered here). 

As we consider this to be one of the drivers for for UBI based apps, that's why having these runtimes installed in this manner becomes the default for the UBI stack. That's not to say it can't be overridden, but out of the box, this builder aims to create images that are fully eligible for support. (Or at least as far as they can be within the context of the build & run components managed by Paketo).

text/0000-ubi-based-stacks.md
BarDweller: Added note =)

text/0000-ubi-based-stacks.md
ForestEckhardt: > I explain the motivation behind wishing to use system based deps within the section "UBI without rpms" a little further on in the RFC, but in essence, there's a support driven motivation here, that in order to be totally supported, a customer on UBI needs to also be using a UBI installed runtime (eg, java, or node, which are the ones being considered here).

Gotcha! I just went back and looked over that section again and I think that makes sense. I guess I would be curious to see how this work and the understand more the drawbacks of using `Extensions` because I think that having two different types of buildpack like resource for installing dependencies could get confusing and I would love to potentially consolidate on one or the other. I don't think however answering that question should block this RFC so I think that everything else from my perspective as a builder maintainer makes sense!

ForestEckhardt (approved):

text/0000-ubi-based-stacks.md
```diff
@@ -0,0 +1,193 @@
...
+
+The stack will name and tag the release images with the following pattern:
+
+```
+paketobuildpacks/{phase}-ubi8-{variant}:{version}
+```
+
+For example we could see the following images for UBI8 stacks:
+
+* `paketobuildpacks/build-ubi8-base:latest`
```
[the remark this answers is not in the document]
BarDweller: In testing, we've been using ubi minimal as a 'default' run image, in practice, it always ends up being deselected when an extension does the switch. But it may be used if the extensions are disabled (say via an env var, to enable runtimes to be installed via tarballs).. I will update to add a runtime-less base image for the purposes of allowing this scenario =)

robdimsdale (approved):

robdimsdale:
I am curious to know what we think about automation for the new repositories (to create and publish the stacks, builders, and extensions) - will this be completely new automation or will it be a modification of what already exists in https://github.com/paketo-buildpacks/github-config

But, I don't think that implementation needs to be called out in this RFC, so I didn't add it as a comment.

sophiewigmore (approved):

mhdawson:
@robdimsdale my goal at least is that we re-use as much as we can from existing repositories in terms of automation and approach. You can see the work in progress Node.js extension here - https://github.com/nodeshift/ubi-nodejs-extension and you should see that it follows the same pattern as the other Node.js build packs and we are trying to leverage as much of the existing approach/automation as possible.

robdimsdale:
> my goal at least is that we re-use as much as we can from existing repositories in terms of automation and approach. You can see the work in progress Node.js extension here - https://github.com/nodeshift/ubi-nodejs-extension and you should see that it follows the same pattern as the other Node.js build packs and we are trying to leverage as much of the existing approach/automation as possible.

Yup - I saw that, thanks. It sounds like we're aligned in terms of looking for opportunities to reuse and extend existing automation, which is great.

text/0000-ubi-based-stacks.md
```diff
@@ -0,0 +1,190 @@
...
+
+The initial image shall be considered as the 'base' variant, for the purposes of differentiation when/if 
+subsequent images are published. 
+
+### Image Naming and Tagging
+
+The stack will name and tag the release images with the following pattern:
+
+```
+paketobuildpacks/{phase}-ubi8-{variant}:{version}
```
dmikusa, line 74:
```suggestion
paketocommunity/{phase}-ubi8-{variant}:{version}
```

Because we're starting this as a Paketo Community project, all of the images should be `paketocommunity/*`. The creds/tooling is going to publish under https://hub.docker.com/u/paketocommunity.

text/0000-ubi-based-stacks.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+For example we could see the following images for UBI8 stacks:
+
+* `paketobuildpacks/build-ubi8-base:latest`
```
dmikusa, line 79:
```suggestion
* `paketocommunity/build-ubi8-base:latest`
```

text/0000-ubi-based-stacks.md
```diff
@@ -0,0 +1,190 @@
...
+* `paketobuildpacks/run-default-ubi8-base:1.2.3`
+* `paketobuildpacks/run-java-ubi8-base:1.2.3`
+* `paketobuildpacks/run-nodejs-ubi8-base:1.2.3`
```
dmikusa, line 82:
```suggestion
* `paketocommunity/run-default-ubi8-base:1.2.3`
* `paketocommunity/run-java-ubi8-base:1.2.3`
* `paketocommunity/run-nodejs-ubi8-base:1.2.3`
```

dmikusa (changes requested):

robdimsdale (approved):

dmikusa (approved):

PR #288, opened 2023-04-26
Proposal to publish multi-arch buildpacks

jpena-r7:
[Readable](https://github.com/jpena-r7/rfcs/blob/manifest-list-buildpacks/text/0000-manifest-list-buildpacks.md)

## Checklist

### DIFF
--- a/text/0000-manifest-list-buildpacks.md
+++ b/text/0000-manifest-list-buildpacks.md
@@ -1,0 +1,133 @@
+# Publish multi-arch buildpacks
+
+## Summary
+
+In order to enable end users to create arm64 images with paketo buildpacks I am proposing that paketo start publishing multi-arch buildpack images for all buildpacks. @dmikusa put together [this](https://github.com/dmikusa/paketo-arm64) excellent guide for building arm64 containers with paketo buildpacks, but it puts a heavy burden on end users to create and maintain copies of paketo buildpacks. This proposal is to start publishing manifest list images that support x86_64 and arm64 architectures. This would be accomplished by using `docker manifest` or similar.
+
+For simplicity I will refer to x86_64 as amd64 throughout this document.
+
+## Motivation
+
+The number one question I get about [paketo] buildpacks is whether they support arm. It is safe to say that workloads are generally moving to arm processors and end users are already evaluating how to build containers on arm. It is difficult for me as an end user to seriously consider building arm images with paketo buildpacks without putting in some engineering effort into building arm versions of the paketo buildpacks I may need. While the above example is definitely helpful, a lot of the steps can be built into the github actions workflows that create and release buildpacks for general consumption. I believe arm support was the number one requested roadmap item for 2023, and this proposal is an attempt to jump start that process.
+
+
+## Detailed Explanation
+
+Right now binaries are created on/for amd64 and then packaged up into buildpacks using pack. It should be easy to create binaries for arm64 which can then be packaged up into buildpacks using pack. These architecture-specific buildpacks can be pushed to a registry with a platform specific tag as shown in the examples below. 
+
+* `paketobuildpacks/ca-certificates:3.6-amd64`
+* `paketobuildpacks/ca-certificates:3.6-arm64`
+
+The last step would be to create a manifest list image using `docker manifest` commands (or similar) and push it to the registry.
+
+Users would pull the manifest list image such as: `paketobuildpacks/ca-certificates:3.6`
+
+This is what the manifest might look like for `paketobuildpacks/ca-certificates:3.6` (as an example)
+
+```
+# crane manifest paketobuildpacks/ca-certificates:3.6 | jq .
+{
+  "schemaVersion": 2,
+  "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
+  "manifests": [
+    {
+      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
+      "size": 12345,
+      "digest": "sha256:abcdefchijklmnodka0204kdkbjladkj02jfjbe2458801ekdbbks024f555kdkd",
+      "platform": {
+        "architecture": "amd64",
+        "os": "linux"
+      }
+    },
+    {
+      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
+      "size": 12345,
+      "digest": "sha256:3333defchijklmnodka0204kdkbjladkj02jfjbe2458801ekdbbks024f555kdkd",
+      "platform": {
+        "architecture": "arm64",
+        "os": "linux"
+      }
+    }
+  ]
+}
+# 
+```
+
+## Rationale and Alternatives
+
+The upstream pack project is already creating multi-arch images for pack and the lifecycle using `docker buildx`, which creates manifest list images. An alternative approach is to create and publish separate images that are tagged by architecture. This would be the steps I mentioned above without the final manifest list image. I think this is an anti-pattern at this point given that multi-arch images are already a norm. I also suspect end users would end up creating their own manifest list images anyway. I think the biggest driver for this is to avoid end users cloning and rebuilding paketo buildpacks.
+
+## Implementation
+
+### Changes to packit
+
+`Os` and `Architecture` string fields can be added to the [`Dependency`](https://github.com/paketo-buildpacks/packit/blob/v2/postal/buildpack.go#L15) struct in `packit/postal` to allow buildpacks to interrogate and use the new metadata when determining what external dependencies to download. This will allow `buildpack.toml` files to be updated to add in the `os` and `architecture` metadata for all dependencies. They would all be set to something like `os=linux` and `architecture=amd64` initially.
+
+Some logic can be added to the [Resolve](https://github.com/paketo-buildpacks/packit/blob/18202009038b0df285ba0fb7d8b43abbf60d3ed0/postal/service.go#L89) method of `packit/postal` to pick a version that matches the `GOOS` environment variable whenever the `os` and `architecture` fields have been specified.
+
+### Changes to `buildpack.toml` files
+
+There are primarily two types of buildpacks: those that run without any external dependencies such as [ca-certificates](https://github.com/paketo-buildpacks/ca-certificates), and those that download and install external dependencies such as [azul-zulu](https://github.com/paketo-buildpacks/azul-zulu). For the former no changes should be needed other than compiling for amd64 and arm64. The later will require some changes to `buildpack.toml`.
+
+The spec for [`buildpack.toml`](https://github.com/buildpacks/spec/blob/main/buildpack.md) makes no mention of `os` or `architecture`, which means the adding those fields to entries in the `metadata.dependencies` list seems like a perfectly reasonable option. [This](https://github.com/paketo-buildpacks/rfcs/blob/decouple-dependencies/text/0000-decouple-dependencies.md) rfc proposes adding `os` and `arch` fields to the versions in the dependency table. I suggest sticking with the same field names that are used in the manifest image (`os` and `architecture`) for consistency, but the general idea is the same.
+
+Here is an example of what the dependencies would look like.
+
+```toml
+  [[metadata.dependencies]]
+    cpe = "cpe:2.3:a:golang:go:1.18.4:*:*:*:*:*:*:*"
+    id = "go"
+    licenses = ["BSD-3-Clause"]
+    name = "Go"
+    stacks = ["*"]
+    strip-components = 1
+    version = "1.18.4"
+    os = "linux"
+    architecture = "amd64"
+    purl = "pkg:generic/go@go1.19?checksum=c9b099b68d93f5c5c8a8844a89f8db07eaa58270e3a1e01804f17f4cf8df02f5&download_url=https://go.dev/dl/go1.18.4.linux-amd64.tar.gz"
+    sha256 = "c9b099b68d93f5c5c8a8844a89f8db07eaa58270e3a1e01804f17f4cf8df02f5"
+    source = "https://go.dev/dl/go1.18.4.linux-amd64.tar.gz"
+    source_sha256 = "c9b099b68d93f5c5c8a8844a89f8db07eaa58270e3a1e01804f17f4cf8df02f5"
+    uri = "https://go.dev/dl/go1.18.4.linux-amd64.tar.gz"
+
+
+  [[metadata.dependencies]]
+    cpe = "cpe:2.3:a:golang:go:1.18.4:*:*:*:*:*:*:*"
+    id = "go"
+    licenses = ["BSD-3-Clause"]
+    name = "Go"
+    stacks = ["*"]
+    strip-components = 1
+    version = "1.18.4"
+    os = "linux"
+    architecture = "arm64"
+    purl = "pkg:generic/go@go1.19?checksum=35014d92b50d97da41dade965df7ebeb9a715da600206aa59ce1b2d05527421f&download_url=https://go.dev/dl/go1.18.4.linux-arm64.tar.gz"
+    sha256 = "35014d92b50d97da41dade965df7ebeb9a715da600206aa59ce1b2d05527421f"
+    source = "https://go.dev/dl/go1.18.4.linux-arm64.tar.gz"
+    source_sha256 = "35014d92b50d97da41dade965df7ebeb9a715da600206aa59ce1b2d05527421f"
+    uri = "https://go.dev/dl/go1.18.4.linux-arm64.tar.gz"
+```
+
+### Github actions and workflows
+
+I haven't looked at github workflows for all buildpacks, but I believe [jam](https://github.com/paketo-buildpacks/jam) is being used to package buildpacks. I did see skopeo being used to push images ([cpython](https://github.com/paketo-buildpacks/cpython/blob/main/.github/workflows/push-buildpackage.yml#L62)).
+
+My proposal is to update jam to create buildpacks for amd64 and arm64. I haven't used jam yet, so I don't have specific code changes for it and would appreciate thoughts on how this could be added.
+
+These can then be pushed using skopeo with the above mentioned architecture-specific tags. Finally some `docker manifest` commands can be used to create the manifest list image and then push it to the registry.
+
+Users would have the same experience using the manifest list images on amd64, while allowing those who want/need arm64 images to benefit as well.
+
+
+## Prior Art
+
+As mentioned above @dmikusa put together [this](https://github.com/dmikusa/paketo-arm64) excellent guide for building arm64 images with paketo buildpacks. This repo, based on the aforementioned example, is publishing arm version of paketo builders and buildpacks. While it is great to see community involvement in pushing this forward, I think it is time for the paketo maintainers to agree on a path forward so the community can start contributing towards that goal.
+
+
+## Unresolved Questions and Bikeshedding
+
+### Testing on arm64
+
+As discussed [here](https://github.com/actions/runner-images/issues/5631), github actions currently only supports amd64 runners. This means that unit and integration tests cannot be run on arm64.
+
+There are two ways to approach this issue. The first is to only run tests for amd64 and label all arm64 artifacts as experimental. The second is to use self-hosted or third-party arm64 runners so everything can be tested on both architectures. Given that paketo bulidpacks are used for production workloads, I think having arm64 runners will be necessary.


text/0000-manifest-list-buildpacks.md
```diff
@@ -0,0 +1,133 @@
...
+# 
+```
+
+## Rationale and Alternatives
+
+The upstream pack project is already creating multi-arch images for pack and the lifecycle using `docker buildx`, which creates manifest list images. An alternative approach is to create and publish separate images that are tagged by architecture. This would be the steps I mentioned above without the final manifest list image. I think this is an anti-pattern at this point given that multi-arch images are already a norm. I also suspect end users would end up creating their own manifest list images anyway. I think the biggest driver for this is to avoid end users cloning and rebuilding paketo buildpacks.
+
+## Implementation
+
+### Changes to packit
```
loewenstein, line 62: Should we have a similar chapter about *Changes to libpak*?

loewenstein:
I haven't really been following the existing arm64 work in the cloud native buildpacks and paketo communities. One thing I am wondering about in the context of this RFC is what the longer-term picture is? Is this RFC a stepping stone towards `pack build --multi-arch` (or whatever the flag for this would be) producing multi architecture application images with Paketo buildpacks - I assume that's what we want in the end.

text/0000-manifest-list-buildpacks.md
```diff
@@ -0,0 +1,133 @@
...
+
+Users would have the same experience using the manifest list images on amd64, while allowing those who want/need arm64 images to benefit as well.
+
+
+## Prior Art
+
+As mentioned above @dmikusa put together [this](https://github.com/dmikusa/paketo-arm64) excellent guide for building arm64 images with paketo buildpacks. This repo, based on the aforementioned example, is publishing arm version of paketo builders and buildpacks. While it is great to see community involvement in pushing this forward, I think it is time for the paketo maintainers to agree on a path forward so the community can start contributing towards that goal.
+
+
+## Unresolved Questions and Bikeshedding
```
robdimsdale, line 127: The biggest issue I see with arm support on buildpacks is compiling dependencies (like cpython) for arm64.

We can compile the buildpacks themselves via Go's built-in cross-compilation, but that doesn't extend to arbitrary compilation (e.g. Cpython is written in C and uses `gcc` to compile).

Many language families use prebuilt, upstream dependencies (e.g. Go), so if they provide pre-built versions for the arm64 architecture we can just use them using the changes you describe above.

Some buildpacks (like cpython) support building from source during build-time, so that would be a workaround too. But for buildpacks which use pre-compiled dependencies and don't have the escape hatch of compiling during build (e.g. web-servers, ruby), they just won't work on arm64 until we have the ability to compile the dependencies for arm64.

text/0000-manifest-list-buildpacks.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Testing on arm64
+
+As discussed [here](https://github.com/actions/runner-images/issues/5631), github actions currently only supports amd64 runners. This means that unit and integration tests cannot be run on arm64.
+
+There are two ways to approach this issue. The first is to only run tests for amd64 and label all arm64 artifacts as experimental. The second is to use self-hosted or third-party arm64 runners so everything can be tested on both architectures. Given that paketo bulidpacks are used for production workloads, I think having arm64 runners will be necessary.
```
robdimsdale, line 133: We can do both - start with experimental support for arm images for buildpacks that don't require dependencies compiled for arm64 (see my comment above) while we work on getting arm64 runners.

For buildpacks that don't have dependencies that need compiling, we should be able to cross-compile the buildpacks themselves and drop-in arm64 dependencies from upstream, and distribute them as "experimental" before we are truly able to test them.

That seems like a helpful path forward - the community gets easy access to arm64 buildpacks and we get feedback on them, while we work on getting the necessary prerequisites (i.e. arm64 runners) to be confident in labeling the buildpacks production ready.

loewenstein: +1 to start this endeavour with an experimental flag at first.

text/0000-manifest-list-buildpacks.md
```diff
@@ -0,0 +1,133 @@
...
+## Detailed Explanation
+
+Right now binaries are created on/for amd64 and then packaged up into buildpacks using pack. It should be easy to create binaries for arm64 which can then be packaged up into buildpacks using pack. These architecture-specific buildpacks can be pushed to a registry with a platform specific tag as shown in the examples below. 
+
+* `paketobuildpacks/ca-certificates:3.6-amd64`
+* `paketobuildpacks/ca-certificates:3.6-arm64`
+
+The last step would be to create a manifest list image using `docker manifest` commands (or similar) and push it to the registry.
+
+Users would pull the manifest list image such as: `paketobuildpacks/ca-certificates:3.6`
```
robdimsdale, line 23: I don't know much about multi-arch images, but I assume hosting a multi-arch manifest at this location is backwards-compatible with the (amd64-only) image that is currently hosted there?

text/0000-manifest-list-buildpacks.md
```diff
@@ -0,0 +1,133 @@
...
+  ]
+}
+# 
+```
+
+## Rationale and Alternatives
+
+The upstream pack project is already creating multi-arch images for pack and the lifecycle using `docker buildx`, which creates manifest list images. An alternative approach is to create and publish separate images that are tagged by architecture. This would be the steps I mentioned above without the final manifest list image. I think this is an anti-pattern at this point given that multi-arch images are already a norm. I also suspect end users would end up creating their own manifest list images anyway. I think the biggest driver for this is to avoid end users cloning and rebuilding paketo buildpacks.
+
+## Implementation
```
robdimsdale, line 60: This RFC will need to be coordinated with #287 because that will move the location of the dependency metadata out of buildpack.toml into a dynamically-discovered location.

I think the two RFCs are compatible, but we will need to ensure that changes proposed by the first RFC to be accepted are reflected in the second RFC.

I don't think this is a problem, but I think we should be aware of it.

cc @paketo-buildpacks/steering-committee for keeping track of RFCs that have potential to overlap.

ForestEckhardt: Yup we talked about this in the WG of 2023-04-27 we are going to keep an eye on both of these and adapt accordingly.

text/0000-manifest-list-buildpacks.md
```diff
@@ -0,0 +1,133 @@
...
+Users would have the same experience using the manifest list images on amd64, while allowing those who want/need arm64 images to benefit as well.
+
+
+## Prior Art
+
+As mentioned above @dmikusa put together [this](https://github.com/dmikusa/paketo-arm64) excellent guide for building arm64 images with paketo buildpacks. This repo, based on the aforementioned example, is publishing arm version of paketo builders and buildpacks. While it is great to see community involvement in pushing this forward, I think it is time for the paketo maintainers to agree on a path forward so the community can start contributing towards that goal.
+
+
+## Unresolved Questions and Bikeshedding
+
```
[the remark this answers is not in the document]
robdimsdale: cc @ForestEckhardt I think you have been tracking the upstream CNB stuff the most

dmikusa:
> One thing I am wondering about in the context of this RFC is what the longer-term picture is? Is this RFC a stepping stone towards `pack build --multi-arch` (or whatever the flag for this would be) producing multi architecture application images with Paketo buildpacks - I assume that's what we want in the end.

@loewenstein the current work upstream is adding a lower level feature like `Docker manifest` so you can compose two images into an index. 

Down the road that will be more automatic. The normal `pack` commands will be enhanced to support building multi arch images.

In the meantime Paketo will have to fill in the gaps in our pipelines. Maybe with some additional tooling too. That's what we're doing with the stack builds now, enhancing jam to publish a multi arch image for the stack.

text/0000-manifest-list-buildpacks.md
```diff
@@ -0,0 +1,133 @@
...
+    uri = "https://go.dev/dl/go1.18.4.linux-arm64.tar.gz"
+```
+
+### Github actions and workflows
+
+I haven't looked at github workflows for all buildpacks, but I believe [jam](https://github.com/paketo-buildpacks/jam) is being used to package buildpacks. I did see skopeo being used to push images ([cpython](https://github.com/paketo-buildpacks/cpython/blob/main/.github/workflows/push-buildpackage.yml#L62)).
+
+My proposal is to update jam to create buildpacks for amd64 and arm64. I haven't used jam yet, so I don't have specific code changes for it and would appreciate thoughts on how this could be added.
+
+These can then be pushed using skopeo with the above mentioned architecture-specific tags. Finally some `docker manifest` commands can be used to create the manifest list image and then push it to the registry.
```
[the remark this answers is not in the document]
dmikusa: +1

text/0000-manifest-list-buildpacks.md
dmikusa: As a project, we have access to ARM runners so we can build packages where necessary. That will probably require work, not sure if the existing build will just run on Arm64, there could be differences.

I believe that we start with what Buildpacks we have though and work up from there. The Arm64 support will have to be beta for a while well we test it and get all the Buildpacks functional on it.

text/0000-manifest-list-buildpacks.md
dmikusa: We should absolutely cross compile the Buildpacks. I have been doing this in the Arm64 scripts I have and it works perfectly.

We do have access to arm runners through BuildJet but we should reserve these for things where it is absolutely necessary like compiling dependencies. They are not free like the standard GitHub runners.

text/0000-manifest-list-buildpacks.md
dmikusa: I think so but that would be a good experiment. Just to confirm.

text/0000-manifest-list-buildpacks.md
dmikusa: I also don't think we need to run all of the test suites on both architectures or use an experimental label. There's very little that is architecture specific about a buildpack. It is really just the dependencies. That code can be mocked out in tests so we can be confident functionality works correctly regardless of the architecture.

We could potentially look at a small suite of smoke tests for both architectures that we run before release of composites or possibly using some sort of test flags to only run a subset of tests on Arm64. I think we should start without doing this though and see how things go, there will be a period of experimental/beta support for Arm64 as we develop it. We can just watch to see if any platform specific issues crop up and react accordingly.

dmikusa:
@jpena-r7 Thanks so much for writing this up. Excellent to pull all of this information together and help get everyone focused on this topic.

text/0000-manifest-list-buildpacks.md
robdimsdale: Honestly I'm mostly concerned about the integration / end-to-end testing of the arm64 compiled dependencies. We already run into issues just with the amd64 compiled dependencies and we have fairly decent test suites for those. I would not be confident calling the arm64-compiled-dependency buildpacks production-ready without integration testing them

robdimsdale: Yes and no. For the buildpacks that don't use `libpak` (i.e. everything except java/utilities), in order to get the current level of integration test coverage and confidence we have for amd64, we have to get each and every buildpack working on arm64. The integration tests for each buildpack use the buildpack before it. For example, the integration tests for `pip-install` run a `pack build` with the `pip-install` buildpack but also with `cpython` and `pip`. Sure, we could just unit-test the buildpack on arm64, but I don't feel comfortable with that reduced level of support.

If you have a different level of comfort for the java buildpacks - which would make sense because you don't really have integration tests for the java buildpacks in the same way as the other Paketo buildpacks - then I'd understand that, but I don't think we can extrapolate that to become a general assertion for all Paketo buildpacks.

text/0000-manifest-list-buildpacks.md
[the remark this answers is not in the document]
joe-kimmel-vmw: 0.17 release of lifecycle (forthcoming this month) will be the first one to support buildpacks without a stack ID.  stack IDs will be deprecated with support for  at least 18+ months

text/0000-manifest-list-buildpacks.md
jpena-r7: All set. I add a section for libpak as well.

text/0000-manifest-list-buildpacks.md
jpena-r7: Yes. This image for the current platform will be pulled.

I set up the following action to test a multi-arch builders on amd64 and arm64 and it pulls the correct image. The same would be true for buildpacks.

https://github.com/jericop/cnb-builder/blob/01a194b802311e09c9acffb53470a374f8d618b3/.github/workflows/build-and-publish.yml#L90
https://github.com/jericop/cnb-builder/blob/01a194b802311e09c9acffb53470a374f8d618b3/.github/workflows/build-and-publish.yml#L102
https://github.com/jericop/cnb-builder/blob/01a194b802311e09c9acffb53470a374f8d618b3/.github/actions/multi-arch-builder-test/action.yml

text/0000-manifest-list-buildpacks.md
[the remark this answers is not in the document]
jpena-r7: Thanks for the update @joe-kimmel-vmw. I add that link and some others to the doc for easy reference. I also make a point about back compatibility in the new `Proposed approach` section.

text/0000-manifest-list-buildpacks.md
robdimsdale: Great - thanks for clarifying.

robdimsdale:
@jpena-r7 FYI looks like you made changes under the @jericop account which hasn't signed a CLA. I'm not sure if you want to recreate them under your Rapid 7 account or sign the CLA on the other account.

text/0000-manifest-list-buildpacks.md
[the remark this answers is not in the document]
robdimsdale: Ok, so it sounds like we can move forward with this RFC to create multi-arch buildpacks, and then figure out what it means to have multi-architecture support without stack IDs in a separate RFC in the future.

text/0000-manifest-list-buildpacks.md
robdimsdale: There's a thread below which has more context on exactly what we compile/test on which runners. I'll close this thread in favor of that.

text/0000-manifest-list-buildpacks.md
```diff
@@ -0,0 +1,159 @@
...
+    source = "https://go.dev/dl/go1.18.4.linux-arm64.tar.gz"
+    source_sha256 = "35014d92b50d97da41dade965df7ebeb9a715da600206aa59ce1b2d05527421f"
+    uri = "https://go.dev/dl/go1.18.4.linux-arm64.tar.gz"
+```
+
+### Github actions and workflows
+
+I haven't looked at github workflows for all buildpacks, but I believe [jam](https://github.com/paketo-buildpacks/jam) is being used to package buildpacks. I did see skopeo being used to push images ([cpython](https://github.com/paketo-buildpacks/cpython/blob/main/.github/workflows/push-buildpackage.yml#L62)).
+
+My proposal is to update jam to create buildpacks for amd64 and arm64. I haven't used jam yet, so I don't have specific code changes for it and would appreciate thoughts on how this could be added.
```
robdimsdale, line 135: As I'm thinking more about this, it's not obvious to me what, if any, changes are required to `jam pack` to support building multi-arch buildpacks. `jam pack` doesn't compile the buildpack; it just packages it. It also doesn't create OCI images - `pack buildpack package` does this.

Expanding on this in more detail:

The current workflow to create a buildpack is:

1. Compile the `build` and `run` binaries (e.g. `go build`).
1. Run either `jam pack` (for `packit` buildpacks) or `packager` (for `libbuildpack` buildpacks). This primarily creates a tgz of the buildpack files. For `packit` buildpacks, this also does some extra steps like setting the buildpack version and optionally downloading dependencies. I'm not sure what other steps `packager` takes.
1. `pack buildpack package` to turn that `tgz` into an OCI image.

Currently, `jam pack` creates a single tarball of a single buildpack with no concept of multi-architecture. You can optionally filter dependencies by `--stack` to only include the stack-specific dependencies in the resultant metadata (and downloaded dependencies), but given that stacks can be multi-architecture, this isn't really sufficient to filter dependencies by architecture.

It's not really obvious to me what it would mean for `jam pack` to support packaging up a buildpack in a multi-architecture way. It doesn't compile the binaries, and it doesn't produce OCI images. Filtering the dependencies by architecture would be an optimization to reduce metadata (and resultant buildpack size if downloading dependencies), but it's not a necessary step for creating the buildpack.

I suppose you could make modifications to `jam pack` to support packaging multiple versions of `build` and `run` for each target architecture and create a wrapper around them to invoke the correct one at run-time. But it seems just as easy to invoke `jam pack` multiple times - once per architecture - to create architecture-specific tarballs.

I don't know if `pack buildpack package` already supports multi-arch images. I took a quick look at the code, and it [only seems to be aware of `platform`](https://github.com/buildpacks/pack/blob/12bc0e241895c9fec3940164ea9771f7ea4f5e53/buildpackage/config_reader.go#L22), which [only contains `OS`](https://github.com/buildpacks/pack/blob/12bc0e241895c9fec3940164ea9771f7ea4f5e53/pkg/dist/dist.go#L38) (i.e. `linux` or `windows`). This leads me to think either we'd have to make changes to `pack`, or we'd have to create separate docker images for each architecture, and bundle them together. Perhaps that's what `docker buildx imagetools create` does - I'm not sure. Perhaps @dmikusa or @dashaun have more info on how to build multi-arch buildpacks in a single manifest?

text/0000-manifest-list-buildpacks.md
```diff
@@ -0,0 +1,159 @@
...
+The number one question I get about [paketo] buildpacks is whether they support arm. It is safe to say that workloads are generally moving to arm processors and end users are already evaluating how to build containers on arm. It is difficult for me as an end user to seriously consider building arm images with paketo buildpacks without putting in some engineering effort into building arm versions of the paketo buildpacks I may need. While the above example is definitely helpful, a lot of the steps can be built into the github actions workflows that create and release buildpacks for general consumption. I believe arm support was the number one requested roadmap item for 2023, and this proposal is an attempt to jump start that process.
+
+## Detailed Explanation
+
+Right now binaries are created on/for amd64 and then packaged up into buildpacks using pack. It should be easy to create binaries for arm64 which can then be packaged up into buildpacks using pack. These architecture-specific buildpacks can be pushed to a registry with a platform specific tag as shown in the examples below. 
+
+* `paketobuildpacks/ca-certificates:3.6-amd64`
+* `paketobuildpacks/ca-certificates:3.6-arm64`
```
robdimsdale, line 18: As I'm reading about `docker buildx`, a lot of the examples seem to push a single image with multiple tags from a single `Dockerfile`. It looks like `docker buildx imagetools create` does the same thing but with multiple source images that already exist on a registry.

Is it possible to avoid the intermediate step of publishing buildpacks to the `:<version>-<arch>` tag? It seems unnecessary once we've created a single manifest/image for all platforms. Could we create the images on a local registry and just push the single multi-arch image?

jericop:
> @jpena-r7 FYI looks like you made changes under the @jericop account which hasn't signed a CLA. I'm not sure if you want to recreate them under your Rapid 7 account or sign the CLA on the other account.

I plan to start contributing from @jericop going forward so I signed the CLA. Sorry for the confusion.

text/0000-manifest-list-buildpacks.md
jpena-r7: You are correct that `pack` does not support multi-arch yet, and the new changes being introduced will allow users to create manifest images after they have already been built.

[This](https://github.com/jericop/cnb-builder/blob/01a194b802311e09c9acffb53470a374f8d618b3/.github/actions/multi-arch-builder-test/action.yml) is an example of creating multi-arch builders. It creates and publishes arch-specific tags, but it is meant for demonstration purposes and testing so it gives an idea.

Creating multi-arch buildpacks should be simpler and may not even require a local registry because no other images are pulled in. That should allow you to create the arch-specific buildpack images locally and then create a manifest from there and push the manifest.

Let's say you compile a buildpack for amd64 in one folder `amd64` and arm64 in another folder `arm64`. You then create local buildpack images for each arch such as `mybuildpack:0.0.1-amd64` and `mybuildpack:0.0.1-arm64`. Then you can just create and push a manifest image as shown below and it would work on both architectures.

`docker buildx imagetools create -t mybuildpack:0.0.1 mybuildpack:0.0.1-amd64 mybuildpack:0.0.1-amd64`

There is one issue that needs to be considered though. At this time pack seems to set the architecture to `amd64`. I had to work around this issue when creating the multi-arch builder [here](https://github.com/jericop/cnb-builder/blob/01a194b802311e09c9acffb53470a374f8d618b3/.github/workflows/build-and-publish.yml#L27), but I just wanted to call this out.

robdimsdale: Yes, that is exactly what I was thinking. I don't think we need to make any changes to `jam pack` to achieve this.

It would look something like the following:

1. for arch in [amd64, arm64]
    1. compile buildpack executables for target architecture via `go build`; put them in arch-specific temp directory
    1. package buildpack with `jam pack` into arch-specific temp directory
    1. create arch-specific OCI image via `pack buildpack package`, passing in contents of arch-specific temp directory
    1. upload image to local docker daemon/container-runtime/engine (whatever the correct terminology is here) with arch-specific tag

1. create single multi-arch image via `docker builds imagetools create` and the arch-specific tagged images previously uploaded to local daemon
1. publish single image to relevant registries

text/0000-manifest-list-buildpacks.md
dmikusa: I'm OK with agreeing to disagree on this point, I think we can do that and still move forward with this effort. Each language family team needs to make the cost/effort versus benefit evaluation and act accordingly. I'm confident each team will make the right choices for their user bases, so perhaps we can put the following takeaways into the RFC:

1. The project has ARM64 runners for teams to use as needed.
2. Each language team will need to decide what "production-ready" means. Things for each team to consider: confidence in the code, automated release for timely updates, and feature parity with x86.

Anything else come to mind as a good guideline?

robdimsdale: That sounds perfectly reasonable. I don't think I have much else to add to the takeaways.

text/0000-manifest-list-buildpacks.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Implementation
+
+### Upstream changes to pack to consider
+
+[This](https://github.com/buildpacks/rfcs/blob/main/text/0096-remove-stacks-mixins.md) rfc captures a lot of the concepts implemented in the spec changes. The following [PR](https://github.com/buildpacks/spec/pull/336/files#) to the buildpack spec introduces two new environment variables that will be set by the lifecycle during the detect and build phases, `CNB_TARGET_OS` and `CNB_TARGET_ARCH`. These changes are available in version `0.12` of the platform api as outlined [here](https://github.com/buildpacks/spec/pull/335). These changes were partially implemented in [this](https://github.com/buildpacks/lifecycle/issues/743) issue, but they have not yet made into pack ([here](https://github.com/buildpacks/pack/blob/v0.30.0-pre1/builder/config_reader.go)).
+
```
dmikusa, line 64: I think we should acknowledge stack removal but mark it as out of scope for this RFC, and just target the existing stack-based implementation. There's a whole ton of work that needs to happen for stack removal and some of it is just going to take time as we transition from one to the other.

I think if we focus on the stack-based implementation we have now, a.) we can simplify this RFC, b.) we can start work immediately on it, and c.) we can still implement because we just assume that a stack will have the require arch or the build wouldn't even run (I believe `pack build` would fail cause it couldn't download the right stack images, or it would fallback to amd64 images if on a Mac which permits both).

text/0000-manifest-list-buildpacks.md
```diff
@@ -0,0 +1,159 @@
...
+### Proposed approach 
+
+Once all upstream changes to pack have been fully implemented and the newest version of pack and the lifecycle are being used the values in `CNB_TARGET_OS` and `CNB_TARGET_ARCH` will be set by paketo in the builder config toml. This means the changes I'm proposing to  `libpak` and `packit` below can just read those environment variables and match any dependencies to the appropriate os and architecture. 
+
+Since there are a lot of moving pieces to this project, I suggest the change from using stacks to targets (run and build images) be done separately. This would mean that the `CNB_TARGET_OS` and `CNB_TARGET_ARCH` environment variables would need to be set in the function/method that reads the buildpack. At build time buildpacks should attempt to read those environment variables if set, but otherwise fall back to determining os and architecture from another source such as `GOOS` and `GOARCH` environment variables. This should ultimately match whatever is used to set those values in the builder too.
```
dmikusa, line 69: For now, I think we should just skip the `CNB_TARGET_*` stuff and use `GOARCH` to match the right architecture for dependencies where it is required. This should be reliable as the stack that gets picked will have the right architecture. Then the build will use `GOARCH` and install deps for that arch.

text/0000-manifest-list-buildpacks.md
dmikusa: More of an implementation detail, not sure it needs to go into the RFC, but `skopeo` won't work with multi-arch images so `jam` may need some tooling for that. Not sure if the commands I added for publishing the stacks will work with buildpacks as well? Maybe 🤔 

On the pipeline-builder side, we use `crane` and it does not work with multi-arch images either. We'll need a similar tool. It could be a good chance for both pipelines to remove some 3rd party tools and rally around `jam` as our solution.

dmikusa: More thoughts. Pack is adding [a `manifest` command for generating manifest images](https://github.com/buildpacks/pack/pull/1705). This is part of their first step towards multi-arch support. If it is ready in time, it could be something we utilize instead of `docker`. Not sure what folks think about this?

text/0000-manifest-list-buildpacks.md
```diff
@@ -0,0 +1,159 @@
...
+  ]
+}
+# 
+```
+
+## Rationale and Alternatives
+
+The upstream pack project is already creating multi-arch images for pack and the lifecycle using `docker buildx`, which creates manifest list images. An alternative approach is to create and publish separate images that are tagged by architecture. This would be the steps I mentioned above without the final manifest list image. I think this is an anti-pattern at this point given that multi-arch images are already a norm. I also suspect end users would end up creating their own manifest list images anyway. I think the biggest driver for this is to avoid end users cloning and rebuilding paketo buildpacks.
+
+## Implementation
```
dmikusa, line 59: I think that this RFC is missing information about stacks and builders. Both are really necessary to provide first-class support for Paketo users. I'd be OK calling it out-of-scope, if we don't want this RFC to get bigger, but I still think we should mention it because it's part of the overall picture to say that we support arm64.

jericop: Great point and I completely agree. I picked up where you left off on multi-arch stacks and will then look at builders. Once we have those in place I think we can move forward with this RFC.

text/0000-manifest-list-buildpacks.md
sophiewigmore: Circling back on this, what's the outcome? Can we go for publishing the single multi-arch image?

text/0000-manifest-list-buildpacks.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Changes to packit
+
+`Os` and `Architecture` string fields can be added to the [`Dependency`](https://github.com/paketo-buildpacks/packit/blob/v2/postal/buildpack.go#L15) struct in `packit/postal` to allow buildpacks to interrogate and use the new metadata when determining what external dependencies to download. This will allow `buildpack.toml` files to be updated to add in the `os` and `architecture` metadata for all dependencies. They would all be set to something like `os=linux` and `architecture=amd64` initially.
+
+Some logic can be added to the [Resolve](https://github.com/paketo-buildpacks/packit/blob/18202009038b0df285ba0fb7d8b43abbf60d3ed0/postal/service.go#L89) method of `packit/postal` to pick a version that matches the `GOOS` and `GOARCH` environment variables whenever the `os` and `architecture` fields have been specified.
```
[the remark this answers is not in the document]
sophiewigmore: Maybe the packit-specific code recommendations can be captured in an issue also

dmikusa:
> With work being in progress already to fulfill this not yet accepted RFC... Should we make sure this actually get's merged as a basis of that ongoing work? WDYT @paketo-buildpacks/steering-committee?

I'm unsure of what we actually need to put into an RFC. 

I could see us putting together an RFC that signals the intent to publish multi-arch images across Paketo. It would really just be a signal to users that we intend to do this across the whole project. IMO, it's probably better to just keep writing blog posts about our progress. Those are just as easy to site when folks ask questions about multi-arch support.

I don't think there's any technical coordination that needs to happen to publish multi-arch images though. Upstream has made tooling available to publish multi-arch images. We've done it for a number of buildpacks and it works fine. I don't see any reason why folks would not use that functionality in their implementations, but I suppose we could include that in an RFC.

Maybe there are some details with the stacks/builders that would be helpful to work out in an RFC. I would defer to that team though. From what I've observed, they're already working on it and I don't want to add a process if it's not needed. If everyone is in agreement on how we proceed, then we probably don't need an RFC and we just focus on the work.

That said, happy to help push an RFC forward if folks think it's helpful. Just let me know.

loewenstein-sap:
> I'm unsure of what we actually need to put into an RFC.

Sure. All I am saying is that this is a draft PR for an RFC - so either we get it merged in one shape or another or we close it with a comment that it doesn't need an RFC.

Either way, I would suggest we at least create a tracking issue for all the work involved, similarly to the one I created for adopting Ubuntu Noble. That way we can ping all maintainers and track overall progress in a central place.

dmikusa:
@loewenstein +1 to both those points. I'd prefer we close this RFC-draft. Then use the issue to track the works that's been done and what's still to be completed.

dmikusa:
If anyone objects to closing this RFC-draft. Please let us know, or we'll close this in one week.

loewenstein-sap:
Seems like multi-arch is well under way without this RFC. Closing now as suggested by @dmikusa above.

PR #289, opened 2023-05-02
Add Bionic end-of-support RFC.

robdimsdale:
[Readable](https://github.com/paketo-buildpacks/rfcs/blob/bionic-eos/text/0000-bionic-eos.md)

### DIFF
--- a/text/0000-bionic-eos.md
+++ b/text/0000-bionic-eos.md
@@ -31,6 +31,7 @@ This end of support will entail the following:
 ### Documentation
 1. Ensure all documentation (and samples) replaces all references to Bionic with Jammy (22.04).
 1. Write a blog post to announce this end of support.
+1. Add documentation around migrating. This would be both general (i.e. check default builders) and language-specific (e.g. builders might be present in existing Spring Boot configuration).
 
 We will leave it up to the relevant maintainers (stacks, builders, language-families, content) to decide on a suitable timeline for completion of the above items. For example, language family maintainers may wish to continue building and including Bionic dependencies even after the Bionic stacks/builders are removed if they believe it provides value to their consumers. This RFC will not be fully implemented until all the above points are addressed, though.
 
--- a/text/0000-bionic-eos.md
+++ b/text/0000-bionic-eos.md
@@ -1,0 +1,58 @@
+# End of Life for Ubuntu 18.04 (Bionic)
+
+## Summary
+
+On May 31st 2023, Ubuntu 18.04 (Bionic Beaver) will go out of support for OSS customers ([source](https://ubuntu.com/blog/18-04-end-of-standard-support)). At this time, Paketo will stop supporting this stack.
+
+We will stop building new stacks, builders, and Bionic-specific dependencies, but we will not remove any of the existing artifacts (images, dependencies, etc). This will ensure backwards-compatibility where needed.
+
+## Motivation
+
+There will be no more updates to Ubuntu Bionic after the End of Support, so the stacks will become vulnerable as CVEs are discovered and not fixed upstream. Similarly, buildpack dependencies (like go, cpython, etc) may not fix vulnerabilities if they are only exploitable on Ubuntu Bionic.
+
+It is a better user experience to explictly remove support for these stacks rather than continue to support them knowing that they are potentially vulnerable. It will also reduce complexity in CI and reduce the number of dependencies in a buildpack's `buildpack.toml` file.
+
+## Detailed Explanation
+
+This end of support will entail the following:
+
+### Stacks
+1. Disable the CI that builds the Bionic stacks.
+1. Archive the GitHub repositories for the Bionic stacks (with a note in the README indicating end of support, and with all other content preserved for posterity/discoverability).
+
+### Builders
+1. Disable the CI that builds the Bionic builders.
+1. Archive the GitHub repositories for the Bionic builders (with a note in the README indicating end of support, and with all other content preserved for posterity/discoverability).
+
+### Buildpacks
+1. Stop compiling dependencies for Bionic in buildpacks that currently support them.
+1. Remove Bionic-specific dependencies from the `buildpack.toml` metadata.
+
+### Documentation
+1. Ensure all documentation (and samples) replaces all references to Bionic with Jammy (22.04).
+1. Write a blog post to announce this end of support.
+
+We will leave it up to the relevant maintainers (stacks, builders, language-families, content) to decide on a suitable timeline for completion of the above items. For example, language family maintainers may wish to continue building and including Bionic dependencies even after the Bionic stacks/builders are removed if they believe it provides value to their consumers. This RFC will not be fully implemented until all the above points are addressed, though.
+
+## Rationale and Alternatives
+
+1. Continue to run the stacks/builders CI. This is pointless because there will not be new updates so the CI will be wasting resources checking for updates.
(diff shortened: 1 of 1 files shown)


text/0000-bionic-eos.md
```diff
@@ -0,0 +1,56 @@
...
+1. Disable the CI that builds the Bionic builders.
+1. Archive the GitHub repositories for the Bionic builders (with a note in the README indicating end of support, and with all other content preserved for posterity/discoverability).
+
+### Buildpacks
+1. Stop compiling dependencies for Bionic in buildpacks that currently support them.
+1. Remove Bionic-specific dependencies from the `buildpack.toml` metadata.
+
+### Documentation
+1. Ensure all documentation (and samples) replaces all references to Bionic with Jammy (22.04).
+1. Consider writing a blog post to announce this end of support.
```
dmikusa, line 31:
```suggestion
1. Write a blog post to announce this end of support.
```

I definitely think we should do this. I think it would be good to post something now about 30 days before and in time to comment on this RFC, and also something when it actually happens at the end of the month.

text/0000-bionic-eos.md
```diff
@@ -0,0 +1,56 @@
...
+1. Continue to include the dependencies in `buildpack.toml`. This is confusing to buildpack authors and potentially misleading if they aren't being updated/compiled upstream.
+1. Leave documentation/samples/blog as-is. This is a poor user experience for consumers as they will see references to Bionic stacks even though they are not supported
+
+## Implementation
+
+* Update stack/builder READMEs with end-of-support notices and archive the repositories (this will also disable the GitHub actions that build the stacks/builders)
+* Remove Bionic dependency compilation code and GitHub actions for buildpacks that compile their dependencies
+* Remove Bionic dependencies from `buildpack.toml`
+* Update website/samples/blog to remove references to Bionic stacks, builders, and dependencies
+
```
dmikusa, line 49: Add notes on migrating. Thoughts off the top of my head, which I shared on Slack.

```
Migration from Bionic to Jammy with buildpacks should be straightforward. Run `pack config default-builder paketobuildpacks/builder-jammy-base` (or `-full` or `-tiny`, depending on your preference). This will change the default builder used when you create images with pack.

It is possible to override the default, so check for the following as well:

- In some cases, you might be passing the `--builder` or `-B` flag to `pack build`, which will override the builder for that particular build. In that case, please make sure to update the commands you’re running or scripts to also use the Jammy builders. 

- For Spring Boot users, check your pom.xml or build.gradle files to see if you are specifying a builder (search the file for `builder`), and update those as well.

- One more place to check, look at your CI pipelines to see if you have any which are setting a builder, possibly when running pack, and update those also.
```

dmikusa: Also, we really, really need to try and get `pack builder suggest` either a.) updated or b.) redirect the existing stacks to Jammy. We really don't want that to be pointing to old builders. That throws off the new user experience a lot.

ForestEckhardt (dismissed):

sophiewigmore:
This all looks reasonable to me. I'm in agreement wit Dan's comments above, but aside from that I approve of the content here. I'm holding off on an official approval for a bit longer while we wait for anyone in the community to voice questions/concerns.

text/0000-bionic-eos.md
robdimsdale: Given that there isn't really a viable way for Paketo as an OSS to continue supporting a Bionic stack when upstream support has stopped, I don't think there's much value in two blog posts.

I think it's valuable to write a single blog post at the point of transition explaining what happened, but writing two just seems noisy and unnecessary effort. And potentially a bad experience if we invite folks to comment and then have to dismiss their comments because they want to see Bionic supported longer.

Maybe I'm over thinking the last point. But I'm not sure two blog posts is better than one.

dmikusa: I think it's a fair point that there's very little one could comment about this issue. You're 100% right that we cannot extend the time to EOL. That's what most people would ask about and the EOL date is what it is and that's final.

I keep going back to giving folks as much time as possible to make the switch though. You can already switch to Jammy now, so telling people to go start switching now gives them that much more time to make the switch. That said, it's not like it stops working on Jun 1st, so it's not the end of the world either way.

I'd be happy to write up a blog post to that effect to post ASAP if that would help. It'll be short.

robdimsdale: Sure. I have no objections to a blog post that says: "Paketo will stop supporting Bionic after May 31 2023.". Arguably we should have written that months ago.

My concerns are just pragmatic:
* Until we finalize this RFC we can't state what the actual outcome will be (i.e. do we remove old stacks, do we just stop updating them? Are we leaving it up to language family maintainers to decide when to drop Bionic dependencies, etc) until this RFC is accepted.
* We don't publish a lot of blog posts, so we're going to end up with two blog posts back to back, 2-3 weeks apart, both saying pretty much the same thing about Bionic EOL.
* I don't have a lot of time over the next couple of weeks to wordsmith a blog post, certainly not two. Maybe I hold myself to an unnecessarily high standard, but I find that blog posts take longer than other forms of communication (like RFCs, issues, etc).

I think we should publish a blog post at the time of EOL with a clear set of expectations (i.e. the outcome of this RFC). I think that will be a better reference for Paketo's support than the accepted RFC.

dmikusa: I am just anticipating the "why didn't you announce this sooner" comments, which are fair, and thought we could get ahead of that, even if it's just a month. 

Anyway, I agree with what you're saying, all good points, and I'm OK with that approach.

dmikusa: Can we still change this from "Consider writing" to "Write", so that it's clear will will 100% write at least one blog post.

robdimsdale: Yeah, I'm confused because I thought I merged that change 🤔

robdimsdale: oh I can't - maybe because I force pushed an update and it invalidated your change. I'll make your change and push it.

robdimsdale: I force-pushed your suggestion which invalidated and closed this issue. Good job we're broadly in agreement 😄

robdimsdale: Also, I hope that by not deleting the old stacks and builders, we create a softer offboarding experience. Folks can still use the bionic stacks/builders, they just might become insecure as CVEs are discovered and not fixed.

text/0000-bionic-eos.md
dmikusa: Pack release should be soon. 🥳

robdimsdale: I agree with the migration notes. I added a step to include this in the documentation. I didn't put the specifics in because I think that's too detailed for the RFC.

dmikusa (approved):

ForestEckhardt (approved):

ForestEckhardt:
@paketo-buildpacks/stacks-maintainers, @paketo-buildpacks/nodejs-maintainers, @paketo-buildpacks/ruby-maintainers, and @paketo-buildpacks/python-maintainers the EOS date is rapidly approaching. I think it would be smart for us to have this closed before that date so that work can start as soon as possible. Your feedback and review would be greatly appreciated!

brayanhenao (approved):

sophiewigmore (approved):

TisVictress (approved):

sophiewigmore:
@ForestEckhardt I think you now have all of the approvals you need from each team.

PR #290, opened 2023-05-04
RFC for chmod buildpack

anthonydahanne:
## Summary

[readable](https://github.com/anthonydahanne/rfcs/blob/chown-chmod-buildpack/text/utilities/0000-chmod.md)

## Checklist


text/0000-chmod.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Rationale and Alternatives
+
+We have considered solving the issue at different levels:
+
+* At the stack level: reverting to former stacks that define the same user at build time and runtime.
+
+After all, previous stacks like Ubuntu Bionic did not differentiate the runtime user from the build time user, both being named `cnb` with the UID 1000
+
+We began noticing this issue with the latter stacks (some based on Ubuntu Jammy) that guarantee [a more secure environment](https://github.com/buildpacks/rfcs/blob/main/text/0085-run-uid.md) by having the build time user set to UID 1001, and runtime user set to 1000 at runtime.
```
robdimsdale, line 33: The Paketo stacks actually have a run-time UID of `1002`. Other commercial stacks may have run-time UIDs of `1000` or other values.

Regardless, your general point still stands.

dmikusa:
```suggestion
We began noticing this issue with the latter stacks (some based on Ubuntu Jammy) that guarantee [a more secure environment](https://github.com/buildpacks/rfcs/blob/main/text/0085-run-uid.md) by having the build time user set to UID 1002, and runtime user set to 1000 at runtime.
```

text/0000-chmod.md
```diff
@@ -0,0 +1,62 @@
+# Chmod Buildpack
```
dmikusa, line 1: First thing, I think this should be a Utilities buildpack, so the RFC can be a Utilities RFC. All that means, is just move this RFC proposal under the `text/utilities/` subdirectory.

text/0000-chmod.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Summary
+
+A chmod buildpack that allows users to change directories and files (produced at buildtime) UNIX permissions.
+
+This buildpack would be configured to run last, so that it would be able to change any directories and files produced by previous buildpacks.
+
+## Motivation
+
+Buildstacks and Builders rely on Stacks.
```
dmikusa, line 11:
```suggestion
Buildpacks and Builders rely on Stacks.
```

text/0000-chmod.md
```diff
@@ -0,0 +1,62 @@
...
+
+Stacks define build time and runtime base images.
+
+Some stacks do not use the same user / UID between build and run images, [for security reasons](https://github.com/paketo-buildpacks/rfcs/blob/main/text/0045-user-ids.md), since a runtime user should just be able to run a process; not create directories nor install packages.
+
+Unfortunately, some buildpacks do need to create directories and files: they run setup scripts during runtime that fail because they can't write folders nor files, since they were run with a runtime user, who can be restricted with some Stacks.
```
dmikusa, line 17:
```suggestion
Unfortunately, some applications, like `dist-zip` archives, do need to create directories and files: they run setup scripts during runtime that fail because they can't write folders nor files, since they were run with a runtime user, who can be restricted with some Stacks. 
```

This is a small technicality, but I'd say "application" here because buildpacks only run at build time.

text/0000-chmod.md
```diff
@@ -0,0 +1,62 @@
...
+
+This flag, an environment variable named `BP_CHMOD_MAPPING` would be configured with a map of folders and their desired owner and permissions.
+
+For example:
+
+`pack build -e BP_CHMOD_MAPPING='{"/workspace":"0750", "/workspace/bin":"0755"}' myimage`
+
+would activate the chmod buildpack and 
+
+* make it change recursively the `/workspace` folder of the runtime image to be owned by UID 1000 and GID 1000 with permissions `0750`
```
dmikusa, line 53:
```suggestion
* make it change recursively the `/workspace` folder of the runtime image to have permissions `0750`
```

A buildpack can't `chown` or `chgrp` something cause it isn't running as root. It should be enough to set permissions though because the group will be consistent from build & run, so you can make things group executable and it'll run at both build and runtime.

text/0000-chmod.md
```diff
@@ -0,0 +1,62 @@
...
+* make it change recursively the `/workspace/bin` folder of the runtime image to be owned by UID 1000 and GID 1000 with permissions `0755` (making all this folder files executable)
```
dmikusa, line 54:
```suggestion
* make it change recursively the `/workspace/bin` folder of the runtime image to have permissions `0775` (making all these folder files executable)
```

text/0000-chmod.md
```diff
@@ -0,0 +1,62 @@
...
+## Implementation
+
+The chmod buildpack would be added last to builders and would be activated if a specific flag is set.
+
+This flag, an environment variable named `BP_CHMOD_MAPPING` would be configured with a map of folders and their desired owner and permissions.
+
+For example:
+
+`pack build -e BP_CHMOD_MAPPING='{"/workspace":"0750", "/workspace/bin":"0755"}' myimage`
+
```
dmikusa, line 50: This is the part I am most unsure about & I'll say upfront, my comments here could probably also go under bikeshedding :) 

I agree on using an env variable for input since that is the chosen path for Paketo buildpacks, but I get stuck on the format. Shoving JSON or other structured text into an env variable is a real pain. I've had to do this with other CLI tools, and it is not a great user experience. Quoting get's messed up, and is tricky with different shells (and Windows).

I'm trying to think through other examples of data like this in our buildpacks, but nothing is coming to mind. I have a few ideas though.

1. `BP_CHMOD_MAPPING`, but in a more compact format, like `/workspace:0750:r /workspace/bin:0755` (or maybe `|` as the delimiter). I think JSON is OK in the example, but recursive should be a configurable option so that would make the JSON more complicated/verbose. And quoting is painful.

2. `BP_CHMOD_MAPPING_FILE` points to a file with the mappings in it (maybe one per line, same compact format or JSON, it is more manageable in a file). This could be packaged with the app or it could be mounted into the container as a volume mount. Possibly with a default file like `Permissionsfile` at the root.

3. A binding with type `chmod`. Mentioning to be complete, but I don't think this is the right option, cause it doesn't contain secret data and bindings really should be for secret data.

Perhaps we go with a combination of the 1.) and 2.)? If you have only one or two mappings, an env variable is nice. If you have 20, probably not so much.

I'm also wondering about the MVP here, which is another point to consider. 

Anyway, I'm not specifically suggesting anything here, but hoping to spark a conversation and see what folks think about the format.

text/0000-chmod.md
```diff
@@ -0,0 +1,62 @@
...
+would activate the chmod buildpack and 
+
+* make it change recursively the `/workspace` folder of the runtime image to be owned by UID 1000 and GID 1000 with permissions `0750`
```
robdimsdale, line 53: If the intentino is to both `chown 1000:1000` and `chmod <value>`, can we be more explicit about the `chown` throughout this RFC?

Also if we're `chown`ing directories, why bother with `chmod` at all?

robdimsdale: Ah, I just saw @dmikusa's comment above. Maybe we can't `chown` at all as we're not running as root.

dmikusa: Yes, when Anthony & I first discussed this I'd said both. Not thinking about needing to be root to `chown` stuff :( I think this is just left over from his first draft.

text/0000-chmod.md
anthonydahanne: It's interesting you mention option 2. since it's very similar to what the `native` buildpack [provides](https://github.com/paketo-buildpacks/native-image/blob/main/README.md?plain=1#L27). 

At least, that'd make the UX with paketo similar across buildpacks, so yeah makes sense, 1+2, I'll update the RFC

text/0000-chmod.md
anthonydahanne: done

text/0000-chmod.md
anthonydahanne: ok it's pushed!

text/utilities/0000-chmod.md
```diff
@@ -0,0 +1,80 @@
...
+and:
+
+`chmod-mappings.txt`
+```text
+/workspace:0750:r
+/workspace/bin:0755
+```
+
+would have the same effect as the previous example using `BP_CHMOD_MAPPING`
+
```
dmikusa, line 73:
```suggestion

If both `BP_CHMOD_MAPPING` and `BP_CHMOD_MAPPING_FILE` are applied, then `BP_CHMOD_MAPPING` is applied first, then the mappings in the file are applied in order from the top to the bottom of the file.
```

Probably not hugely important, but just makes this edge case specified.

dmikusa:
I'm 👍 , just one last minor suggestion.

dmikusa (approved):

text/utilities/0000-chmod.md
```diff
@@ -0,0 +1,81 @@
...
+would activate the chmod buildpack and 
+
+* make it change recursively the `/workspace` folder of the runtime image to have permissions `0750`
+* make it change non-recursively the `/workspace/bin` folder of the runtime image to have permissions `0775` (making all these folder files executable)
+
+2. An environment variable named `BP_CHMOD_MAPPING_FILE` would be configured with the location of a file defining the mapping (convenient for users configuring numerous mappings)
+
+For example, the combination of:
+
+`pack build -e BP_CHMOD_MAPPING_FILE=chmod-mappings.txt myimage`
```
robdimsdale, line 62: Why not auto-detect on this file? It seems unlikely that this file will exist for any other use-case and we're not shadowing existing filenames (like `pom.xml`). Asking the user to type extra env vars seems cumbersome and unnecessary.

text/utilities/0000-chmod.md
```diff
@@ -0,0 +1,81 @@
...
+
+`chmod-mappings.txt`
+```text
+/workspace:0750:r
+/workspace/bin:0755
+```
+
+would have the same effect as the previous example using `BP_CHMOD_MAPPING`
+
+If both `BP_CHMOD_MAPPING` and `BP_CHMOD_MAPPING_FILE` are applied, then `BP_CHMOD_MAPPING` is applied first, then the mappings in the file are applied in order from the top to the bottom of the file.
```
robdimsdale, line 74: Similar to my comment above, this could instead be:

> if both `BP_CHMOD_MAPPING` is provided and the `chmod-mappings.txt` is present, then `BP_CHMOD_MAPPING` is applied first, then the mappings in the file are applied in order from the top to the bottom of the file.

robdimsdale (approved):
This looks fine to me. I think we can optimize the user experience by auto-detecting on the presence of a `chmod-mappings.txt` file instead of requiring `-e BP_CHMOD_MAPPING_FILE=chmod-mappings.txt` but I don't feel strongly about this.

text/utilities/0000-chmod.md
dmikusa: I have been making an assumption here 🤦 basically that `BP_CHMOD_MAPPING_FILE` would default to a sensible file name and that if users include that file, then this would automatically work like you're saying @robdimsdale. Like how we handle pom.xml/build.gradle search in those buildpacks.

@anthonydahanne Is that what you were thinking? Maybe we should clarify here? 

My concern with not having the env variable is that invariably someone will ask if they can put the file somewhere else and change where the buildpack looks for it.

robdimsdale: Sure. I think that's fair. I like the compromise of providing sensible defaults with extensibility where necessary.

robdimsdale: There was some strong feedback in the working group meeting about ensuring it is hard/impossible for users to accidentally execute this buildpack with dangerous permissions (e.g. `chmod -R 755 /workspace`).

So, it sounds like it would be prudent to keep the environment variable pointing to the file as a required argument.

text/utilities/0000-chmod.md
```diff
@@ -0,0 +1,81 @@
...
+
+* At the buildpack level: [implementing special flags in some buildpacks](https://github.com/paketo-buildpacks/dist-zip/pull/174) that would allow users of those buildpacks to make some folders group writable
+
+But then, we discovered that it was not enough, some other flag would need to be implemented to let the build time user run a different set of scripts; and of course, we would need to replicate this behavior in several other buildpacks packaging applications that set themselves up at runtime.
+
+Allowing the user to keep on using the same secure Stacks, and unmodified buildpacks, using a chmod buildpack at the end of their build seems like the best solution
+
+## Implementation
+
+The chmod buildpack would be added last to builders and would be activated if a specific flag is set.
```
robdimsdale, line 43: There was some strong feedback in the working group meeting about ensuring it is hard/impossible for users to accidentally execute this buildpack with dangerous permissions (e.g. `chmod -R 755 /workspace`).

So, it sounds like it would be prudent to keep this buildpack out of the builders, and require users to opt-in by explicitly providing the buildpack in a custom build order.

robdimsdale: cc @BarDweller @jpena-r7

text/utilities/0000-chmod.md
robdimsdale: I don't personally have strong opinions on this, I'm just representing the feedback from the working group meeting.

text/utilities/0000-chmod.md
BarDweller: I really don't like this buildpack.. I mean, I totally get why it's needed.. but the potential for damage here is huge. 

My main concern is that with this being a standard offering, that you'd see a propagation of 'just chmod everything 755' style 'fixes' starting to be applied by users who discover this 'makes it work' without understanding what they've done. 

Unix permissions are not simple, I get that, but this isn't a fix for the permissions being bad, this is a hammer that will be applied indiscriminately and by people whose main concern is 'making it all work', potentially with little to no understanding of what was broken, or the security, or application implications to the broad permission changes they just applied. (You can consider that this dumps back on the user the responsibility to understand the container being built, which they already made the choice to delegate to buildpacks)

In an ideal world, we shouldn't need this at all. The relevant buildpacks should be able to determine the appropriate permissions for their files to ensure that the application is able to function at runtime. If this is not possible today, then we really need to start identifying those scenarios, and working out how to deal with them. 

If we consider that this is needed because of certain scenarios, AND we envisage those scenarios may be fixed in future, then I'd argue this buildpack should probably offer 'presets' that are known to fix permissions.. so that in future, as individual scenarios are fixed upstream, or no longer a concern, that this buildpack is able to no-op the changes it would make. 

When individual files/dirs are altered via this buildpack, without a connection to why, it'll be pretty tough in future for an end user to know what to remove, and when to remove, from this buildpacks config if an underlying issue is resolved by a framework/runtime.

That said.. I do see that this is still needed.. today.. because this is the simplest/easiest option to fix the permissions issues we see today. The thought that this buildpack can very easily be unintentionally used to undo all the careful preparation made by prior buildpacks, and potentially expose the application to content it should never have been able to read or serve is pretty chilling though. 

Re disabling at the platform level:
I don't think disabling it at a platform level makes sense.. you either need the changes to make the application work, or you don't. If a platform chose to disable this buildpack, it would just become a less useful platform.. sure, maybe it's more secure, but if its also no longer able to build a working application, wouldn't someone just switch to something else until it works? 

Re tracking actions via a label
I don't like the idea of unbounded labels either.. just setting something with a summary of how many files/dirs were altered, might give an indication of how many things were affected.. Also the build log should explicitly log out the files/dirs altered saying from what, to what permissions. This will at least allow the log to be used for diagnosis of how files ended up with unexpected permissions in the future (Imagine you are a buildpack author handling an Issue raised that an app built with your buildpack isn't functioning correctly, how would you know this buildpack had potentially altered the permissions of the files you created.. etc)


Overall, the main thing here is that the documentation for this should be explicitly clear regarding the potential risks of using this buildpack. This should include the RFC itself containing a clear warning about the potential consequences of applying permissions recursively. 

Understood that this is just a tool, and it's how you use it that makes it dangerous, not the tool itself..  However, I've seen too many 'fix it by `chmod -R 755 /directory` ' hack-solutions that end up causing disaster for production containers.

anthonydahanne:
Working Group comments
* Add big warnings to the RFC

text/utilities/0000-chmod.md
dmikusa: >My main concern is that with this being a standard offering, that you'd see a propagation of 'just chmod everything 755' style 'fixes' starting to be applied by users who discover this 'makes it work' without understanding what they've done.

I disagree. When things generally work out of the box, you don't need this and won't use this. Permissions are not an issue *most* of the time with buildpacks, so most of the time this won't even be used. No one can see the future, but that's my theory.

>In an ideal world, we shouldn't need this at all. The relevant buildpacks should be able to determine the appropriate permissions for their files to ensure that the application is able to function at runtime. If this is not possible today, then we really need to start identifying those scenarios, and working out how to deal with them.

This can't happen 100% of the time. There are edge cases where the buildpack cannot know the permissions to apply. If a user ships a script they want to shell out and run, we can't mark that executable for them.

The same happens with the dist-zip buildpack. That format uses scripts to execute the application and there are some cases where only the user really knows how the permissions should be set (it's further complicated by using a zip format that doesn't persist permissions).

> The thought that this buildpack can very easily be unintentionally used to undo all the careful preparation made by prior buildpacks, and potentially expose the application to content it should never have been able to read or serve is pretty chilling though.

I think this is a bit sensational. Buildpacks make an effort so that things just work and are secure out of the box, but there are still plenty of ways for a dev to mess things up. Devs can include inline buildpacks, or just make their own buildpacks. Most build tools also offer a way to run arbitrary commands at build time (Maven, Gradle, npm, composer, etc...).

The nice thing about this approach is that we can offer some controls, like logging what get's altered and when things get altered, as well as affixing labels to the generated images so that it's clear when something's been altered. That's extremely useful as a platform operator because you can now audit when someone's doing this or even block it by prohibiting containers with those labels from being run. That's way better than the alternative approaches users would have to take without this buildpack, and way better than the standard Dockerfile approach.

> Re disabling at the platform level:

The point is that a platform operator can make this choice if they want to prevent devs from doing this. It's additional flexibility and control, which is something that buildpacks offer above and beyond Dockerfiles.

> Overall, the main thing here is that the documentation for this should be explicitly clear regarding the potential risks of using this buildpack. This should include the RFC itself containing a clear warning about the potential consequences of applying permissions recursively.

I don't mind adding warnings to the documentation. It's important to let users know when to be wary but we shouldn't put anything so over the top as to scare users off. We should also be sure to document the labels/audit logging/etc.. so that they can see how this option can effectively be used.

text/utilities/0000-chmod.md
anthonydahanne: ok we keep setting the environment variable as a mandatory step; it will make it more foolproof.

anthonydahanne: I'll actually add a note about this

text/utilities/0000-chmod.md
anthonydahanne: Added big warnings as a first chapter.

BarDweller (approved):
Happy with the warning =)

PR #296, opened 2023-09-20
RFC: Codify the Creation of New Maintainer Groups

ForestEckhardt:
[Readable](https://github.com/paketo-buildpacks/rfcs/blob/ForestEckhardt-patch-1/text/0000-new-maintainer-group-process.md)


loewenstein (approved):

text/0000-new-maintainer-group-process.md
```diff
@@ -0,0 +1,22 @@
+# Codify the Creation of New Maintainer Groups
+
+## Summary
+
+This RFC is intended to define processes for the creation of new maintainer groups.
+
+## Motivation
+
+Currently the creation of new maintainer groups is an ad hoc process that has to be hashed out every time a new group is requested. This means that intervention from the Steering Committee is often needed in situations where they should not need to weigh in. It also means that there is no standard record for the introduction of new groups meaning that the reasoning behind the formation of these new maintainer groups is lost. Codifying this process should make it easier to identify when new maintainer groups should be formed and make the process of doing so much easier. It should also give approval power to the stakeholders that will be directly affected by the introduction of new maintainer groups, namely the existing maintainer groups as opposed to the Steering Committee.
```
dmikusa, line 9: Has there been a case where this is an issue? I'm not really sure I see the problem that this addresses. We create groups pretty infrequently, and from what I can recall, that's been done via RFC. Is there a case where an official group was formed without an RFC?

text/0000-new-maintainer-group-process.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Implementation
+
+The following are the 3 scenarios in which a new maintainer group would be formed and the procedures for forming each group type.
+
+### New Maintainer Group for New Repositories
+When proposing a new project that the proposer feels does not fall under the domain of an existing maintainer group they can suggest creating a new maintainer group. As part of the proposal, members of the initial maintainer group should be suggested or a request for volunteers should be open by the proposer of the new project. This group will then be approved by the Steering Committee to be the initial set of maintainers.
```
dmikusa, line 16:
```suggestion
When proposing a new project that the proposer feels does not fall under the domain of an existing maintainer group they can suggest creating a new maintainer group. As part of the proposal, members of the initial maintainer group should be suggested or a request for volunteers should be open by the proposer of the new project. This group and the proposed initial set of members will then be reviewed and either approved or rejected by the Steering Committee.
```

text/0000-new-maintainer-group-process.md
```diff
@@ -0,0 +1,22 @@
...
+
+### New Maintainer Group for Covering Repositories that Cross Maintainer Groups
+When a project that requires coordination of components across multiple repositories that fall under the responsibility of existing maintainer groups but there are people that would like to be able to maintain the project as a whole.The process will involve opening a proposal to create a new maintainer group and outline the motivation for creating a new maintainer group and which repositories this new maintainer group would become responsible for. As part of the proposal, members of the initial maintainer group should be suggested or a request for volunteers should be open. Then the current maintainers of all repositories that are part of the proposal will have to vote on whether or not they want to share control over the repositories proposed and whether or not the maintainers proposed are suitable.
```
dmikusa, line 22:
```suggestion
### New Maintainer Group for Covering Repositories that Cross Maintainer Groups
When a project that requires coordination of components across multiple repositories that fall under the responsibility of existing maintainer groups but there are people that would like to be able to maintain the project as a whole.The process will involve opening a proposal to create a new maintainer group and outline the motivation for creating a new maintainer group and which repositories this new maintainer group would become responsible for. As part of the proposal, members of the initial maintainer group should be suggested or a request for volunteers should be open. Then the steering committee and current maintainers of all repositories that are part of the proposal will have to vote on whether or not they want to share control over the repositories proposed and whether or not the maintainers proposed are suitable.
```

I don't totally follow the use case here. Is this when you want to consolidate two separate groups into one group?

Same comment as before. I think steering committee should be part of all group-related decisions.

text/0000-new-maintainer-group-process.md
ForestEckhardt: I think that most of our maintainer groups were created before a formalized RFC process existed. I have had a couple of people approach me asking if there was an official process. I figured that formalizing the process made sense and would clear up question of voting going forward.

text/0000-new-maintainer-group-process.md
ForestEckhardt: > I don't totally follow the use case here. Is this when you want to consolidate two separate groups into one group?

This spawned out the UBI workflow. Ozzy expressed interest in becoming a maintainer of all of the components that make up the UBI workflow however those components span multiple maintainer teams. So I thought it might makes sense to make something like the UBI Maintainer team and give them joint access to UBI related repositories.

> Same comment as before. I think steering committee should be part of all group-related decisions.

I am not sure that I understand why you would want the Steering Committees approval here as it seems like something that could be hashed out among the existing maintainers. So I am curious what benefit you see to the Steering Committee being involved

text/0000-new-maintainer-group-process.md
ForestEckhardt: I think that makes sense to me. In my head that was implied but I think it is fair to lay that out.

text/0000-new-maintainer-group-process.md
```diff
@@ -0,0 +1,22 @@
...
+
+### New Maintainer Group for Existing Repositories that are Maintained by One Existing Group
+When the repositories exist and are already under the control of an existing maintainer team but there is a desire to have a new maintainer group take over or co-maintain a portion of the existing repositories. The process will involve opening a proposal to create a new maintainer group and outline the motivation for creating a new maintainer group and which repositories this new maintainer group would become responsible for. As part of the proposal, members of the initial maintainer group should be suggested or a request for volunteers should be open. Then the current maintainers will have to vote on whether or not they want to either cede/share control over the repositories proposed and whether or not the maintainers proposed are suitable.
```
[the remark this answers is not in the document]
ForestEckhardt: > Also, I feel very strongly that any group decisions should be also approved by the steering committee. The committee is responsible for guiding the project and part of that is the organization of groups. The steering committee needs to review the request and make sure it fits within the context of the entire project, not just want the specific group wants to do.

I am not deeply opposed to having the Steering Committee have approval on these proposals. I was trying to keep the Steering Committee hands off in a situation that I felt could be up to maintainer groups to decide. I am just a little curious of how you think the independent introduction of a maintainer group could drastically shape the direction of the project.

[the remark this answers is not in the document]
dmikusa: My rationale is this:

1. Having a SC review, means we can help to ensure consistency across the project and sanity check requests. We have a view of the entire project, and not every maintainer has the same perspective.

2. I think it would be good to have a sanity check on the list of proposed members as well. I'm not sure we'd care about the individual members, but we can ensure that the introduction of a new group meets standards and requirements like having enough members to be sustainable.

3. I think the review of these should be very minimal demand on the SC. You're going to read the proposal. It's going to meet the requirements and we 👍 or it's not and we'll reject with feedback.

text/0000-new-maintainer-group-process.md
dmikusa: I don't personally like the idea of having groups that overlap. I think it has the potential to create confusion about ownership or introduce ways for conflicts that are difficult to resolve, like if there are two groups that are maintainers of a repo and they each have different visions about how something should be done, or if there are two groups that are maintainers and one group does not want an individual as a new member, but then the other group takes on that person as a new member. It adds complications.

I think it would also make voting calculations more complicated. We require a percentage of maintainers to vote for an issue before it's approved. Right now, that means you look at the owning team of the RFC and use the team size to calculate the required number of votes. With this proposal, it's not that simple. You need to do more assessment in terms of what teams must vote and what the required number of positive votes is to pass an RFC. Because non-votes essentially count as No votes, that could make passing RFCs harder as well.

I think we can always evaluate our group structure. If over time, it doesn't represent the ownership of repos then we can adjust things. Otherwise, I think users just need to become members of multiple groups. It is a little more work, but it allows each group to fully own membership and the owned code/repos.

text/0000-new-maintainer-group-process.md
dmikusa: OK. Sounds reasonable. I like the idea of having transparent processes.

I would suggest that we call out where this process is going to live. Probably add something to the Implementation section that says "Add a page to the website to document this process". People shouldn't have to dig through RFCs to understand the process.

text/0000-new-maintainer-group-process.md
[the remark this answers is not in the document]
loewenstein: My rational for approving the RFC as is was, that the group making this decision would already have the right to distribute the access permissions by doing a local vote. So I thought this can really just be left to the maintainer groups that are affected.

But if we keep SC involvement low, I am fine with that as well.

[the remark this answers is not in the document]
dmikusa: Yes, I can definitely appreciate keeping our involvement low. I think it's worth us putting in a small bit of time here though.  We can make sure that someone is evaluating decisions and considering the big picture, it'll help with overall project consistency.

Think of the TEAMS.md doc in the community repo. It's already pretty big. If teams can split and make more teams that *could* make working with and contributing to the project more confusing. We should evaluate team decisions critically & I think that's what the SC can bring to the picture.

text/0000-new-maintainer-group-process.md
ForestEckhardt: I am ok with removing this. Maybe it is something we look back on, maybe not.

loewenstein (approved):
@ForestEckhardt @dmikusa I am ok, if you are ok. But do we really want to create even more RFCs? I think we have too many already - feels like RFCs are our hammer and everything that needs some members to agree on something is hence a nail?

PR #297, opened 2023-09-26
RFC: Standardize Paketo Dependency Metadata Format

ForestEckhardt:
[Readable](https://github.com/paketo-buildpacks/rfcs/blob/standard-dependency-metadata-format/text/0000-standard-dependency-metadata-format.md)


text/0000-standard-dependency-metadata-format.md
```diff
@@ -0,0 +1,48 @@
...
+    uri = "<URI for information of license>" #optional
+```
+
+**Note:** Both the `distros` and `licenses` fields are optional, however if they are given then the non-optional components of them must be set.
+
+## Prior Art
+- The layout of distributions is pulled from the [Buildpacks Spec](https://github.com/buildpacks/spec/blob/main/buildpack.md#buildpacktoml-toml).
+
+## Unresolved Questions / Bikeshedding
+- Should we make a new tooling repository as part of this initial RFC or should that wait for implementation?
```
dmikusa, line 46: I think we should leave tooling out.

text/0000-standard-dependency-metadata-format.md
```diff
@@ -0,0 +1,48 @@
...
+- Should this include a buffer for stacks or should the stack removal be the forcing function that forces this conversion?
```
dmikusa, line 47: 100% needs a migration path. If not, it leaves an area undefined that will need to be defined for a while still. I'd be OK with starting that out of the gate as "deprecated" pending removal upstream, or something like that.

text/0000-standard-dependency-metadata-format.md
```diff
@@ -0,0 +1,48 @@
...
+- Does this require the creatation of our own Paketo Spec or some other API style document or is this RFC sufficent?
```
dmikusa, line 48: I think it depends on the goal. Are we going to ever certify someone as "Paketo compliant"? If so, then we should start working towards an official spec. If not, if the intent is to guide implementation then I think the RFCs are OK. We do maintain a list of the "active" RFCs, so someone interested in compatibility can reference them.

I don't think we need to put this into the documentation until it affects normal end users. I think it's a reasonable expectation for buildpack authors to sometimes need to refer to an RFC directly.

text/0000-standard-dependency-metadata-format.md
```diff
@@ -0,0 +1,48 @@
...
+checksum = "<dependency algo:checksum>"
+id = "<dependency ID>"
+os = "<dependency compatible OS>"
+uri = "<dependency URI>"
+version = "dependency version"
+
+cpes = [ "<dependency cpe>" ] #optional
+name = "<dependency name>" #optional
+purl = "<dependency purl>" #optional
+strip-components = <number of directories to strip off dependency artifact> #optional
```
dmikusa, line 29: More of an FYI, but this is not a setting supported by libpak. Idk if we'd actually need/use this, but if it's optional then I'm OK including it.

text/0000-standard-dependency-metadata-format.md
```diff
@@ -0,0 +1,48 @@
...
+The following in the proposed metadata format:
+
+```toml
+[[metadata.dependencies]]
+arch = "<dependency compatible architecture>"
+checksum = "<dependency algo:checksum>"
+id = "<dependency ID>"
+os = "<dependency compatible OS>"
+uri = "<dependency URI>"
+version = "dependency version"
```
dmikusa, line 24: I'm worried that this will get tedious. It is entirely possible to have binaries, like JAR files, that are not architecture/OS/distro dependent. The distro is a list so we could have multiple, it's also option so we can leave it off. Arch and OS are required and individual.

Thus how would I represent something that runs on all arch/os? Or what if it runs on two arch's like ARM64 and AMD64, do I then need to duplicate the metadata and include it twice?

Off the top of my head...

- If we have arch and OS as optional, then we could specify that the lack of them indicates support for all
- If we make them arrays, then we can specify support for multiple items.

I'm open to other ideas. Thoughts?

text/0000-standard-dependency-metadata-format.md
```diff
@@ -0,0 +1,48 @@
...
+strip-components = <number of directories to strip off dependency artifact> #optional
+
+    [[metadata.dependencies.distros]] #optional
+    name = "<compatible OS distribution name>"
+    version = "<compatible OS distribution version>" #optional
+
+    [[metadata.dependencies.licenses]] #optional
+    type = "<license of dependency>"
+    uri = "<URI for information of license>" #optional
+```
```
dmikusa, line 38: We just added source and source sha fields. I believe packit has them already. Should we include those in this RFC?

text/0000-standard-dependency-metadata-format.md
anthonydahanne: >If we have arch and OS as optional, then we could specify that the lack of them indicates support for all

+1

text/0000-standard-dependency-metadata-format.md
ForestEckhardt: Oh that is a good call I forgot about that.

text/0000-standard-dependency-metadata-format.md
```diff
@@ -0,0 +1,48 @@
...
+os = "<dependency compatible OS>"
+uri = "<dependency URI>"
+version = "dependency version"
+
+cpes = [ "<dependency cpe>" ] #optional
+name = "<dependency name>" #optional
+purl = "<dependency purl>" #optional
+strip-components = <number of directories to strip off dependency artifact> #optional
+
+    [[metadata.dependencies.distros]] #optional
```
anthonydahanne, line 31: why not `metadata.dependencies.distributions` ? I'm not sure `distros` is an abbreviation that makes sense in a field name

anthonydahanne: well, this ship has apparently sailed already..
https://github.com/buildpacks/spec/blob/main/buildpack.md#buildpacktoml-toml
nevermind 🤷‍♂️

text/0000-standard-dependency-metadata-format.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Motivation
+
+There are several projects, such as the introduction of ARM or the removal of stacks, that are being discussed that would force Paketo to update the metadata for dependencies. It seems prudent that if we are going to have to do a large update to dependency metadata that it might also be a good time for us to converge as a project on the dependency metadata format that we should be using going forward.
+
+This could also be a good first step in converging on shared tooling. By having a shared dependency metadata format we would have a good common convergence point to begin building universal tooling for the Paketo project. 
+
+## Implementation
+
+The following in the proposed metadata format:
```
c0d1ngm0nk3y, line 15:
```suggestion
The following is the proposed metadata format:
```

text/0000-standard-dependency-metadata-format.md
```diff
@@ -0,0 +1,48 @@
...
+[[metadata.dependencies]]
+arch = "<dependency compatible architecture>"
+checksum = "<dependency algo:checksum>"
+id = "<dependency ID>"
+os = "<dependency compatible OS>"
+uri = "<dependency URI>"
+version = "dependency version"
+
+cpes = [ "<dependency cpe>" ] #optional
+name = "<dependency name>" #optional
```
c0d1ngm0nk3y, line 27:
```suggestion
name = "<dependency name>" #optional
deprecation_date = "<dependency eol>" #optional
```

text/0000-standard-dependency-metadata-format.md
c0d1ngm0nk3y, line 48:
```suggestion
- Does this require the creation of our own Paketo Spec or some other API style document or is this RFC sufficient?
```

c0d1ngm0nk3y: I think that might be quite nice to have this. It would be easy to find the current metadata format (and not bury it in some RFC) und future changes will be easy to track.

text/0000-standard-dependency-metadata-format.md
modulo11: The `deprecation_date` is also missing.

text/0000-standard-dependency-metadata-format.md
```diff
@@ -0,0 +1,48 @@
...
+purl = "<dependency purl>" #optional
+strip-components = <number of directories to strip off dependency artifact> #optional
+
+    [[metadata.dependencies.distros]] #optional
+    name = "<compatible OS distribution name>"
+    version = "<compatible OS distribution version>" #optional
+
+    [[metadata.dependencies.licenses]] #optional
```
modulo11, line 35: I did a quick [comparison](https://gist.github.com/modulo11/1e8a5ef74f6408ddba40c209521a8765) which showed that the `licenses` field is of different type between libpak and packit. While the approach chosen in the RFC would be also my prefered way, it leads to additional (one-time) work on maintainers of packit based buildpacks.

text/0000-standard-dependency-metadata-format.md
dmikusa: > I think I would ask this question a different way - how are you planning on handling the stack removal (replacing it with os etc above).

Not sure if this was addressed to me or @ForestEckhardt.

I haven't thought about it too much. The upstream is not ready for it yet, and it's missing some functionality we need (there's another RFC proposed to fill this gap, but we'd need this or a similar RFC to remove stacks).

I would say that my goal is to limit repetition and to keep the TOML reasonable to read. That was my question here. The way things are proposed currently, it would require a lot of repeated TOML for various dependencies. We need to find a more efficient way to represent that in the metadata. My $0.02.

text/0000-standard-dependency-metadata-format.md
dmikusa: I wonder if it makes sense to include the logic around this field?

- no deprecation_date means we don't know/can't find the date
- a depreciation_date means it ends on the given date.
- if the depreciation_date is in the past, it's EOL
- if the depreciation_date is in the future, it's not yet EOL.

Any states or implied information I'm forgetting?

text/0000-standard-dependency-metadata-format.md
```diff
@@ -0,0 +1,48 @@
...
+There are several projects, such as the introduction of ARM or the removal of stacks, that are being discussed that would force Paketo to update the metadata for dependencies. It seems prudent that if we are going to have to do a large update to dependency metadata that it might also be a good time for us to converge as a project on the dependency metadata format that we should be using going forward.
+
+This could also be a good first step in converging on shared tooling. By having a shared dependency metadata format we would have a good common convergence point to begin building universal tooling for the Paketo project. 
+
+## Implementation
+
+The following in the proposed metadata format:
+
+```toml
+[[metadata.dependencies]]
```
mhdawson, line 18: This is the example I mentioned in terms of how we used dependencies in the ubi-nodejs-extension - https://github.com/paketo-community/ubi-nodejs-extension/blob/7caa04ee9a7e22dd1948c62fcaca03679be0db99/extension.toml#L15

```

  [[metadata.dependencies]]
    id = "node"
    name = "Ubi Node Extension"
    stacks = ["io.buildpacks.stacks.ubi8"]
    source = "paketocommunity/run-nodejs-18-ubi-base"
    version = "18.1000"
```

In the case of of the Node.js extension, dependencies are currently handled as follows:
- build - list of os packages to installed baked into extension
- run - pre-bundled into run image onto which application is 'switched'. In this case we use the info in the dependencies to decide which run image to use based on the Node.js Version requested.  This is similar to how the existing buildpacks use the metadata to find the Node.js package to use, except that we set the source to the name of a docker image instead of a tar.gz to install.

mhdawson: The was we use the depdencies, it would be good if `checksum` was optional.

loewenstein: I wonder if we could achieve the same behaviour for ubi extensions with something more explicit and less ad hoc... I.e. for "normal" dependencies it seems legit to require a checksum. I don't know, maybe make source and checksum one way, but offer another for the *virtual* (for lack of better word) dependencies that will be provided by the special run image?

dmikusa: I second what @loewenstein is saying. This feels like a bit of a different use case. The metadata up for discussion here has always been associated with buildpacks that install dependencies directly. I'm sure there are assumptions built into that process and generalizing the metadata to support other installation mechanisms is going to complicate things or possibly create bugs due to age-old assumptions. While this seems like a small thing, I think it's also going to cause a lot of additional work to be required, because of past assumptions.

I 100% agree that this is a problem though. There's no way for a buildpack author to really know what's going to be in an image that was modified by an extension. I had hoped that upstream would provide some mechanism for this, but I'm not aware of one. In the absence of something upstream, I'm not opposed to creating our own solution (and possibly pushing to have that made the upstream solution). I'd want to talk more with upstream before we do anything though, to confirm they're not planning something, and also make that plan its own RFC.

text/0000-standard-dependency-metadata-format.md
ForestEckhardt: Yes, I think that it will require a little bit of work for packit but I think it is ultimately a relatively small amount of pain.

text/0000-standard-dependency-metadata-format.md
mhdawson: I agree it is different in some ways, but it is similar in the search for a match of the dependency, currently we can re-used the same supporting methods versus having to duplicate it. See below:

https://github.com/paketo-community/ubi-nodejs-extension/blob/7caa04ee9a7e22dd1948c62fcaca03679be0db99/generate.go#L62
```
				dependency, err := dependencyManager.Resolve(extensionFilePath, entry.Name, version, context.Stack)
```

mhdawson: Maybe there could be the concept of different types of dependencies, each with their own required fields, and with a common set of required fields that are used by the Resolve?

text/0000-standard-dependency-metadata-format.md
ForestEckhardt: This will be completely optional and can be ignored.

text/0000-standard-dependency-metadata-format.md
ForestEckhardt: I think that I am more than happy with making the OS and Arch optional (and will do so in my next commit)

text/0000-standard-dependency-metadata-format.md
ForestEckhardt: I am going to add `source` and `source-checksum` as well as `deprecation-date` as optional fields.

text/0000-standard-dependency-metadata-format.md
ForestEckhardt: I think that logic checks out. I am going to add this as `deprecation-date` (changing to _ to - to reflect the other variables).

text/0000-standard-dependency-metadata-format.md
dmikusa: tl;dr - I think this all merits more discussion and I don't want to hold up this RFC on that discussion. It's too much scope creep IMHO.

Breaking it down more...

- You can make a case that the dependencies we install are similar to what's being installed through the extension. 
- You could also make a case they are different, specifically that what gets installed from a dependency is not the same as what gets installed via a package (sometimes one dependency is split into multiple packages, debug, source, docs, etc.). 
- There are also differences in the source of those dependencies, and do we need to make distinctions about that? 
- There's also the build plan, and can we use that more effectively to communicate buildpack needs?
- You could also make a case that what both buildpacks and extensions care about is certain tools or binaries that get installed, not the dependency or package itself (sometimes these are the same, but not always). i.e. I want `bash` or I want `javac` and I don't care what package gives me that. 

I think it's worth rethinking this process and what buildpacks need, rather than trying to inject this into the existing process.

text/0000-standard-dependency-metadata-format.md
loewenstein: Actually, the stack removal has been done as breaking change with [Buildpack API 0.10](https://github.com/buildpacks/spec/releases/tag/buildpack%2Fv0.10). I would assume we should tackle this rather sooner than later.

text/0000-standard-dependency-metadata-format.md
loewenstein: What I would do immediately, as part of the implementation of this RFC is creating a Go package to streamline dependency metadata handling for both `libpack` and `packit`.

text/0000-standard-dependency-metadata-format.md
```diff
@@ -0,0 +1,67 @@
...
+id = "<dependency ID>"
+uri = "<dependency URI>"
+version = "dependency version"
+
+arch = "<dependency compatible architecture>" #optional
+cpes = [ "<dependency cpe>" ] #optional
+eol-date = "<dependency eol>" #optional
+name = "<dependency name>" #optional
+os = "<dependency compatible OS>" #optional
+purl = "<dependency purl>" #optional
```
loewenstein, line 37: Why do we support multiple `cpes`, but only a single `purl`?

modulo11: Makes sense to align this with `cpes` even if only one entry is added for most dependencies.

dmikusa: Pardon this question, but does anyone actually use PURLs?  I know scanners and users do look at CPEs. I've never heard anyone mention PURL besides asking what it is. Before we roll out the red carpet for PURL, I just want to get a feel for how much people use it.

It's not a huge pain to support, but when you talk about changing things with it, that's going to require work. Makes me want to understand the usage first before committing to changing things.

loewenstein: For example https://osv.dev uses purls. More broadly speaking, we should adapt to whatever e.g. security researchers use for reporting vulnerabilities. For a starter, I think that means we should allow a package to have multiple names. I don't see any further package naming scheme coming besides CPE and pURL - but if a relevant new scheme comes up, we should adopt that as well.

Granted, it also depends on using the "correct" names, which might turn out to be difficult as well. Most of the current pURLs used for example are unlikely to match vulnerabilities. I think I brought that up in the past and @ForestEckhardt  (and @sophiewigmore?) wanted to maybe spend some thought and look into that topic.

sophiewigmore: Current pURL implementation context: When added pURLs we were trying to cover all of our bases while we were still trying to figure out emerging trends in how people track packages/CVEs. Our best guess at that point was to leverage https://github.com/package-url/packageurl-go while getting dependency metadata together.

Fast forward to now - I have heard 0 feedback on if they're useful fields or if any useful metadata can be gleaned from the pURLs we provide now. We haven't invested any time as far as I know into getting an answers in this area. I think its likely still a useful field to provide for the same reasons Jan mentioned.

dmikusa: Unless someone is going to commit to going through the metadata and cleaning up the PURLs, I feel like we should drop that metadata. 

My understanding of why we adopted them in the first place matches what @sophiewigmore mentioned. We were basically throwing things out there to see what would stick. PURLs don't seem to have stuck, so we should drop the PURL field rather than continue publishing something misleading and likely incorrect.

That is unless someone feels strongly enough about keeping it to review and update all the PURLs that we publish.

loewenstein: I agree that the worst would be to stick to the existing `generic` purls that very likely don't help in finding vulnerabilities.

There seem to be [some adopters](https://github.com/package-url/purl-spec#users-adopters-and-links) though, amongst them Google and Github, I wonder if we could also opt to keep purls (plural!) in as optional metadata (and making CPEs optional as well maybe?) - but remove all the `generic` purls we have currently assigned.

One example would be the Apache Tomcat buildpack. While the distribution that gets installed is different from the Tomcat Jar that is available as Maven GAV - there can be vulnerabilities for it found in osv.dev based on the `maven` purl and it might make sense for this buildpack to include the purl as well for this specific reason.

ForestEckhardt: I have made the field `purls` and made it so that multiple PURLS can be provided.

text/0000-standard-dependency-metadata-format.md
ForestEckhardt: I think that I agree with the idea that we should maybe push this down the road and worry about this at a later date. I would like to see more mature implementation of this before committing to a defined schema.

text/0000-standard-dependency-metadata-format.md
ForestEckhardt: I think that I am perfectly happy with leaving the tooling out the the RFC if have a standing agreement that we should make a unified library to process the dependency metadata.

text/0000-standard-dependency-metadata-format.md
ForestEckhardt: I am not entirely sure what we should do about a "spec". It feels like something that is needed as a result of this RFC but I am not sure that this RFC is the right place to hash out what we should do and where it should live. Should I remove this as a concern for now and open a separate discussion somewhere on what we should do to document these kinds of standards within the Paketo community? I would like to try and keep this rolling forward for the time being so that we can continue to advance both the multi-arch as well as the decoupled dependencies buildpacks rfcs forward.

dmikusa: +1 for splitting out a spec discussion. It's a good discussion, one we need to have. It'll take a lot of work to do right. It shouldn't hold up this or other efforts. My $0.02.

text/0000-standard-dependency-metadata-format.md
loewenstein: What better way to have a standing agreement than putting this int the RFC though? We don't need to name a repo or fix that we would enforce common libraries and tooling. We could instead just make it clear that following the specification will be the minimum requirement for Paketo buildpacks and also provide guidance that - when feasible and meaningful - buildpacks should align the work on libraries and tooling for handling dependencies.

We might still have a provision to cover the concerns raised because of UBI buildpacks / extensions being different, when it comes to dependencies and this case might need further, but separate considerations.

text/0000-standard-dependency-metadata-format.md
loewenstein: It doesn't have to block the merge of the RFC though. And, given the "informal" description that is part of the RFC, it can be adopted while we figure out a process for writing Paketo specific specifications.

dmikusa (approved):

anthonydahanne:
Good to merge?

loewenstein (approved):

loewenstein:
@ForestEckhardt Are we waiting for more feedback or just forgot to merge this?

PR #299, opened 2023-10-12
Addendum to dependency mapping RFC

sophiewigmore:
## Summary

Allow dependency mappings to be in format `algorithm-hash` in addition to `algorithm:hash` in order to work on Kubernetes. See for more information https://github.com/paketo-buildpacks/packit/issues/521

## Checklist

### DIFF
--- a/text/0010-dependency-mappings.md
+++ b/text/0010-dependency-mappings.md
@@ -14,7 +14,7 @@ Consumers of Paketo buildpacks may be building in an environment where certain d
 If one or more bindings of kind or type `dependency-mapping` are present, then
 a buildpack should use the bindings in its dependency resolution process. The
 dependency's digest, included as `checksum` in the form of
-`<algorithm>:<hash>` or `<algorithm>-<hash>`, (or `sha256` in the form of
+`<algorithm>:<hash>` or `<algorithm>_<hash>`, (or `sha256` in the form of
 `<hash>`), should be used to lookup the alternatively mapped URI.
 
 If the digest is provided in the form of `<hash>`, it will be assumed to be of algorithm type `sha256`.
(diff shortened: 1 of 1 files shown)


ForestEckhardt (approved):

text/0010-dependency-mappings.md
```diff
@@ -13,16 +13,16 @@ Consumers of Paketo buildpacks may be building in an environment where certain d
 ### Discovery
 If one or more bindings of kind or type `dependency-mapping` are present, then
 a buildpack should use the bindings in its dependency resolution process. The
-dependency's digest, included as `checksum` in the form of `<algorithm>:<hash>`
-(or `sha256` in the form of `<hash>`), should be used to lookup the alternatively
-mapped URI.
+dependency's digest, included as `checksum` in the form of
+`<algorithm>:<hash>` or `<algorithm>-<hash>`, (or `sha256` in the form of
+`<hash>`), should be used to lookup the alternatively mapped URI.
```
dmikusa, line 18: tl;dr - I'm fine with this approach. 

Question: what about defining a configurable delimiter instead of adding a second delimiter to the RFC? `BP_BINDING_ALGORITHM_DELIMITER` or something like that? It could default to `:` which is the present behavior and then could be changed to any other value that is useful.

Two is probably enough that we won't have to revisit this again, but making it configurable would guarantee we don't have to rehash this topic again. It also follows similar patterns in the Cloud-Native buildpacks spec.

sophiewigmore: We discussed this at the working group meeting this week so I'll summarize my stance - I think I lean towards having the simplest possible user experience.  While you're right, it would be good to avoid having this discussion again in the future by making it configurable, I think that the two options probably will suffice as you said. In general, I'd love for this feature to just work for users without needed to add additional configuration, especially because dependency mappings can already be confusing as is. I'd prefer avoiding any additional user-provided configuration.

ChuckQuinnIV:
Fair point - do we want to adjust this to use `_` instead of `-`? Underscore should also be valid in both Kubernetes & Windows

sophiewigmore:
@ChuckQuinnIV that works - looks like it should work on Windows. @paketo-buildpacks/tooling-maintainers any concerns with using a `_` instead of `-`?

robdimsdale:
> @ChuckQuinnIV that works, it looks like it should work on Windows. @paketo-buildpacks/tooling-maintainers any concerns with using a `_` instead of `-`?

If it works then I'm happy. I don't see any obvious concerns (assuming `_` is not _also_ used in algorithms 😅 )

sophiewigmore:
@robdimsdale not that I've seen

robdimsdale (approved):

dmikusa (approved):

sophiewigmore:
@paketo-buildpacks/steering-committee am I good to merge this in?

loewenstein (approved):

PR #300, opened 2023-10-31
Add RFC to discourage the use of old bionic builders

c0d1ngm0nk3y:
[readable](https://github.com/paketo-buildpacks/rfcs/blob/2b9f8b3c41827be4e2d4a55f0c214ef6dd1574e5/text/0000-disccourage-bionic-builders.md)

## Summary
Users are still using `bionic` builders, we should make them aware that they are using a unmaintained stack.

## Use Cases
Using `paketobuildpacks/builder:base` sounds OK and there is no indication that `bionic` is used and this is not kept up to date.

## Checklist

### DIFF
--- a/text/0000-disccourage-bionic-builders.md
+++ b/text/0000-disccourage-bionic-builders.md
@@ -1,0 +1,39 @@
+# Users should be discouraged of still using bionic builders
+
+## Summary
+
+User might still be using `bionic` based builders (e.g. `paketobuildpacks/builder:base`) without realizing that it reached end of life. The user should be made aware that this builder is not maintained anymore.
+
+## Motivation
+
+Since the names of the `bionic` builders have no `bionic` in the name (e.g. `paketobuildpacks/builder:base`), it is not unlikely that users might not be aware that they are using a stack thit is end of life for a couple of month.
+
+So we should make the users aware about this fact, so that they can decide if they want to switch to `jammy` or keep using `bionic`.
+
+Since the switch to the `jammy` builder might need effort, it is probably benefical to point the users already in that direction.
+
+## Detailed Explanation
+
+We could add a new buildpack to the builders that will cause every build to fail unless the environment variable `BP_USE_DEPRECATED_STACK` is provided. With that, we would force the user to either move to a maintained builder or to make the decision to keep using the builder by adding this environment variable.
+
+## Rationale and Alternatives
+
+* The naming scheme of the `bionic` builders somehow implies that the builder keeps up to date. So it might even be risky to keep users use a stack after end of life.
+* We could also remove the tags to make it unconvenient to keep using the builders. If there is a necessity, the builder could still be used by digest.
+* We could only warn when using the builder, but there would be a risk of this warning not being read.
+
+## Implementation
+
+{{Give a high-level overview of implementation requirements and concerns. Be specific about areas of code that need to change, and what their potential effects are. Discuss which repositories and sub-components will be affected, and what its overall code effect might be.}}
+
+{{THIS SECTION IS REQUIRED FOR RATIFICATION -- you can skip it if you don't know the technical details when first submitting the proposal, but it must be there before it's accepted.}}
+
+## Prior Art
+
+This is special because the names for the `bionic` builders (`paketobuildpacks/builder:tiny`, `paketobuildpacks/builder:base`, `paketobuildpacks/builder:full`) appear to be agnostic. The `jammy` builders do not have the same problem anymore.
+
+## Unresolved Questions and Bikeshedding
+
+{{Write about any arbitrary decisions that need to be made (syntax, colors, formatting, minor UX decisions), and any questions for the proposal that have not been answered.}}
+
+{{REMOVE THIS SECTION BEFORE RATIFICATION!}}


ForestEckhardt:
During the working group meeting, there was a discussion that this functionality might be better handled higher up in the stack and @BarDweller mentioned that he would bring it up in the next buildpacks working group meeting.

One of the ideas that was proposed was to have the CNB project allow us to set a variable when building the builders themselves that would mark them as deprecated. Then it could be up to specific platforms on how they would handle using a deprecated stack (i.e. print a warning, fail the build, mark the resulting image, etc.).

We should discuss the outcome of the upstream conversation in following working group meetings.

c0d1ngm0nk3y:
Mhhh, doesn't sound too pragmatic too me to be honest. I have 2 concerns...

* We have an issue *now* with the `bionic` builders. Later, for the `jammy` builders, I don't see it as a big problem since the name somehow indicates that you are using `jammy`. Whereas `paketobuildpacks/builder:base` sounds like a good idea to use. But it can actually even have some severe consequenses when you look deeper.

* Why should it be up to the platform? I do not see where this complexity will actually help.

My main intention with this issue was to remind users still using `paketobuildpacks/builder:base` that they should be aware of the risks. But let's see if there are actionable outcomes.

dmikusa:
If this is something we could act quickly on, I think it would have merit for the project. There are still a lot of Spring users on the old stack because the default won't change until the next release of Spring Boot, which is in a couple of weeks. If we could inject a message to users, that would be very helpful.

That said, if it takes more than a couple of weeks to make that happen, I think the benefit will drop significantly, as users will start upgrading to the latest Spring Boot and will move to Jammy automatically.

A couple of issues come to mind with injecting a new buildpack:

1. It requires us to publish a new bionic builder to include it, and I'm not sure we can do that anymore (at least not without additional overhead). I believe that the projects have been archived and the pipelines taken down.
2. It would require users to update their stack to get that new change.
3. It wouldn't work if users are overriding with custom sets of buildpacks.

The only thing that comes to mind that we could do without publishing new buildpacks/images would be a brown out. Essentially, we advertise ahead of time that we are temporarily taking down the old bionic builders to alert people. Then for a period of a couple of hours, we make the bionic images unavailable. This will break users builds if they are still using them & hopefully get their attention. After the time window expires, we put the images back and if people keep using them, then that's on them.

c0d1ngm0nk3y:
> 3\. It wouldn't work if users are overriding with custom sets of buildpacks.

It came to me afterwards. That is true :(

>  we are temporarily taking down the old bionic builders to alert people. 

I thought something similar. What about removing the `tag` and communicating that you can still use the `digest` instead. Could even be permanently.

dmikusa:
> I thought something similar. What about removing the tag and communicating that you can still use the digest instead. Could even be permanently.

Oh, pulling the tag is a good idea. I wasn't 100% sure how to temporarily disable the image, but that would work well. I'd say we start with a temporary removal to get people's attention, with possibly a permanent one down the road. That would enable people to keep using it if they really, really need to but definitely requires opt-in.

I don't personally feel like we need an RFC for this approach, but just a vote across the steering committee & maintainers. If the majority are in favor, then we can make that happen. If others feel differently, please let us know.

loewenstein:
> > I thought something similar. What about removing the tag and communicating that you can still use the digest instead. Could even be permanently.
> 
> Oh, pulling the tag is a good idea. I wasn't 100% sure how to temporarily disable the image, but that would work well. I'd say we start with a temporary removal to get people's attention, with possibly a permanent one down the road. That would enable people to keep using it if they really, really need to but definitely requires opt-in.
> 
> I don't personally feel like we need an RFC for this approach, but just a vote across the steering committee & maintainers. If the majority are in favor, then we can make that happen. If others feel differently, please let us know.

Why would you go for a temporal removal first, if our goal is to force opt-in. If that is to make sure that even inattentive users get their pipelines working again eventually? What would then be the condition to later permanently disable?

Maybe we can permanently disable the tags and at the same introduce replacements with an `-unsafe` suffix? That's more convenient than copying a digest, but we don't want to apply too much force that just leads to frustration.

dmikusa:
My thought with a temporary removal first would be to attempt to get people's attention. We'll of course make announcements in advance, but inevitably people won't see our announcements. The temporary brown-out is a way to get more attention. Things go back to the way they were so if it really, really causes issues for someone then they have a little more time to address the problem before we permanently make the change.

If folks feel like that's unnecessary, we could skip it and make a permanent change. I don't have strong feelings one way or the other.

loewenstein:
> My thought with a temporary removal first would be to attempt to get people's attention. We'll of course make announcements in advance, but inevitably people won't see our announcements. The temporary brown-out is a way to get more attention. Things go back to the way they were so if it really, really causes issues for someone then they have a little more time to address the problem before we permanently make the change.
> 
> If folks feel like that's unnecessary, we could skip it and make a permanent change. I don't have strong feelings one way or the other.

I don't have a strong opinion either. Just thought what the difference would be - whoever missed the first call will still be broken by the later permanent removal. But yeah, even a convenient `base-unsafe` tag would need to be set and projects might have an enforced code freeze that makes even something like that difficult.

c0d1ngm0nk3y:
> I don't personally feel like we need an RFC for this approach,...

I agree. Should I just close it?

dmikusa:
Yes, let's close. This is out [plan moving forward](https://blog.paketo.io/posts/paketo-bionic-builder-is-unsafe/).

sophiewigmore:
Update from WG meeting 11/28 - attendees are good to close this RFC out

ForestEckhardt:
@c0d1ngm0nk3y Are you ok with closing out this RFC?

c0d1ngm0nk3y:
> @c0d1ngm0nk3y Are you ok with closing out this RFC?

Yes :)

PR #307, opened 2024-05-15
Proposal to add OCI Image Annotations on Buildpacks

candrews:
## Checklist

### DIFF
--- a/text/0000-oci-image-annotations.md
+++ b/text/0000-oci-image-annotations.md
@@ -1,0 +1,48 @@
+# OCI Image Annotations on Buildpacks
+
+## Summary
+
+All Paketo Buildpacks must have at least the following OCI image annotations with values as defined by the [OCI Image Format Specification's Pre-Defined Annotation Keys](https://github.com/opencontainers/image-spec/blob/v1.1.0/annotations.md#pre-defined-annotation-keys):
+* `org.opencontainers.image.source`
+* `org.opencontainers.image.revision`
+* `org.opencontainers.image.title`
+* `org.opencontainers.image.version`
+
+The following additional annotation are recommended:
+* `org.opencontainers.image.url`
+* `org.opencontainers.image.description`
+
+## Motivation
+
+Knowing the origin and other metadata for a buildpack (which is an OCI image) is very helpful. Some examples of such use cases include finding release notes, user manuals, bug reporting procedures, and license information. Currently, it can be difficult to find the source control repository of a buildpack as that information is not available in a standard way.
+
+The OCI Image Format Specification's Pre-Defined Annotation Keys provide a standardized way to discover additional information about an OCI image. Because these annotations are standardized and widely used, tools have come to use them. For example, [Snyk](https://snyk.io/blog/how-and-when-to-use-docker-labels-oci-container-annotations/) and [Renovate](https://github.com/renovatebot/renovate/blob/34.115.1/lib/modules/datasource/docker/readme.md use these annotations.
+
+## Detailed Explanation
+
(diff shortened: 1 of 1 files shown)


text/0000-oci-image-annotations.md
```diff
@@ -0,0 +1,48 @@
...
+
+All Paketo Buildpacks must have at least the following OCI image annotations with values as defined by the [OCI Image Format Specification's Pre-Defined Annotation Keys](https://github.com/opencontainers/image-spec/blob/v1.1.0/annotations.md#pre-defined-annotation-keys):
+* `org.opencontainers.image.source`
+* `org.opencontainers.image.revision`
+* `org.opencontainers.image.title`
+* `org.opencontainers.image.version`
+
+The following additional annotation are recommended:
+* `org.opencontainers.image.url`
+* `org.opencontainers.image.description`
```
sophiewigmore, line 13: according to the spec you sent, would `org.opencontainers.image.created`, `org.opencontainers.image.authors`, `org.opencontainers.image.documentation`, etc be required as well?

candrews: I don't believe they would be required, but if they are available, they should be specified. For paketo, I don't see any of that information readily available in `buildpack.toml` or elsewhere, so I didn't include them in this RFC.

sophiewigmore: How do you know which fields in the upstream spec you sent are required?

candrews: I'm assuming that no fields are required, only that if they are present, they are required to meet the requirements for that field's value.

That's what I see in the wild, at least. I can't find a single image that specifies all of those fields, but I see a number that specify some.

sophiewigmore: Ah ok. The wording of the RFC `All Paketo Buildpacks must have at least the following OCI image annotations` made me think that this was required to be OCI compliant which was why I was confused.

I think this RFC seems like a good idea, especially if we can pull from the buildpack.toml fields as you suggested

ForestEckhardt:
As a quick clarifying question this is seeking to add these image labels to the buildpack images we release. Not to add them to the images that are produced by the buildpacks correct?

candrews:
> As a quick clarifying question this is seeking to add these image labels to the buildpack images we release. Not to add them to the images that are produced by the buildpacks correct?

Correct.

Although I do think that's a good idea, I think it would best be pursued in a separate RFC. https://github.com/paketo-buildpacks/image-labels does most of that already, so the implementation of such an RFC would be to apply it by default with reasonable default values.

candrews:
> It this is something that you would like added to the buildpack images themselves that I think it my make sense to try and push this up the stack into `pack`. We use the `pack buildpack package` to turn a "formatted" tarball into an OCI image so it may make sense to ask them to add these terms to that tarball by default using the values present in the `buildpack.toml`.

That makes sense to me! That would benefit more than just paketo too, which is nice.


> I would be more than happy to help facilitate an interaction if that is something that you are interested in.

I'm certainly interested, can you please point me in the right direction? Thank you!

ForestEckhardt:
@candrews Is there anything else that needs to be done on this RFC? Should we close it out in favor of an upstream PR or do you want to modify this one?

candrews:
> @candrews Is there anything else that needs to be done on this RFC? Should we close it out in favor of an upstream PR or do you want to modify this one?

I submitted the RFC to buildpacks at https://github.com/buildpacks/rfcs/pull/314 - could you please facilitate this process?

ForestEckhardt:
Hey there @candrews,
I see that the upstream proposal was accepted. Is there anything that still needs to happen with this RFC or has everything you wanted been accomplished in the upstream RFC?

candrews:
I believe the upstream RFC covers it. Now I hope it gets implemented 🤞 

Thank you!

PR #313, opened 2024-12-16
Propose the merge of static, tiny and base stacks and builders

loewenstein-sap:
[Preview](https://github.com/paketo-buildpacks/rfcs/blob/ubuntu-core-builder/text/0000-core-builder.md)


text/0000-core-builder.md
```diff
@@ -0,0 +1,46 @@
...
+We would replace the current `base`, `tiny` and `static` stacks with a single stack that includes all packages required by the buildpacks. Similarly, we would replace the current `base`, `tiny` and `static` builders with a single builder.
+
+## Rationale and Alternatives
+
+We could leave everything as it is. We would stick to maintaining four stacks repositories and eight builder repositories and creating new ones for every LTS we adopt. But that would increase the burden on the maintainers.
+
+## Implementation
+
+I would propose to create the new stack and builder as part of moving to Ubuntu Noble Numbat. I.e. 
+- we should create a repository `paketo-buildpacks/stack-ubuntu-noble`
```
dmikusa, line 27: I think we should consider the implications of https://github.com/buildpacks/rfcs/blob/main/text/0096-remove-stacks-mixins.md with this change.

1. Does it make sense to continue using the word `stack` in our repo name. The repo name and image names would be tricky to change later without potentially breaking things.

2. While I think removing stacks should be a separate change for us, I think we should at least consider how that might go and if we can easily do it with this set up. The idea being to use this change to position us well for stack removal.

text/0000-core-builder.md
```diff
@@ -0,0 +1,46 @@
...
+Users will be able to use the new builder to build and run their applications using any of the language family buildpacks included. Instead of requireing a decision for an optimised builder, they can "just" start and later decide on an optimised `run` image.
+
+## Motivation
+
+The Paketo project will be easier to use for users - they can start without being asked to decide on optimisation.
+The stacks and builders maintainer team will have less work to do - they will only need to maintain a single stack and builder instead of three.
+
+## Detailed Explanation
+
+We would replace the current `base`, `tiny` and `static` stacks with a single stack that includes all packages required by the buildpacks. Similarly, we would replace the current `base`, `tiny` and `static` builders with a single builder.
```
dmikusa, line 18: One thing that I think we should write into the detailed explanation is that we will by default squash layers in the builder. I propose that we do this per composite image. So our builder will contain *only* composite images and each composite image will get squashed into one layer, so that our builder does not exceed the layer limit. The pack cli supports this now, so it should be feasible.

text/0000-core-builder.md
```diff
@@ -0,0 +1,46 @@
...
+I would propose to delete the `paketo-buildpacks/noble-base-stack`, `paketo-buildpacks/noble-tiny-stack` and `paketo-buildpacks/noble-static-stack` repositories.
+I would propose to delete the `paketo-buildpacks/builder-noble-buildpackless-static`, `paketo-buildpacks/builder-noble-buildpackless-tiny` and `paketo-buildpacks/builder-noble-buildpackless-base` repositories.
+
+## Prior Art
+
+None.
+
+## Unresolved Questions and Bikeshedding
+
+What about the `full` stack and builders?
```
dmikusa, line 46: I feel like we should include full in this proposal. If we don't, that would effectively mean dropping support for buildpacks that only work in the full builder.

Have you looked at the difference between the base build image and the full build image? If it's not too much of a difference, we could conceivably do the same thing and just have another run image for full.

text/0000-core-builder.md
```diff
@@ -0,0 +1,46 @@
...
+
+I would propose to delete the `paketo-buildpacks/noble-base-stack`, `paketo-buildpacks/noble-tiny-stack` and `paketo-buildpacks/noble-static-stack` repositories.
+I would propose to delete the `paketo-buildpacks/builder-noble-buildpackless-static`, `paketo-buildpacks/builder-noble-buildpackless-tiny` and `paketo-buildpacks/builder-noble-buildpackless-base` repositories.
+
+## Prior Art
+
+None.
+
+## Unresolved Questions and Bikeshedding
+
```
dmikusa, line 45: Q: What do we do about stack/target & builder metadata?

I think it would be nice if the proposal described the new metadata to use for the stack/target & builder. Or is the plan to continue on using the existing metadata?  I don't have a strong opinion, but I think we should specify what that will be so it's clear.

dmikusa, line 45: Q: With pack, it is very easy to specify a different run image. Is it similarly easy when not using pack?

We have other platforms like Spring Boot users, Tekton users, and other CI systems where users don't use pack cli. Have we considered these to make sure that the user experience will still work well with this new approach?

dmikusa, line 45: Q: How does this impact language-specific builders?

I feel like the two are orthogonal, so this effort won't really change what is/can be done with a language-specific builder. It might be worth clarifying the intent though.

dmikusa, line 45: Q: Does this proposal also include an extension to auto-select a runtime?

I feel like the answer is no, that we would do that as a separate RFC or just a separate effort. I think it would be worth listing that as out of scope if that is the intent.

dmikusa (commented):
I like it. The simplification it proposes is good. I like that is can represent a reduction in work/effort as well. I posted some initial thoughts/questions to get this conversation going.

text/0000-core-builder.md
loewenstein-sap: This seems to be a builder implementation detail and an orthogonal concern. However, if this is non controversial, I am not opposed to adding it in. Would you mind to propose this as a suggestion? Should probably refer to the `pack` support for it and maybe briefly touch upon potential drawbacks?

text/0000-core-builder.md
loewenstein-sap: Mmh, I was used to still using stack as the overarching term for the combination of build and run images... I briefly [asked on Slack](https://cloud-native.slack.com/archives/C033DV9EBDF/p1734385684806259) and [Heroku seems to use "Bae Images"](https://cloud-native.slack.com/archives/C033DV9EBDF/p1734389150511999?thread_ts=1734385684.806259&cid=C033DV9EBDF).

Following that, we would have this:
```suggestion
- we should create a repository `paketo-buildpacks/ubuntu-noble-base-images`
```

I have no strong opinion on the matter though. Maybe react with 👍 or 👎 to indicate if the above suggestion should be merged.

text/0000-core-builder.md
loewenstein-sap: - Spring: https://docs.spring.io/spring-boot/maven-plugin/build-image.html#build-image.customization
- kpack: Doesn't support this setting explicitly (and unfortunately project.toml doesn't either)
- Tekton: https://github.com/buildpacks/tekton-integration/tree/main/task/buildpacks/0.4#parameters

Seems like we might want to raise this with CNB and work towards run image support in project.toml (which is supported by kpack and hence would solve this pretty generically).

loewenstein-sap: You are right, the RFC as-is does not tackle this and I would prefer to leave it as that.

Extensions are an interesting feature for this, but they come with their own weight, which I am not sure is worth it for this case.

text/0000-core-builder.md
loewenstein-sap: I haven't looked into this. I'd hesitate to increase the size and attack surface of the builder based on the needs of a few buildpacks. The build image is by far not as critical as the run images, but it still feels like a lot like catering for edge cases and put the weight on standard cases.

We should probably just keep providing a full builder in addition to the one proposed here. WDYT?

text/0000-core-builder.md
```diff
@@ -0,0 +1,46 @@
+# Ubuntu Core Builder
+
+## Summary
+
+Over the past, the Paketo project has created four different stacks and builders based on the Ubuntu Linux distribution. The difference is in the number of packages included which in turn is based on the requirements of the different buildpacks included.
+This RFC proposes to create a new stack repository that will create a single `build` image and a number of `run` images based on the Ubuntu distribution and replace the existing `static`, `tiny` and `base` stacks.
+This RFC further proposes to create a new builder repository that will create a single builder using the build image as a base layer and the `base` run image as the default `run` image.
+
+Users will be able to use the new builder to build and run their applications using any of the language family buildpacks included. Instead of requireing a decision for an optimised builder, they can "just" start and later decide on an optimised `run` image.
```
pacostas, line 9: I agree with that, it's confusing to choose which builder to use. But I think it would be better if the image starts by default being optimized and in case the user needs more packages into the run image (a bigger run stack image) then he/she could choose a different run image.

text/0000-core-builder.md
```diff
@@ -0,0 +1,46 @@
+# Ubuntu Core Builder
+
+## Summary
+
+Over the past, the Paketo project has created four different stacks and builders based on the Ubuntu Linux distribution. The difference is in the number of packages included which in turn is based on the requirements of the different buildpacks included.
+This RFC proposes to create a new stack repository that will create a single `build` image and a number of `run` images based on the Ubuntu distribution and replace the existing `static`, `tiny` and `base` stacks.
```
pacostas, line 6: I think it would be preferable to have run images based on the engine (e.g. nodejs, java, etc.) This would minimize to the absolute necessary, the packages that are needed for the engine during the runtime. We can also still have the static, tiny and the base for backwards compatibility.

text/0000-core-builder.md
loewenstein-sap: I think this is a trade-off to consider (as usual) and I would be inclined to prioritize "works out of the box" over "provides minimal dependencies".

Extensions could be a way to remove the need to choose one over the other. However, they introduce other trade-offs. For one, with the recent news on Kaniko (see [here](https://github.com/GoogleContainerTools/kaniko/issues/3348) and [here](https://github.com/GoogleContainerTools/kaniko/pull/3345)) I'd say the future of at least the implementation of Extensions is unclear. Independent of that, I think an implementation for choosing the best run image based on language choice using Extensions can quickly become unwieldy - as far as I remember this would require Extension(s) and special Buildpack(s) to work in collaboration with each other via the Build Plan.
It is likely possible, but I am not sure it is worth the effort.

text/0000-core-builder.md
loewenstein-sap: Actually, as @pacostas points out below, there could be language specific run images instead of (or in addition to) language specific builders.

text/0000-core-builder.md
loewenstein-sap: Indeed, I would consider the `static`, `tiny`, and `base` run images as kind of a means for compatibility. Although `base` also serves as an enabler for "works out of the box" (see my comment above).

text/0000-core-builder.md
```diff
@@ -0,0 +1,46 @@
...
+    - we should use the above `build` image as a base layer for the `builder` image
+    - we should use the above `run` image as the default `run` image for the `builder.toml`
+    - we should use the above `run-static` and `run-tiny` images as additional `run` images for the `builder.toml`
+
+I would propose to delete the `paketo-buildpacks/noble-base-stack`, `paketo-buildpacks/noble-tiny-stack` and `paketo-buildpacks/noble-static-stack` repositories.
+I would propose to delete the `paketo-buildpacks/builder-noble-buildpackless-static`, `paketo-buildpacks/builder-noble-buildpackless-tiny` and `paketo-buildpacks/builder-noble-buildpackless-base` repositories.
+
+## Prior Art
+
+None.
```
pacostas, line 42: We have implemented that for the stacks https://github.com/paketo-community/ubi-base-stack so practically by taking the stack.toml of the aforementioned stacks and putting them under the stacks directory and also changing the info on the images.json to match to what is on the stacks directory, should work. The only thing that hasn't been thoroughly tested is the usns flow when it supports multiple stacks. This implementation supports multi-arch and is also backward compatible with the existing stacks, so it will not break any of the current workflows/stacks.

text/0000-core-builder.md
pacostas: An alternative is the extension, which can choose which run image to use.

c0d1ngm0nk3y: > An alternative is the extension, which can choose which run image to use.

But doesn't have this the problem that user input (`--run-image`) would be overruled?

text/0000-core-builder.md
loewenstein-sap: @pacostas Does the below give a good explanation of what you've done for UBI and how this could be taken as the blueprint for this RFCs proposal?

```suggestion
This has been done in https://github.com/paketo-community/ubi-base-stack already. The `stacks` folder and `images.json` should give guidance on how to create multiple run images.
```

text/0000-core-builder.md
pacostas: Yes, good point. That's interesting, I dont know what happens in that case. I'll checkout it out, it might be something we need to clarify/patch on pack

pacostas: Yes, but for each language we need to have the corresponding lanuage-builder, which is not that bad, but we will end up having as many builders as the languages are. Also, I think buildpacks are supposed to be a little bit more abstract when we build an app e.g. using builder-ubuntu instead of builder-nodejs-ubuntu.

loewenstein-sap: @pacostas I guess what @dmikusa is hinting at is the relation to https://github.com/paketo-buildpacks/rfcs/blob/main/text/0055-create-language-family-builders.md

loewenstein-sap: @dmikusa If you think the notion of language-specific builder and/or run images should be included in the RFC, would you mind adding a concrete text change suggestion. Thanks.

pacostas: The extension is a subset of this implementation, so in that case there is no need to have multiple builders, as the extension can select any of the available run images, so having only one builder is enough.

dmikusa: Minimally, I think the RFC should propose the builder & stack ids to use. The other builder & stack metadata is not as important.

I think we should start adding target metadata, if we're not doing that already. My understanding is that it can live side-by-side with stack metadata. So we should probably specify in the RFC what target metadata to set, see https://github.com/buildpacks/rfcs/blob/main/text/0096-remove-stacks-mixins.md#base-image-metadata.

I'd also like to see included the specific current image references and the future image reference.

I think this will help with implementation, but I think it'll also just make it more clear how we are consolidating things. So maybe have it in a before/after style?

dmikusa: @loewenstein-sap I agree. I think we should just specifically spell that out here.

```suggestion
Q: Does this proposal require Paketo buildpacks language families to implement an extension to switch or auto-select runtimes?
A: No. That is out of the scope of this RFC, and will be the topic of a future discussion and possibly future RFC.
```

dmikusa:
```suggestion
Q: How does this impact the [language-specific builders RFC](https://github.com/paketo-buildpacks/rfcs/blob/main/text/0055-create-language-family-builders.md)?
A: It does not impact language-specific builders and Paketo buildpacks language families can continue to ship language-specific builders as needed.

```

dmikusa:
```suggestion
Q: With pack, it is very easy to specify a different run image. Is it similarly easy when not using pack?
A: We know this to work with some platforms. The notable exception is kpack.

For example:

- Spring: https://docs.spring.io/spring-boot/maven-plugin/build-image.html#build-image.customization
- Tekton: https://github.com/buildpacks/tekton-integration/tree/main/task/buildpacks/0.4#parameters
```

text/0000-core-builder.md
dmikusa: > I think this is a trade-off to consider (as usual) and I would be inclined to prioritize "works out of the box" over "provides minimal dependencies".

+1 - I think it needs to just work out of the box. If it's not the smallest possible image, that's OK. We can then provide other runtime images that allow you to make smaller app images.

I think a nice side effect of this approach is that it socializes and makes commonplace the usage of a runtime image not provided by the builder. This could even allow a user to take our runtime image and add or remove stuff in their own custom runtime image. You could do this before, but it's making it more accessible / "normal" if we're suggesting folks do it.

text/0000-core-builder.md
dmikusa:
```suggestion
We would replace the current `base`, `tiny` and `static` stacks with a single stack that includes all packages required by the buildpacks. Similarly, we would replace the current `base`, `tiny` and `static` builders with a single builder.

Container images have a layer limit and in the past, we have had issues with builders that include a lot of buildpacks because this can quickly cause the builder to hit the layer limit. To prevent this from happening, we will by default squash layers for all composite buildpacks included in the builder. This will be done by using the `--flatten` of `pack builder create`, and it should allow the builder to scale to include all of the Paketo buildpacks.
```

I don't think this is controversial. The functionality exists and we've talked about doing it with our builders before, but I just don't think it was ever implemented.

As far as drawbacks, I don't think there are any major ones. You could potentially see larger image sizes because if you have individual layers per-buildpack and if you have a builder that has multiple copies of a buildpack, say the environment-variables buildpack, then you'd only get one layer for that buildpack. If you squash them so there's a layer per composite, then you won't get that layer sharing.

I don't think this will really matter because in reality, we're still getting duplicates because composites often have different versions of a buildpack and each version is its own layer. Also, having fewer layers as a whole is typically considered better for images (look at any major image on Docker Hub) and it will typically download faster because the client has to do more work for lots of little layers.

text/0000-core-builder.md
loewenstein-sap: > Minimally, I think the RFC should propose the builder & stack ids to use. The other builder & stack metadata is not as important.

Actually, we might even want to avoid stack id and stack metadata in general and switch over to using the new metadata. Are there many things to consider other that the few explicit references to the tiny stack id? Might be worth tackling as part of this.

text/0000-core-builder.md
loewenstein-sap: @dmikusa Thoughts?

text/0000-core-builder.md
loewenstein-sap: The [list of packages in full](https://github.com/paketo-buildpacks/jammy-full-stack/blob/6f7fccb60993efc39b78008300b03b9ac15c5ecc/stack/stack.toml#L21-L202) is quite a lot larger than [the list of packages in base](https://github.com/paketo-buildpacks/jammy-base-stack/blob/ce49ff79924e71eb0d55c75f3c96cc3af730226e/stack/stack.toml#L21-L38). 

For me, adding full packages to this new build image is not an option. I guess I will put a note into the RFC to show that if we need a full build image, we need separate build, run and builder images.

text/0000-core-builder.md
dmikusa: > briefly [asked on Slack](https://cloud-native.slack.com/archives/C033DV9EBDF/p1734385684806259) and [Heroku seems to use "Bae Images"](https://cloud-native.slack.com/archives/C033DV9EBDF/p1734389150511999?thread_ts=1734385684.806259&cid=C033DV9EBDF).

👍  Yes, that's what I've heard used as well. Base image. I think they wanted to use more standard terminology. 

> I have no strong opinion on the matter though. Maybe react with 👍 or 👎 to indicate if the above suggestion should be merged.

👍  It's probably not a huge thing. It just seemed like a good opportunity.

text/0000-core-builder.md
dmikusa:
```suggestion
Q: What about the `full` stack and builders?
A: This proposal does not include the `full` builder and recommends that it be deprecated. If approved, this RFC will make the Jammy Full stack the last supported iteration of the full stack. There is presently only one buildpack, PHP, which requires the full stack and is thus impacted. This buildpack will be able to continue using the Jammy Full stack through [April 2027](https://ubuntu.com/about/release-cycle) with the goal of converting it to run on the new Noble all-in-one stack before that point.
```

Ugh, yes. That's a lot of stuff we'd have to add. I agree that it's too much to stuff all in one build image. It's also a lot of work to support one additional buildpack. How about this?

text/0000-core-builder.md
dmikusa:
```suggestion
This RFC proposes to create a new stack repository that will create a single `build` image and a number of `run` images based on the Ubuntu distribution and replace the existing `static`, `tiny` and `base` stacks. The `full` stack will be deprecated immediately, making the Jammy stack the last iteration of it.
```

text/0000-core-builder.md
```diff
@@ -0,0 +1,57 @@
...
+Users will be able to use the new builder to build and run their applications using any of the language family buildpacks included. Instead of requireing a decision for an optimised builder, they can "just" start and later decide on an optimised `run` image.
+
+## Motivation
+
+The Paketo project will be easier to use for users - they can start without being asked to decide on optimisation.
+The stacks and builders maintainer team will have less work to do - they will only need to maintain a single stack and builder instead of three.
+
+## Detailed Explanation
+
+We would replace the current `base`, `tiny` and `static` stacks with a single stack that includes all packages required by the buildpacks. Similarly, we would replace the current `base`, `tiny` and `static` builders with a single builder.
```
dmikusa, line 18:
```suggestion
We would replace the current `base`, `tiny` and `static` stacks with a single stack that includes all packages required by the buildpacks. Similarly, we would replace the current `base`, `tiny` and `static` builders with a single builder.

The `full` stack will be deprecated, making the present Jammy stack the last iteration of the `full` stack released by the Paketo team. This provides support for the full stack through [April of 2027](https://ubuntu.com/about/release-cycle), which is a sufficiently long deprecation period. Upon approval of this RFC, the Paketo Steering Committee will write a blog post announcing the stack changes and the deprecation of the full stack. The PHP buildpack will then need to be altered so that it can run on the new base stack or it will need to be changed to utilize extensions to install any additional packages that it requires over and above what is in the base image.
```

text/0000-core-builder.md
loewenstein-sap: > I'd also like to see included the specific current image references and the future image reference.
> 
> I think this will help with implementation, but I think it'll also just make it more clear how we are consolidating things. So maybe have it in a before/after style?

I added https://github.com/paketo-buildpacks/rfcs/pull/313/files#diff-a6cd29e680d46138ca93c24308bc8ac8f126e1d73463d1691e5091c2faf11eadR42-R48

anthonydahanne:
Since specific language builders are explicitly not threatened by this RFC, and for the sake of simplification, fine by me!

text/0000-core-builder.md
loewenstein-sap: I added https://github.com/paketo-buildpacks/rfcs/pull/313/commits/ff8bafd51c3756fb3a5cf695c5b95bfd83590a9c to address new and old metadata. WDYT @dmikusa?

text/0000-core-builder.md
loewenstein-sap: @pacostas Any strong opinion on your end or are you fine with resolving this conversation as-is, i.e. with a default image leaning towards "runs out-of-the-box" rather than "runs with minimal base image"?

loewenstein-sap:
After https://github.com/paketo-buildpacks/rfcs/pull/313/commits/ff8bafd51c3756fb3a5cf695c5b95bfd83590a9c, there doesn't seem to be too many things open here, so I would hope we can soon call for a final commenting period. WDYT @paketo-buildpacks/steering-committee?

@robdimsdale Are you still active in the @paketo-buildpacks/builders-maintainers team? Would be great if you could chime in and comment.

text/0000-core-builder.md
pacostas: @loewenstein-sap Thank you for asking, sounds good to resolve this conversation as the implementation of this RFC supports both scenarios which can easily enable in the future, whether we want to have the optimized images or not.

mhdawson (approved):
I think this is a good idea, and I'm ok with starting where the default image is the large one/covers all cases one.

I think the automation that @pacostas has done for the ubi stack should make it relatively easy to implement as well.

LGTM

text/0000-core-builder.md
loewenstein-sap: @dmikusa what do you think, can we mark this as resolved?

loewenstein-sap:
@paketo-buildpacks/steering-committee time to merge?

dmikusa:
I'm fine with it. We have two-thirds of the steering committee voting 👍 as well as one member of stacks/builders team. Let's start the final call for comments. If no concerns are raised by Friday 2025-1-24, then we'll merge this.

@ForestEckhardt @robdimsdale Any thoughts on this one?

dmikusa:
Seeing no further feedback, this is approved and I'm going to merge.

PR #314, opened 2024-12-17
Propose to manage project status

loewenstein-sap:


text/0000-buildpack-status.md
```diff
@@ -0,0 +1,38 @@
+# Introduce a maintenance status for buildpacks
+
+## Summary
+
+Given the limited capacity of the maintainer teams, we should start to introduce a disctinction between the different levels of maintenance individual buildpacks provide.
```
dmikusa, line 5:
```suggestion
Given the finite capacity of the maintainer teams, we should introduce a distinction between the different levels of maintenance individual buildpacks provide.
```

dmikusa: Limited sounds negative, IMO, finite is just a reflection of reality.

text/0000-buildpack-status.md
```diff
@@ -0,0 +1,38 @@
...
+We should check with maintainers of the buildpacks if they agree with the proposed level of maintenance.
+
+## Prior Art
+
+None.
+
+## Unresolved Questions and Bikeshedding
+
+- What levels of maintenance should we have? I came up with two different levels of maintenance and one to express that the buildpack is not maintained anymore. I am happy to introduce finer grained levels of maintenance if that is desired.
+- What about the buildpacks that are not maintained anymore? Should we archive them or move them to the paketo-community organization?
```
dmikusa, line 38:
```suggestion
- What about the buildpacks that are not maintained anymore? Should we archive them or move them to the paketo-community organization?
- What about paketo-community? Do we really need that if we have this tier system? 
```

dmikusa: I'd be more in favor of getting rid of paketo-community and just moving any useful/maintained buildpacks into the corresponding tier in paketo-buildpacks. There's a bunch of overhead with paketo-community in terms of managing it, keeping GH teams in sync, keep secrets in sync, etc... and I don't feel like it has really achieved it's purpose, which is to attract folks to contribute community managed buildpacks.

text/0000-buildpack-status.md
```diff
@@ -0,0 +1,38 @@
...
+We should document the different levels of maintenance.
+We should check with maintainers of the buildpacks if they agree with the proposed level of maintenance.
+
+## Prior Art
+
+None.
+
+## Unresolved Questions and Bikeshedding
+
+- What levels of maintenance should we have? I came up with two different levels of maintenance and one to express that the buildpack is not maintained anymore. I am happy to introduce finer grained levels of maintenance if that is desired.
```
dmikusa, line 37: I think we have to be careful with terms like maintenance and support, as to some readers that will imply a contract. That the project has actual guarantees on things, when in fact everything we do is "best effort". This is part of what I'd mentioned above about readers assuming things, and I think using an opaque term for the different levels will help prevent any miscommunication here.

mhdawson: Maybe using the term "activity level" which reflects what's in the wording interms of how actively the buildpacks may be updated but does not imply any promised maintenance so maybe something like:

* Active
* Security updates only
* incubating, bootstrapping
* Inactive

dmikusa: I feel like this creates a marketing/branding problem for buildpacks that are generally stable and not getting feature work. By saying they're "security updates only", that comes off (to me) like they're about to go EOL. This is coming from a typical software lifecycle like for a language runtime or Linux distribution where "security updates only" is typically the last thing that happens before that version goes EOL.

I also think there are at least three different things that we're trying to describe/convey to users with this description and it is hard to do, which is why I like the idea of an opaque description. The three things that come to my mind, things I'd want to know as a user, are 1.) the release reliability, are we releasing this often? is it likely to get a release when I need it to?  2.) Is it getting security updates in a timely manner? Is someone watching & merging these? If something breaks with the automated process, is someone going to fix it?  3.) Are there active maintainers? Is someone looking at issues, engaging with the community, doing development, etc...

mhdawson: Thinking about it a bit more Tiers are used in the Node.js project as outlined in https://github.com/nodejs/node/blob/main/BUILDING.md#strategy. That is an example of where the use of Tiers has been successful used.

text/0000-buildpack-status.md
```diff
@@ -0,0 +1,38 @@
+# Introduce a maintenance status for buildpacks
+
+## Summary
+
+Given the finite capacity of the maintainer teams, we should introduce a distinction between the different levels of maintenance individual buildpacks provide.
+The proposed levels are:
+- **Active Maintenance**: The buildpack is actively maintained and updated by the maintainers. Feature requests and bug reports are actively worked on. The maintainers have capacity for coordinated cross project work.
+- **Security Maintenance**: The buildpack is maintained and updated by the maintainers, but the focus is on critical bugs and project hygene like dependency updates. The maintainers might have limited capacity for cross project work.
+- **Out of maintenance**: The buildpack is not actively maintained by the maintainers. The maintainers might still accept PRs and bug reports, but the response time is not guaranteed. The maintainers have no capacity for cross project work.
```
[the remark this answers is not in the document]
loewenstein-sap: I am fine with simple opaque terms like "Tier N" or "Nth Tier", but I'd avoid any mapping to derive quality - even Gold, Silver, Bronze (which is probably pretty universal) - for reasons like "green" traffic lights being blue in Japan and other potential for misunderstandings.

When it comes to describing the qualities of the different tiers, I would like to point out that there are two different perspectives on this. 
1. What can users expect
2. What can Paketo maintainers expect

I don't think that we can afford the second tier from a maintainer perspective, unless we also define that only Tier 1 buildpacks can participate in the projects commons. I.e. only Tier 1 buildpacks get packaged into the maintained builders, only Tier 1 buildpacks "are allowed" to use `libpak` or `packit`, etc. Otherwise, the buildpacks not participating in RFCs and cross project work would block the cross project work.

text/0000-buildpack-status.md
loewenstein-sap: So, let's go for **Tier N** then?

text/0000-buildpack-status.md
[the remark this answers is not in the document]
loewenstein-sap: I've just pushed a new version that picks up on the comments. I decided to stick to three tiers and allow incubation as a separate flag. The main point would be, that imho we shouldn't allow incubation to target Tier 2 at all.
I would also see a clear separation between Tiers 1 and 2 - with only the first Tier being considered for inclusion into project maintained builders and be consulted about common libs and tools. Tier 2 buildpacks are free to use those, but might have to adapt in case of changes.

loewenstein-sap:
With the first @paketo-buildpacks/steering-committee approval in place, I'd like to call for a final review period.

@ForestEckhardt Any feedback from your end?
@mhdawson I hope I accommodated your feedback well.

@paketo-buildpacks/app-monitoring @paketo-buildpacks/builders @paketo-buildpacks/content @paketo-buildpacks/dependencies @paketo-buildpacks/dotnet-core @paketo-buildpacks/go @paketo-buildpacks/java @paketo-buildpacks/nodejs @paketo-buildpacks/php @paketo-buildpacks/python @paketo-buildpacks/ruby @paketo-buildpacks/rust @paketo-buildpacks/stacks @paketo-buildpacks/tooling @paketo-buildpacks/utilities @paketo-buildpacks/web-servers
Now would be a great time to chime in if you had any comments.

anthonydahanne (approved):

robdimsdale (approved):

jericop:
Although I'm not currently a maintainer, I support these changes.

PR #319, opened 2025-04-07
Propose removing Paketo Java Azure composite buildpack

dmikusa:


text/java/0018-retire-java-azure-composite.md
```diff
@@ -0,0 +1,40 @@
...
+5. We don't want to set an expectation that we will create composite buildpacks for every JVM vendor.
+
+## Detailed Explanation
+
+The process is simple.
+
+1. Write a blog post announcing the change. Provide instructions for how to continue using Microsoft JDK without this buildpack.
+2. Give users 30 days to update.
+3. Archive the buildpack on Github. This will stop all workflows.
+
```
anthonydahanne, line 22: shouldn't we also delete at least the `:latest` tag? to make sure nobody uses it with the expectation it's maintained

dmikusa: Hmm, not sure. On one hand, I agree. That would help to make sure that no one is using this with the expectation that it's still being supported. On the other hand, that is an aggressive change. It definitely would break anyone still using it.

I'm OK with it given that part of the mandate of buildpacks is to ensure users are getting all the latest updates, bug and security fixes. If you're on "latest" that tells me you want the latest stuff, so this seems reasonable. If you wanted to have stability, then you'd be using one of the specific tags which we're not changing. 

I'll make sure to cover this in the blog post as well.

text/java/0018-retire-java-azure-composite.md
```diff
@@ -0,0 +1,40 @@
...
+4. We are introducing the jvm-vendors buildpack, which will even more easily allow users to pick different JVMs.
+5. We don't want to set an expectation that we will create composite buildpacks for every JVM vendor.
+
+## Detailed Explanation
+
+The process is simple.
+
+1. Write a blog post announcing the change. Provide instructions for how to continue using Microsoft JDK without this buildpack.
+2. Give users 30 days to update.
+3. Archive the buildpack on Github. This will stop all workflows.
```
dmikusa, line 21:
```suggestion
3. Archive the buildpack on Github. This will stop all workflows.
4. Delete the "latest" tag on the Docker Hub repository for `paketobuildpacks/java-azure`.
```

anthonydahanne (approved):

jericop (approved):

dmikusa:
Not seeing any more feedback on this, and we have sufficient votes so I'm going to merge it.

PR #320, opened 2025-04-14
Update 0061-core-builder.md

dmikusa:
Quick update to the summary to clarify that we're including buildpackless builder too.

### DIFF
--- a/text/0061-core-builder.md
+++ b/text/0061-core-builder.md
@@ -4,9 +4,9 @@
 
 Over the past, the Paketo project has created four different stacks and builders based on the Ubuntu Linux distribution. The difference is in the number of packages included which in turn is based on the requirements of the different buildpacks included.
 This RFC proposes to create a new stack repository that will create a single `build` image and a number of `run` images based on the Ubuntu distribution and replace the existing `static`, `tiny` and `base` stacks. The `full` stack will be deprecated immediately, making the Jammy stack the last iteration of it.
-This RFC further proposes to create a new builder repository that will create a single builder using the build image as a base layer and the `base` run image as the default `run` image.
+This RFC further proposes to create a new builder repository that will create a two builders, both using the build image as a base layer and the `base` run image as the default `run` image. One builder will not contain any buildpacks and the other will contain a full set of Paketo buildpacks.
 
-Users will be able to use the new builder to build and run their applications using any of the language family buildpacks included. Instead of requireing a decision for an optimised builder, they can "just" start and later decide on an optimised `run` image.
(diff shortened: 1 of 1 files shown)


text/0061-core-builder.md
```diff
@@ -4,9 +4,9 @@
 
 Over the past, the Paketo project has created four different stacks and builders based on the Ubuntu Linux distribution. The difference is in the number of packages included which in turn is based on the requirements of the different buildpacks included.
 This RFC proposes to create a new stack repository that will create a single `build` image and a number of `run` images based on the Ubuntu distribution and replace the existing `static`, `tiny` and `base` stacks. The `full` stack will be deprecated immediately, making the Jammy stack the last iteration of it.
-This RFC further proposes to create a new builder repository that will create a single builder using the build image as a base layer and the `base` run image as the default `run` image.
+This RFC further proposes to create a new builder repository that will create a two builders, both using the build image as a base layer and the `base` run image as the default `run` image. One builder will not contain any buildpacks and the other will contain a full set of Paketo buildpacks.
```
loewenstein-sap, line 7: Full set of Paketo buildpacks? I know it's a bit chicken and egg with this and the status RFC, but the plan is to include the full set of Tier 1 buildpacks, isn't it?

dmikusa, line 7:
```suggestion
This RFC further proposes to create a new builder repository that will create a two builders, both using the build image as a base layer and the `base` run image as the default `run` image. One builder will not contain any buildpacks and the other will contain Paketo buildpacks.
```

Yes, probably too much to get into here. Just want to say that one is buildpackless and one has buildpacks. I don't think we need further detail than that here.

pacostas, line 7: I like this idea, but this means that we need to refactor the builders implementation to support generating multiple builders and also generating multiple architectures for each builder. There is no prior work on that side, and also seems that currently there is no volunteer to implement that. As a result, this might take months to get implemented and have a first release.

loewenstein-sap: @c0d1ngm0nk3y would that be something we can pick up?

BarDweller:
Discussed a little in the 05/07 wg meeting, and we were wondering if this would also apply to older images, like jammy, if they are not archived/deprecated when this is implemented. It would meet a goal of reducing overhead explaining build to new team members, reducing repo's etc, But the effort may not be worth it if the retirement of Jammy is close by.

loewenstein-sap:
@BarDweller I wouldn't be opposed to doing this for Jammy as well. Although, I don't think we can remove the `full` builder for Jammy - so we might end up with one repo (or pair of repos) more than for Noble.

PR #321, opened 2025-04-25
Add RFC: Python package manager buildpacks consolidation

sgaist:
## Summary

Consolidate Python package managers multiple buildpacks.

## Use Cases

- Reduce the number of buildpacks to maintain
- Simplify adding new package managers

## Checklist

* [X] I have added an integration test, if necessary.

### DIFF
--- a/text/python/0000-consolidate-python-package-managers.md
+++ b/text/python/0000-consolidate-python-package-managers.md
@@ -1,0 +1,53 @@
+# Consolidate Python package managers buildpacks into one
+
+## Summary
+
+The goal is to centralize the handling of the various existing Python package
+managers under one single buildpack.
+
+## Motivation
+
+This consolidation will allow to reduce the number of buildpacks required to
+handle Python projects. From a consumer point of view, this will make no
+difference however for maintainers, this will simplify their tasks:
+
+- 1 code base against 4 currently
+- 1 release to manage against 4 currently
+- adding a new package manager such a pixi or uv will be simplified
+
+## Detailed Explanation
+
+Currently, each package manager is handled through at least two different
+buildpacks:
+
+1. One that installs the manager
+2. One that detects if the manager should be used and consequently makes use of
+  it
+
+This means that there's currently eight repositories to keep updated with regard
+to new manager versions but also for dependencies, security issues, etc.
+
+This new buildpack would handle the latter for all supported managers effectively
+replacing four buildpacks with one at the expense of a larger code base but with
+the benefit of pooling efforts in a single place.
+
+As a side effect, it also makes it easier to add support for new package
+managers since only one new buildpack needs to be created.
+
+## Implementation
+
+The most simple implementation would be to pick the code from the original
+buildpacks together under one roof and do each detection in sequence as is
+currently done by the lifecycle.
+
+Care should be taken to harmonize the code so as to simplify maintenance as
+well.
+
+## Prior Art
+
+Already existing buildpacks:
+
+- paketo-buildpacks/conda-env-update
+- paketo-buildpacks/pip-install
+- paketo-buildpacks/pipenv-install
+- paketo-buildpacks/poetry-install
--- a/text/python/0000-consolidate-python-package-managers.md
+++ b/text/python/0000-consolidate-python-package-managers.md
@@ -1,9 +1,12 @@
-# Consolidate Python package managers buildpacks into one
+# Consolidate Python package managers buildpacks
 
 ## Summary
 
 The goal is to centralize the handling of the various existing Python package
-managers under one single buildpack.
+managers under two buildpacks. Each handling a specific task:
+
+- Package manager installation
+- Package manager usage
 
 ## Motivation
 
@@ -11,8 +14,8 @@ This consolidation will allow to reduce the number of buildpacks required to
 handle Python projects. From a consumer point of view, this will make no
 difference however for maintainers, this will simplify their tasks:
 
-- 1 code base against 4 currently
-- 1 release to manage against 4 currently
+- 2 code bases against 8 currently
+- 2 release to manage against 8 currently
 - adding a new package manager such a pixi or uv will be simplified
 
 ## Detailed Explanation
@@ -27,18 +30,20 @@ buildpacks:
 This means that there's currently eight repositories to keep updated with regard
 to new manager versions but also for dependencies, security issues, etc.
 
-This new buildpack would handle the latter for all supported managers effectively
-replacing four buildpacks with one at the expense of a larger code base but with
-the benefit of pooling efforts in a single place.
+These new buildpacks would replace eight buildpacks with two at the expense
+of a larger code base but with the benefit of pooling efforts in a single
+place.
 
 As a side effect, it also makes it easier to add support for new package
-managers since only one new buildpack needs to be created.
+managers since they will now be handled centrally.
 
 ## Implementation
 
-The most simple implementation would be to pick the code from the original
-buildpacks together under one roof and do each detection in sequence as is
-currently done by the lifecycle.
+For the package manager user buildpack, the most simple implementation would be
+to pick the code from the original buildpacks together under one roof and do
+each detection in sequence as is currently done by the lifecycle.
+
+A similar approach shall be applied to the installer buildpack.
 
 Care should be taken to harmonize the code so as to simplify maintenance as
 well.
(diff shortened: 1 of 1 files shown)


jericop:
I'm in favor of this. One thing to consider is how we decide what order to process the package manager detections in. Does pip:requirements.txt go first, and then poetry:pyproject.toml, etc? We could just copy the existing order from the python composite buildpack since that ordering is already established. I assume like other buildpack order groups, we would stop "detecting" once we find a package manager and then move on with the appropriate build phase.

sgaist:
@jericop I agree with the ordering. It's better to keep consistent with the current implementation. One thing that came to mind but isn't implemented currently was to add a configuration variable that would allow the buildpack user to set the detection order or to limit which detector should be used (e.g. they know that they only want to support poetry so ensure that the pipeline would fail if any other package manager is detected).

While writing the proof of concept I started wondering whether we should do something similar with the tool provider buildpacks ?

jericop:
I really like the idea of allowing the user to control the detection order probably by setting an environment variable. I could foresee an app that has requirements.txt for pip and pyproject.toml for black or mypy. Good call.

> While writing the proof of concept I started wondering whether we should do something similar with the tool provider buildpacks ?

Not sure which buildpack(s) you are referring to. Could you elaborate?

sgaist:
> I really like the idea of allowing the user to control the detection order probably by setting an environment variable. I could foresee an app that has requirements.txt for pip and pyproject.toml for black or mypy. Good call.
> 

Good I'll add that to the todo list

> > While writing the proof of concept I started wondering whether we should do something similar with the tool provider buildpacks ?
> 
> Not sure which buildpack(s) you are referring to. Could you elaborate?

I am thinking about the `miniconda`, `poetry`, `pip` and `pipenv`. The ones that install the package managers themselves.

jericop:
> I am thinking about the `miniconda`, `poetry`, `pip` and `pipenv`. The ones that install the package managers themselves.

Ah. Got it. I just re-read the text for this RFC a little more carefully and realize you were talking about consolidating the buildpacks that do the "run" part of the package manager such as pip-install and poetry-install. I was just reviewing the READMEs for some of those buildpacks and noticed that they "provide" things for downstream buildpacks. We would need to make sure we sort this out correctly. I wonder if starting with a package manager "install" buildpack that provides the package manager is a good plan for this rfc. One thing that is nice about each individual buildpack is that the README shows what it does and provides in a fairly concise way. If we plan to maintain existing behavior, the README for a consolidated package manager "run" buildpack might be confusing.

If you prefer to focus on a consolidated package manger "run" buildpack, then maybe you could add some more thoughts on how it would handle maintaining the existing behavior of the individual buildpacks it would replace. Regardless, it's good to sort this out beforehand.

sgaist:
You can see a tentative implementation of the "run" part at https://github.com/idiap/python-packagers/. It currently runs each detection in sequence as would the current set of "run" buildpacks. Note that it is not optimal. While implementing a better support for `pyproject.toml`, I found some things that can be improved.

With regard to a consolidated "installers" buildpack, do you think it should be a separated buildpack or a part of the `python-packagers` buildpack ?

dmikusa:
> With regard to a consolidated "installers" buildpack, do you think it should be a separated buildpack or a part of the python-packagers buildpack ?

I would suggest putting it in the same buildpack. I think you're going to have very similar logic for when you should install and when you should run, since you *usually* will need to install it before you can run it. Consolidating those makes sense, IMHO. 

The one thing to be careful about if you decide to consolidate them is that you still want to support the cases where someone just wants one installed (possibly for use with a custom buildpack), and if someone wants to just have them run (possibly for use if someone has another way of supplying the tool). This isn't usually hard to support, but it's easy to overlook.

dmikusa:
> You can see a tentative implementation of the "run" part at https://github.com/idiap/python-packagers/. It currently runs each detection in sequence as would the current set of "run" buildpacks. Note that it is not optimal. While implementing a better support for pyproject.toml, I found some things that can be improved.

This looks great. I don't think there's any problem with running the detections in a known/documented order. That's effectively what is happening when there are multiple buildpacks. They're running in the defined sequence and the lifecycle is going to pick the first successful buildplan that comes out of detection. By running them in order, I think you're going to get a very similar (maybe not exact 🤔) detection behavior.

My $0.02. Shoot for parity with your initial plans. Log stories/issues for improvements, unless they're critical to the success of the migration. This is kind of an MVP approach, which will hopefully get to the point where we could swap in the new buildpack faster.

jericop:
Thanks for putting together a great initial implementation.

> The one thing to be careful about if you decide to consolidate them is that you still want to support the cases where someone just wants one installed (possibly for use with a custom buildpack), and if someone wants to just have them run (possibly for use if someone has another way of supplying the tool). This isn't usually hard to support, but it's easy to overlook.

I can think of a few situations where I have had to stitch together different buildpacks using the paketo-community/build-plan to supply multiple languages (python and golang for example) during the build. I think we should continue to support this use-case because there may be situations where someone has a custom buildpack to package or set the entry point for their app, but they rely on paketo buildpacks to supply cpython, pip, pip-install,etc.

I suggested 2 buildpacks (installer and packager) mostly to keep build plans simple. I'm definitely open to other ideas and suggestions. At the very least we have a path forward to consolidate buildpacks and reduce maintenance, so thanks for all of this awesome progress.

dmikusa:
> The main thing that might be less comfortable is that all the dependencies for all the sub-packages are located in the buildpack.toml.

I think that is fine. It can be a lot of entries, but updates to these are automated, so it's not really that big of a deal.

sgaist:
Just a small ping to see how we can move forward with this.

https://github.com/idiap/python-installers/
https://github.com/idiap/python-packagers/

Are both fully implemented, on par with the current set of buildpacks and even have some more stuff:

- improved `pyproject.toml` handling (i.e. does not select only poetry)
- uv support
- pixi support
- auto-update for miniconda
- auto-update for uv and pixi

Something I had to add in order to support recent miniconda versions is to enable "auto-accept ToS". See their [tos plugin documentation](https://www.anaconda.com/docs/getting-started/tos-plugin) for more information. I don't like it much but I wanted to avoid breaking user environments.

So I gave it some thoughts and I can see two options with regard to that issue:

- Make the ToS acceptation a BP environment variable that users have to set to true in order to continue as they would need to do that anyway in their work environment.
- Move to mamba / micromamba that use conda-forge by default

There are some more things that I think are worth doing:

- Moving tests to use https://github.com/dmikusa/bptest
- Refactor the code to use more constants where applicable (see the BP_XYZ_VERSION PR for example)
- Allow users to select which package manager to use (e.g. uv rather than pip)
- Modify the integration tests to use a lighter weight cpython buildpack

For that last one, the tests are using three different versions which is good to ensure that part is working but getting the buildpack ready for testing takes quite some time and space (see the action added in `github-config` to reclaim space from GitHub runners). What I had in mind was to replicate what I did on my workstation but by code for the test. 

On my workstation:
- Clone the cpython buildpack
- Comment out all but the latest version of the three python versions used by the tests for both jammy and noble.
- Change `integration.json` and point it to my local clone
- Modify `init_test.py` and a version number since building a local buildpack requires to have a number

What could be done in that new version is to fully manage cpython in code (cloning, change of `buildpack.yaml`) and remove it from the `integration.json` file.

That helps shave time off the test starting time. It should also be possible to build it only for the platform currently under test.

On a side note, there are some automation that are not fully working because they are tied to hard coded values such as allowed only by `paketo-bot`.

dmikusa (approved):

sgaist:
> I'm 👍 moving forward on this. I think we should shoot for something that is at parity with what we have now, so while I agree that your list of other things are worth doing, I think we should hold those off until we get this integrated and usable. Then we can add those things.

@dmikusa Thank you ! Sounds like a plan.

> I support this 👍🏻, but I have a slight concern about participation logic. 3 out of 5 are always participating, while in the majority of cases, only one is needed. I think it would make the build process simpler.

@TheSuperiorStanislav  Thanks !
I agree with the simplification. The current design was done mainly to be a drop-in replacement that would behave the same as the set of buildpacks we have currently.
> 
>     * (miniconda)-> Always -> if environment.yml is present

Good for me

> 
>     * (pip)-> Always -> If requirements.txt or file specified in `BP_PIP_REQUIREMENT` is present

Good for me as well

> 
>     * (pipenv) -> Always -> Only if Pipfile.lock is present in the root folder

Agreed

> 
>     * (poetry)-> pyproject.toml is present in the root folder -> Only if poetry.lock is present. (pyproject.toml is not poetry specific and can be present in any Python project)

Not fully agreed, this might break existing projects. That said, I already implemented a modified handling of `pyproject.toml`. It now checks for `build-system` in the file and reacts up on it. Depending on what it finds it will select one of the appropriate (and currently supported) build backend and if there's nothing default to poetry.

> 
>     * ((uv)-> uv.lock is present in the root folder -> Keep it as it

Nothing to add here.

I think that once things are set in motion, we can open an umbrella issue with the list of things we want to have and I'll implemented them.

TheSuperiorStanislav (approved):

jericop (approved):

PR #334, opened 2026-08-04
[RFC] Add PNPM package manager support to Node.js language family

taha2samy-3:


pacostas:
Can you please justify how within 17 minutes, you managed to investigate all the existing RFCs, think about the pnpm implementation, think about the order of the buidpacks and also right the RFC?

